PR #11, opened 2019-10-11
Adds remaining wrapper types to valueType map

Faendrek:
Adds remaining types to typeMap handle simplifying full list of wrapper types found at: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto


stephenh:
Cool, thanks!

stephenh:
For now I'm manually pushing to npm; I pushed out 1.5.0 which should have your change in it.

PR #18, opened 2020-01-14
fix map type detection when field name is not camelCase

SchroederChris:
First of all, thanks for ts-proto, I really appreciate finally having a protoc plugin that generates typescript interfaces instead of classes!

When a field in a .proto file has a map type and a non-camelCase name (e.g. `map<string, TraceSourceInfo> trace_source_info = 10;`), the generated code for this field was typed to an array instead of an index-signature object. Instead of relying on the string comparison of type and field name (which fails when the field name is snake_case), the relevant types can be looked up from the type map.

I didn't fully understand the test infrastructure, so I couldn't add any unit tests unfortunately.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,7 +1,7 @@
 import { google } from '../build/pbjs';
 import { Member, TypeName, TypeNames } from 'ts-poet';
 import { visit } from './main';
-import { fail, upperFirst } from './utils';
+import { fail } from './utils';
 import { asSequence } from 'sequency';
 import FieldDescriptorProto = google.protobuf.FieldDescriptorProto;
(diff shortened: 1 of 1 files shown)


stephenh:
Great, thanks for reporting the issue and great fix @SchroederChris . I added a test case and will push out a release soon, after looking at your other PR.

> First of all, thanks for ts-proto, I really appreciate finally having a protoc plugin that
> generates typescript interfaces instead of classes!

Cool! :-) I'm glad to hear that. When I started ts-proto I wasn't entirely sure the "just use lists / hashes / interfaces" approach would work, given how basically every other proto impl uses classes, but so far it is working out really pretty well, and, yeah, I agree, seems much more pleasant/idiomatic to use.

PR #19, opened 2020-01-14
fix fromJSON and fromPartial functions for map types

SchroederChris:
The generated fromJSON and fromPartial functions didn't iterate over the object entries of the original index-signature object. IF I recall it correctly, they instead interpreted `object` as having the `[FieldName]Entry` type and just tried to call fromJSON / fromPartial on that, which in turn resulted in undefined values instead of a correctly filled object.

As in the other PR, sorry for not adding tests ;-)


stephenh:
Wow, yeah, these were totally broken. Thanks for the fix! You also handled primitive values, which was a good call. I added test cases for maps of messages and maps of primitives, and this should be fixed in v1.7.0, which I just published (it has your other fix as well).

SchroederChris:
Cool!

> Thanks for the fix!

You're welcome :-D

Issue #20, opened 2020-01-15
Support to generate field names "as-is" instead of converting to camelCase


SchroederChris:
Currently, the original field names (as defined in a .proto file) are converted to camelCase for the generated typescript interfaces. It would be nice if there was an option to keep the original names (e.g. in snake_case) instead.

stephenh:
@SchroederChris yep, I think that makes sense, just hadn't included it yet b/c we'd personally not needed it. Want to submit a PR for this?

There is already support for a "context" parameter, i.e.:

```
  const options: Options = { useContext: false };
  if (parameter && parameter.includes('context=true')) {
    options.useContext = true;
  }
```

So should be pretty easy to reuse that approach to change `snakeToCamel(name)` into `maybeSnakeToCamel(opts, name)`.

SchroederChris:
@stephenh I just created pull request #22. That would also be the final change so that we can use your version again instead of our fork.

stephenh:
Great, thanks! Added a sanity check of running `ts-proto` with that option and a small test to make sure there are no compile errors, and pushed out as v1.8.0.

SchroederChris:
Perfect, thank you!

PR #22, opened 2020-01-16
add option to disable snake_case to camelCase conversion

SchroederChris:
pull request for #20 

The new option is named `snakeToCamel`. The default is set to `true`, so that the old behavior isn't changed unless explicitly wanted. @stephenh maybe you can add another test case.

### DIFF
--- a/integration/setup.ts
+++ b/integration/setup.ts
@@ -1,10 +1,11 @@
 import { mkdir, readFile, writeFile } from 'fs';
 import { parse } from 'path';
 import { google } from '../build/pbjs';
-import { generateFile, Options } from '../src/main';
+import { generateFile } from '../src/main';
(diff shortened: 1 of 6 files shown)


SchroederChris:
I just discussed this with a colleague and he argued for changing the option from boolean to a string describing the conversion mode, so that it can be easier extended in the future (for example to the one with hyphens or always converting to snake_case), i.e. sth like `caseConversion=camelCase` or `caseConversion=none` etc. It might be a good idea to do this directly instead of having an incompatible change later on.

jonaskello:
Is this option documented somewhere? It works but I could not find it in the readme (searched for snake).

stephenh:
@jonaskello yes, good catch, I just added it to the readme.

Issue #24, opened 2020-01-23
Support Long instead of throwing error


gnzzz:
Hello,
I'm converting a project to use ts-proto instead of protobufjs. In our protobuf definitions we use `uint64` a fair bit. Since Long isn't supported, but the numbers are larger than `MAX_SAFE_INTEGER`, the error in `longToNumber` is thrown at runtime.

I saw the note on the README about better support for Long, and I would be interested in hearing what your thoughts on it is. From my perspective, changing to always return a Long would be good solution, but that might break other projects. Taking a note from protobufjs and adding a protoc plugin option mimicking `--force-number` and `--force-long` could also be an option.

stephenh:
Hi @gnzzz yep, I think a plugin option is a good idea. There is an option already called `context` that shows how some of that works. Are you up for submitting a PR around that? I could give a few hints but in theory it shouldn't be too bad..

gnzzz:
I got an initial stab working for the option at https://github.com/gnzzz/ts-proto . It feels like I'm butchering the ts-poet statements though, so any pointers there would be helpful.

The output looks pretty good at the moment, and I will write a simple test in the same way as the other. If there are any other scenarios that I should test, please let me know.

stephenh:
@gnzzz this looked really good! I went ahead and merged it.

I was seeing some test/compile errors; can you check out this commit: https://github.com/stephenh/ts-proto/commit/5706ff9a9a4769c03365988ced941f40e1a9955b

...what's the thing with `import * as Long from "long"` vs. `import Long from "long"`? I was pretty sure one of the tsconfig values, like `esModuleInterop` or `allowSyntheticDefaultImports` allowed always using the 2nd version...

Should we change the code generated output to use `import Long from "long"`? For that commit above, I just did it by hand to see if it'd work / get the tests to pass.

gnzzz:
> ...what's the thing with `import * as Long from "long"` vs. `import Long from "long"`? I was pretty sure one of the tsconfig values, like `esModuleInterop` or `allowSyntheticDefaultImports` allowed always using the 2nd version...

[...]

PR #25, opened 2020-01-25
Add `forceLong` option to force use of Long for longer number types

gnzzz:
Pull request for #24. Adds an option `forceLong` which will use `Long` for large numbers. This matches the `--forceLong` in `pbjs`.

With this option, each of the protobuf basic number types maps as following.

| Protobuf number types  | Default Typescript types | `forceLong` Typescript types |
| ----------------------- | ----------------------- | ---------------------------- |
|  double | number | number  |
|  float | number | number |
|  int32 | number | number |
|  int64 | number* | Long |
|  uint32 | number | number |
|  uint64 | number* | Unsigned Long |
|  sint32 | number | number |
|  sint64 | number* | Long |
|  fixed32 | number | number |
|  fixed64 | number* | Unsigned Long |
|  sfixed32 | number | number |
|  sfixed64 | number* |  Long |

Where (*) indicates they might throw an error at runtime.

I.e. default behaviour is effectively `pbjs` `forceNumber`.


stephenh:
@gnzzz I really like your chart ^, maybe submit a PR to add it to the readme? ts-proto doesn't have super-structured docs / more than just the readme, but I think that'd be useful. Maybe in a section dedicated to longs...

PR #27, opened 2020-01-29
use undefined as the default value for oneof fields

SchroederChris:
Hi @stephenh,
we found another small issue. When using oneofs, all the options were populated with their default values in fromJSON / fromPartial / toJSON. When sending these objects to our backend (as json, not binary), they were rejected because they were no longer valid oneofs. I changed the defaults to undefined in that case (the corresponding types already were `[type] | undefined`). For non-primitive fields undefined already was the default btw.


stephenh:
Hi @SchroederChris great find and fix. I cherry-picked in your commit (so that I could re-codegen and add tests on top of it) and pushed this out as 1.10.0. Thanks!

PR #28, opened 2020-02-15
Added options to disable generation of serializers, to/from json, and…

n3rdyme:
- Added options to disable generation of serializers, to/from json, and services. 
- Added header to disable es-lint and ts-check

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -49,9 +49,9 @@ export type Options = {
   useContext: boolean;
   snakeToCamel: boolean;
   forceLong: boolean;
-  serializers: boolean;
-  toFromJson: boolean;
-  serviceStub: boolean;
+  outputEncodeMethods: boolean;
+  outputJsonMethods: boolean;
+  outputServiceImpl: boolean;
 };
 
 export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, parameter: string): FileSpec {
@@ -83,7 +83,7 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     }
   );
 
-  if (options.serializers || options.toFromJson) {
+  if (options.outputEncodeMethods || options.outputJsonMethods) {
     // then add the encoder/decoder/base instance
     visit(
       fileDesc,
@@ -93,12 +93,12 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
           .add('export const %L = ', fullName)
           .beginHash();
         
-        staticMethods = !options.serializers 
+        staticMethods = !options.outputEncodeMethods 
           ? staticMethods 
           : staticMethods.addHashEntry(generateEncode(typeMap, fullName, message, options))
             .addHashEntry(generateDecode(typeMap, fullName, message, options));
           
-        staticMethods = !options.toFromJson 
+        staticMethods = !options.outputJsonMethods 
           ? staticMethods 
           : staticMethods.addHashEntry(generateFromJson(typeMap, fullName, message, options))
             .addHashEntry(generateFromPartial(typeMap, fullName, message, options))
@@ -111,7 +111,7 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
       },
       options,
       (fullName, enumDesc) => {
-        if (!options.toFromJson) {
+        if (!options.outputJsonMethods) {
           return;
         }
         let staticMethods = CodeBlock.empty()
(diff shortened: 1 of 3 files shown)


n3rdyme:
@stephenh If you need a merged pull request for #28 and #31 just let me know.

src/main.ts
```diff
@@ -49,6 +49,9 @@ export type Options = {
   useContext: boolean;
   snakeToCamel: boolean;
   forceLong: boolean;
+  serializers: boolean;
```
stephenh: Overall this looks great; I'm going to bikeshed the names a bit...

Maybe prefix all of them with `output` (or `generate...` but `output...` is shorter) and change to:

* `outputServiceImpl` (I know grpc does use the term "stub" for these, but I think that's a misuse of the term; almost every where else, stub == a test fake)
* `outputEncodeDecodeMethods`
* `outputToFromJsonMethods`

wdyt?

src/plugin.ts
```diff
@@ -5,6 +5,15 @@ import { generateFile } from './main';
...
 import CodeGeneratorRequest = google.protobuf.compiler.CodeGeneratorRequest;
 import CodeGeneratorResponse = google.protobuf.compiler.CodeGeneratorResponse;
+import { FileSpec } from 'ts-poet';
+
+// Comment block at the top of every source file, since these comments require specific
+// syntax incompatible with ts-poet, we will hard-code the string and prepend to the
+// generator output.
+const formatSourceFile = (spec: FileSpec) => 
+`// @ts-nocheck
+/* eslint-disable */
```
stephenh: Per my other comment, it seems like `ts-nocheck` should really not be here, i.e. what are the compile errors we're suppressing? I'd also generally prefer removing the `eslint-disable` and instead making the generated code as idiomatic / clean as possible, but I'm not a hard -1 on that, especially if you have specific eslint errors you're getting now that are just unrealistic to have ts-proto avoid.

n3rdyme: @stephenh, The reasoning behind `@ts-nocheck` is that you are not in control of the client's tsconfig settings, and thus you need to generate 100% perfect code under all conditions and configurations. That just isn't reasonable and there are a few errors already if I remove that setting:
```
generatedproto.ts:127:10 - error TS6133: 'longToNumber' is declared but its value is never read.

127 function longToNumber(long: Long) {
             ~~~~~~~~~~~~

generatedproto.ts:134:6 - error TS6196: 'DeepPartial' is declared but never used.

134 type DeepPartial<T> = {
         ~~~~~~~~~~~
```

The nocheck flag is not harmful and prevents the need for consumers either edit generated files and commit them, or to generate code into a separate directory excluded from typescript.

src/main.ts
n3rdyme: @stephenh,
Sounds good, I shortened a bit of your example and came up with this if it's acceptable:
```
export type Options = {
  useContext: boolean;
  snakeToCamel: boolean;
  forceLong: boolean;
  outputEncodeMethods: boolean;
  outputJsonMethods: boolean;
  outputServiceImpl: boolean;
};
```
Just trying to keep them on the shorter side for the command-line.
Close enough?

src/plugin.ts
n3rdyme: In regards to eslint-disable, that really isn't a possibility to solve in code generation. 2 spaces vs 4, prettier formatting, interfaces start with "I" or interfaces can't start with "I", take your pick, you can't please everyone's lint settings ;)

n3rdyme:
@stephenh 
I hope I addressed all the questions and asks above. As to the source comments for disabling linters, I feel pretty strongly that code generators should provide this. If you are against it being on by default, perhaps we can make this an option instead?

src/plugin.ts
stephenh: eslint I'm flexible on, as you're right, `I` prefix type stuff, that's fair...it's been awhile since I've been in an eslint-rule-heavy codebase with super-specific rules.

Formatting, imo it's easier to lean on "just run prettier after codegen", which technically the code gen library we're using, `ts-poet`, _should_ do, but doesn't in the current v2 version of ts-poet we're using (the v3 version of ts-poet does run prettier automatically on the output with whatever the downstream project's prettier config is; granted this assumes they're using prettier). 

I'm still very hesitant about `ts-nocheck`, because AFAICT it literally turns off on all type checking in the file? That just seems non-negotiable imo? I get we'll have to cater to a kind of most-strict-tsconfig/whatever output, but I'd rather try and do that, than loosing type checking on the entire file.

Let me know if my understanding of ts-nocheck is incorrect.

Fwiw the `longToNumber` / `DeepPartial` are valid issues that I'll have work on a change to conditionally output instead of being unnecessary cruft in the output file.

Which, in general in terms of "unncessary/bad cruft in the output file", I have used "wow that generated code is really ugly so just turn off all the warnings" code generators a lot before, and it seemed like a cop-out insofar as it lets the code generator off the hook/still be lazy vs. just fixing their output to not suck. So, that is my general bias/guiding principle with ts-proto, is that the output should just not suck.

I'll compromise on eslint, you're right that can get esoteric, but would like to try and avoid `ts-nocheck`.

stephenh:
I'll locally merged this PR and pushed it to master (there was a minor conflict from your other PR).

I also made the `longToNumber` and `DeepPartial` output conditional, so removed `ts-nocheck` but kept always outputting `eslint-disable`.

If you have more examples of ts-nocheck-requiring behavior, I'm happy to try and fix those too. I'm slightly (_slightly_ :-)) open to a `outputTsNoCheck` config flag.

stephenh:
Can you check how master works for you with your two PRs merged and then I'll make a release?

Thanks again!

src/main.ts
stephenh: Yep, your names looked good.

n3rdyme:
@stephenh I tested from latest master:
```
commit a7a20484bdd8dc96d402090060e9696fac89ebcb (HEAD -> master)
Author: Stephen Haberman <stephen.haberman@gmail.com>
Date:   Sun Feb 16 15:54:58 2020 -0600

    Link.
```

Everything is working for me. Thanks for the merge!

BTW... if anyone does need @ts-nocheck, they can do this at the top of the offending proto files:
```
//@ts-nocheck
syntax = "proto3";

...
```

Which then outputs in the source as needed:
```
/* eslint-disable */
// @ts-nocheck
//
import { Empty } from '../../google/protobuf/empty';
...
```

So, I guess that works ;)

stephenh:
> So, I guess that works ;)

Lol, that's hilarious. :-)

Great that it works for you. I'd gone ahead and released master (for a few non-doc commits before that) as v1.11.0.

Issue #23, opened 2020-01-17
Support for generating TSDoc content from the message and field level comments in the Protobuf definition


dteirney:
Hello,

Has there been any thought around using the comments for the protobuf message and field definitions to generate TSDoc comments for the generated typescript?

We'd love to be able to have the field comments from the proto files available for devs within VS Code and other editors. Many IDEs and documentation generators support comments using the [TSDoc format](https://github.com/Microsoft/tsdoc).

Many of our proto definitions follow both [comment guidelines for Uber prototool](https://github.com/uber/prototool/blob/dev/style/README.md#documentation) and [Google Documentation AIP](https://aip.dev/192). It would be awesome for those comments to come along for the ride with the TS that's generated, similar to the comments that are generated with the [Google GAPIC typescript generator](https://github.com/googleapis/gapic-generator-typescript).

Thanks!
David

stephenh:
@dteirney sure, that makes a lot of sense. The ts-poet code generator library that ts-proto uses supports outputting comments, I just hadn't bothered hooking that up yet. If you'd like to take a crack at it, that'd be great.

Most of the ts-poet `addFunction`/`addProperty` calls take builders that accept a "javaDoc" parameter for outputting comments, and the incoming CodeGenerationRequest (forget the exact name) protos from the protoc compiler should have `.comment`-type fields that are easily accessible.

(Somewhat of a small disclaimer, the latest ts-poet release, v3, doesn't use this same builder api anymore, so if you dig too much into ts-poet, just note that ts-proto is still on the v2 ts-poet api.)

I can probably sneak this in to at some point too.

n3rdyme:
@dteirney see https://github.com/stephenh/ts-proto/pull/31

stephenh:
This is merged to master and will go out in the next release, which will be next-day-ish or so. Going to go ahead and close. Thanks @n3rdyme ! That was actually more involved than I thought it was going to be, with the index-based/etc. encoding of the comments in the protobuf file.

dteirney:
@n3rdyme thanks! Will test once the next release goes out.

stephenh:
@dteirney version 1.11.0 is out with this change. :-)

PR #31, opened 2020-02-15
#23 - Add source comments from file descriptor

n3rdyme:
Added comments from fileDescriptor.sourceCodeInfo to enums, message interfaces, and services.

Should resolve issue 23
https://github.com/stephenh/ts-proto/issues/23

### DIFF
--- a/src/sourceInfo.ts
+++ b/src/sourceInfo.ts
@@ -1,0 +1,88 @@
+import { google } from '../build/pbjs';
+import FileDescriptorProto = google.protobuf.FileDescriptorProto;
+
+export type FieldID = number;
+
+export const Fields = {
+    file: {
+        syntax: 12,
+        message_type: 4,
+        enum_type: 5,
+        service: 6,
+        extension: 7,
+    },
+    message: {
+        field: 2,
+        nested_type: 3,
+        enum_type: 4,
+    },
+    enum: {
+        value: 2,
+    },
+    service: {
+        method: 2,
+    },
+}
+
+export interface SourceDescription { 
+    readonly span: number[];
+    readonly leadingComments: string;
+    readonly trailingComments: string;
+    readonly leadingDetachedComments: string[];
+};
+
+class EmptyDescription implements SourceDescription {
+    span = [];
+    leadingComments = '';
+    trailingComments = '';
+    leadingDetachedComments = [];
+};
+
+export type SourceInfoMap = {
+    [key: string]: SourceDescription
+};
+
+export default class SourceInfo implements SourceDescription {
+    static empty() {
+        return new SourceInfo({}, new EmptyDescription());
+    }
+
+    static fromDescriptor(file: FileDescriptorProto) {
+        let map: SourceInfoMap = {};
+        if (file && file.sourceCodeInfo && file.sourceCodeInfo.location) {
+            file.sourceCodeInfo.location.reduce((m, loc) => {
+                m[loc.path.join('.')] = loc;
+                return m;
+            }, map);
+        }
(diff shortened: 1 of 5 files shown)


src/sourceInfo.ts
```diff
@@ -0,0 +1,88 @@
...
+    },
+    enum: {
+        value: 2,
+    },
+    service: {
+        method: 2,
+    },
+}
+
+export interface SourceDescription {
```
stephenh: Can you add a link to the docs (assuming they exist, maybe the protobuf declaration file?) of what these fields are?

src/sourceInfo.ts
```diff
@@ -0,0 +1,88 @@
+import { google } from '../build/pbjs';
+import FileDescriptorProto = google.protobuf.FileDescriptorProto;
+
+export type FieldID = number;
+
+export const Fields = {
+    file: {
```
stephenh: I don't have it setup to auto-run, but there is a `.prettierrc` checked in; can you run prettier on this?

src/sourceInfo.ts
```diff
@@ -0,0 +1,88 @@
...
+
+class EmptyDescription implements SourceDescription {
+    span = [];
+    leadingComments = '';
+    trailingComments = '';
+    leadingDetachedComments = [];
+};
+
+export type SourceInfoMap = {
+    [key: string]: SourceDescription
```
stephenh: Maybe add an in-source comment about what some example keys are?

src/sourceInfo.ts
```diff
@@ -0,0 +1,88 @@
...
+    get span() { return this.selfDescription.span; }
+    get leadingComments() { return this.selfDescription.leadingComments; }
+    get trailingComments() { return this.selfDescription.trailingComments; }
+    get leadingDetachedComments() { return this.selfDescription.leadingDetachedComments; }
+
+    lookup(type: FieldID, index: number): SourceDescription {
+        return this.sourceCode[`${type}.${index}`] || new EmptyDescription();
+    }
+
+    open(type: FieldID, index: number): SourceInfo {
```
stephenh: What's the different between `lookup` and `open`? It looks like open...finds multiple results? Add in-source method comments for each?

src/sourceInfo.ts
n3rdyme: @stephenh Done.

src/sourceInfo.ts
n3rdyme: @stephenh Done.

src/sourceInfo.ts
n3rdyme: @stephenh Done.

src/sourceInfo.ts
n3rdyme: @stephenh Added some comments... generally lookup is get one thing, open is open the type so I can lookup more relative to that type. Another way to say it, Message has nested types, so use `open`, for everything else you can just use `lookup`.

n3rdyme:
@stephenh 
I added the requested comments in simple.proto and cleaned up the conditional logic a bit. I assume you did not want me to regenerate all the binary and source files on branch. It seems like that might make merging difficult, so I assume you are keeping PR branches clean?

Otherwise, I think I covered all the asks above. Let me know if you need any additional changes.

stephenh:
Looks great, thanks! I do actually check in the generated source files; I know in downstream / real projects that is an anti-pattern but given ts-proto's job is producing that output, I basically treat it as an mini-integration test to see how the output has changed in the PRs/diffs/etc. I'll go ahead and do a follow up commit with those changes now.

PR #38, opened 2020-03-15
Add use enum names option

jeffpyke:
Adds option to generate enums in typescript with string values rather than integer.
example:
```ts
export enum Animals {
  CAT = "CAT",
  DOG = "DOG",
}
```


bastianeicher:
We use ts-proto to generate a REST client library for accessing our API via grpc-gateway (i.e. the "types only" use case). Our instance of grpc-gateway is configured to serialize enum values as strings rather than ints. Therefore this feature would also be very useful for us.

@stephenh Perhaps the need for an `enumToInt` method could be obviated by allowing `useEnumNames=true` only in combination with `outputEncodeMethods=false`?

stephenh:
@bastianeicher yeah, that is technically true; fwiw I don't think the `enumToInt` would be that hard, so I'd prefer not having an option that only worked with other options enabled/disabled (granted, given the amount of flags that's inevitable at some point). But otherwise yes.

stephenh:
`stringEnums` and `outputEncodeMethods` now play nicely together, as of v1.75.0.

Issue #41, opened 2020-03-27
enums don't default to UNRECOGNIZED value like java


mscolnick:
To be compatible with newly added enums in proto definitions, the typescript enum deserialization should default to `UNRECOGNIZED` / `-1`

https://github.com/stephenh/ts-proto/blob/master/src/main.ts#L321

This is what java generates for enum

```java
public enum ChangeType implements ProtocolMessageEnum {
    ADDED(0),
    REMOVED(1),
    MODIFIED(2),
    UNRECOGNIZED(-1);
```

Happy to make a contribution if you are ok with the update.

stephenh:
Ah yeah, good point, that makes sense. A PR would be great.

Just a heads up, I very recently moved away from `enum` the keyword to this `enum`-via-`const`+`type` pattern b/c babel didn't like using `namespace`s to add static methods, i.e. `FooEnum.toJSON(foo)`, but otherwise, the output/API still looks very enum-ish, so should be pretty amenable to your feature request/PR.

mscolnick:
makes total sense. will put something up within the week. thank you

stephenh:
I merged and pushed to npm release 1.15.0 with this change. Thanks for the PR!

PR #42, opened 2020-03-30
Add unrecognized enum value

mscolnick:
Fixes https://github.com/stephenh/ts-proto/issues/41

Update enum default to be `UNRECOGNIZED` / `-1` to be consistent with how java generates enums.

e.g.
```java
public enum ChangeType implements ProtocolMessageEnum {
    ADDED(0),
    REMOVED(1),
    MODIFIED(2),
    UNRECOGNIZED(-1);
```

### DIFF
--- a/integration/simple/simple-json-test.ts
+++ b/integration/simple/simple-json-test.ts
@@ -45,18 +45,18 @@ describe('simple json', () => {
     expect(s2.state).toEqual(StateEnum.ON);
   });
 
-  it('fails decode json with invalid numeric enum values', () => {
+  it('can decode json with unrecognized numeric enum values', () => {
     // given state is mapped as 1, which is not a valid numeric value
     const s1 = { state: 1 };
-    // then we fail fast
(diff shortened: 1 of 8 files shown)


mscolnick:
@stephenh i updated the type-union to have `-1`. let me know if that feels off, but it allows the consumer to do an exhaustive check on the type and force them to handle the unrecognized case.

integration/simple/simple-json-test.ts
```diff
@@ -45,18 +45,18 @@ describe('simple json', () => {
...
   });
 
-  it('fails decode json with invalid numeric enum values', () => {
+  it('can decode json with unrecognized numeric enum values', () => {
     // given state is mapped as 1, which is not a valid numeric value
     const s1 = { state: 1 };
-    // then we fail fast
-    expect(() => Simple.fromJSON(s1)).toThrow('Invalid value 1');
+    // then we decode to UNRECOGNIZED
+    expect(Simple.fromJSON(s1).state).toEqual(StateEnum.UNRECOGNIZED);
```
stephenh: Great!

PR #43, opened 2020-04-03
Add returning of observables and supporting metadata

iangregsondev:
The following change is very useful for NestJS users. It allows you to return observables in place of promises.

It also supports passing a metadata type as the last argument on the interfaces.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-proto",
-  "version": "1.15.0",
+  "version": "1.16.0",
   "description": "",
   "main": "build/plugin.js",
   "repository": "github:stephenh/ts-proto",


iangregsondev:
@stephenh I would appreciate a quick code review :-)

Working with ts-poet was challenging.

The 2 additional options that were added are opt out by default.

I have updated the README with what they do.

I open to any comments.

I have tested it here locally and all seems to be working as expected.

README.markdown
```diff
@@ -150,6 +150,11 @@ Supported options:
 
 (I.e. you want only interface declarations for your Protobuf types, then pass all three of `outputEncodeMethods`, `outputJsonMethods`, and `outputClientImpl` as `false`, i.e. `--ts_proto_opt=outputEncodeMethods=false,outputJsonMethods=false,outputClientImpl=false`.)
 
+The following options are very useful for NestJS and available if `outputEncodeMethods`, `outputJsonMethods`, and `outputClientImpl` are set as `false`
+
+* If you pass `--ts_proto_opt=useMetadata=true`, the last argument accepts grpc metadata
```
stephenh, line 155: Maybe `addGrpcMetadataParam`? or `addGrpcMetadata`?

README.markdown
```diff
@@ -150,6 +150,11 @@ Supported options:
 
 (I.e. you want only interface declarations for your Protobuf types, then pass all three of `outputEncodeMethods`, `outputJsonMethods`, and `outputClientImpl` as `false`, i.e. `--ts_proto_opt=outputEncodeMethods=false,outputJsonMethods=false,outputClientImpl=false`.)
 
+The following options are very useful for NestJS and available if `outputEncodeMethods`, `outputJsonMethods`, and `outputClientImpl` are set as `false`
```
stephenh, line 153: Just curious, but why do these other options need to be false? AFAIU it should be fine if these methods lived on the `FooMessage` entity, i.e. `FooMessage.toJSON/fromJSON`, but you go ahead and add `metadata` to the `Service` interface?

README.markdown
iangregsondev: Sure, sounds logical - I will change it to addGrpcMetadata

README.markdown
iangregsondev: To be truthful I didn't know if they were applicable, but let me have a look at adding them.

iangregsondev: Hi, could you give me some more details. What I have been able to do is add this information into the concrete types. I need to change the code contents a bit. Let me show an example of the generated code.

```
export interface HeroService {

  FindOne(request: HeroById, metadata: Metadata): Observable<Hero>;

  FindTwo(request: HeroById, metadata: Metadata): Observable<Elses>;

}

export class HeroServiceClientImpl implements HeroService {

  private readonly rpc: Rpc;

  constructor(rpc: Rpc) {
    this.rpc = rpc;
  }

  FindOne(request: HeroById, metadata: Metadata, request: HeroById): Observable<Hero> {
    const data = HeroById.encode(request).finish();
    const promise = this.rpc.request("hero.HeroService", "FindOne", data);
    return promise.then(data => Hero.decode(new Reader(data)));
  }

  FindTwo(request: HeroById, metadata: Metadata, request: HeroById): Observable<Elses> {
    const data = HeroById.encode(request).finish();
    const promise = this.rpc.request("hero.HeroService", "FindTwo", data);
    return promise.then(data => Elses.decode(new Reader(data)));
  }

}
```

So as you can see the metadata is there and it also returns an Observable, I need to update the lines 

```
    return promise.then(data => Elses.decode(new Reader(data)));
```

to

```
    from(promise).pipe(
         map((item) => {
           return Hero.decode(new Reader(data))
        }
    )
```

So the above would take the promise and return an observable. I need to check the above for syntax errors :-)

I didn't really understand the toJson / fromJson 

This should not need to change right ?

iangregsondev: The above concrete types would not be usable by NestJs - as it has its own implementation client library so only interfaces are required BUT I think the above would be useful for somebody using the standard grpc client library and wanted to work directly with observables.

Are you aware of anything else that should change.

iangregsondev:
Having a problem running the tests, not the new ones, but the existing ones.

```
 FAIL  integration/simple/simple-value-wrappers-test.ts
  ● Test suite failed to run

    integration/simple/simple-value-wrappers-test.ts:3:40 - error TS2307: Cannot find module './pbjs'.

    3 import { google, simple as pbjs } from './pbjs';
                                             ~~~~~~~~
    integration/simple/simple-value-wrappers-test.ts:95:25 - error TS7006: Parameter 'c' implicitly has an 'any' type.

    95     expect(s2.coins.map(c => c.value)).toEqual([1]);
                               ~

 FAIL  integration/simple/simple-test.ts
  ● Test suite failed to run

    integration/simple/simple-test.ts:11:40 - error TS2307: Cannot find module './pbjs'.

    11 import { simple as pbjs, google } from './pbjs';
                                              ~~~~~~~~

```

The directory pbjs ??

README.markdown
iangregsondev: Or maybe this isn't a good idea adding it to the concrete types, if a person needs to have an observable from the existing promises that all that is needed is for them to do 

```
   from (...some promise)...
```

But then of course you would have an interface that returns an observable and a concrete type that returns a Promise 

:-(

stephenh:
Per the test failures, I think there is step in the readme that covers generating that.

iangregsondev:
> Parameter 'c' implicitly has an 'any' type.

I have a la

> Per the test failures, I think there is step in the readme that covers generating that.

Yep - I need to run the tests inside the build folder - Got other issues with regards to implicit any's etc - but I think this is because I am using the latest version of typescript that is in strict mode - I can fix this I think. I have seen it before

iangregsondev:
> Per the test failures, I think there is step in the readme that covers generating that.

Ermm - do the tests run for you ?

When I run ./pbjs.sh which is mentioned in the readme - this is what you are referring to right ?

It creates some files inside the build directory, but all test files are loading them relatively.

for example

```
 FAIL  integration/simple/numbers-test.ts
  ● Test suite failed to run

    integration/simple/numbers-test.ts:3:32 - error TS2307: Cannot find module './pbjs'.

    3 import { simple as pbjs } from './pbjs';
                                     ~~~~~~~~

```

but it's relative to the testfile, there is no such file, it resides in the /build directory - right ?

Or maybe I am not quite doing it right - I think I am but something isn't right.

if you are able to launch the tests.

I tried running yarn tests

from the root and also from the build directory - but no go.

I could go through an change the import reference for pbjs to the build directory version but that sounds kind of wrong - if they are working for you

README.markdown
iangregsondev: Hi @stephenh I have added the change to the parameter (addGrpcMetadata) and updated the readme.

I think that's more semantic right?

README.markdown
iangregsondev: I haven't pushed these changed at the moment, I think it needs more discussion. I mean with the NestJs it is really needed as we need an interface that can directly offer an Observable. The interface is NOT implemented but uses as part of the NestJs client library - without this sort of functionality then you have to produce these by hand.

With regards to the concrete types, these can be called using the "from" command which converts a promise to an observable.

```
 from(some promise, from the concrete type).pipe(....do something else)
```

I think we need to think if it is really beneficial to people - I am not sure it would be.

What do you think @stephenh  ?

iangregsondev:
@stephenh I have another stab at the integration tests but I am having problems even getting them to run. The main problem is the missing pbjs but there are some strict typescript compiler errors - i think these are fixable but this fix won't help with the missing pbjs.

I have run the shell script as per the readme but this doesn't help.

It would be nice to know why it's not working BUT also that if we decided that my above changes only make sense on interfaces (for use by NestJS) - then unit testing interfaces is not possible anyway and wouldn't make sense considering interfaces are compiled away at compile time.

src/main.ts
```diff
@@ -933,7 +941,19 @@ function generateService(
     maybeAddComment(info, text => (requestFn = requestFn.addJavadoc(text)));
 
     requestFn = requestFn.addParameter('request', requestType(typeMap, methodDesc));
-    requestFn = requestFn.returns(responsePromise(typeMap, methodDesc));
+
+    // Use metadata as last argument for interface only configuration
+    if (!options.outputClientImpl && !options.outputEncodeMethods && !options.outputJsonMethods && options.addGrpcMetadata) {
+      requestFn = requestFn.addParameter('metadata?', "Metadata@grpc");
```
iangregsondev, line 947: Make the metadata an optional type, the consumer shouldn't be forced to send it.

stephenh:
Hi @iangregsondev okay, so I've looked into the nestjs grpc docs. Thanks for those.

So, are you trying to generate an interface for the client to use, or the server to use, or both?

Per the tests, you've not really given me any concrete errors/steps to work off, other than "pbjs". Fwiw circleci builds the repo from scratch on each commit, so if you follow what it does:

https://circleci.com/gh/stephenh/ts-proto

https://github.com/stephenh/ts-proto/blob/master/.circleci/config.yml

It should work. If you want to open a new issue dedicated to the problem you're having with the tests, and past the commands you're running in order, and the console errors that result, I can take a look.

Let me know about the nestjs client vs. server thing, and we can go from there. So far you've mentioned including the `metadata` parameter, which AFAICT is a server-side only thing (i.e. only the controllers use it), but then some of your pasted code insinuates your actually building a client. Or maybe both?

stephenh:
Huh, what's pullbot? Not sure why it closed this.

iangregsondev:
@stephenh ah ah, is that my end ? Pullbot - i need to have a look :-(

The interface works for both Client and Server.

let me show you the client code - it uses the interface created by the ts-proto modified library

```
  constructor(@Inject("HERO_PACKAGE") private readonly client: ClientGrpc) {}

  onModuleInit() {
    this.heroService = this.client.getService<HeroService>("HeroService")
  }

  @Get()
  getById() {
    const tt = new Metadata()
    tt.add("Authorization", "some bearer token")
    return this.heroService.findOne({ id: 1 }, tt)
  }

```

As you can see above, this is the client and it calls using the inteface.

Here is the server, exactly the same interface, my controller class implements implements exactly the same interface

```
@Controller("authorization-service-feature-oauth")
export class AuthorizationServiceFeatureOauthController implements HeroService {
```

so now I get to use the implemented method

```
  @GrpcMethod("HeroService", "FindOne")
  findOne(request: HeroById, metadata?: Metadata): Observable<Hero> {
    console.log("here is data! ", request.id)
    console.log("this is the metadata ", metadata)

    const items = [
      { id: 1, name: "John" },
      { id: 2, name: "Doe" },
    ]

    const item = items.find(({ id }) => id === request.id)

    return of(item)
  }
```
So the method signature is exactly the same from the interface. It is the same interface. No difference what so ever.

I have to decorate the method with the following

```
  @GrpcMethod("HeroService", "FindOne")
```

This is manual, I can even clean it more - by naming my controller "HeroService" and then the decorator can become just

```
@GrpcMethod()
```

for visibility - here is the interface that was created from the PROTOS and via the fork of the ts-proto.

```
import { Metadata } from "grpc"
import { Observable } from "rxjs"

export interface HeroById {
  id: number
}

export interface Hero {
  id: number
  name: string
}

export interface HeroService {
  findOne(request: HeroById, metadata?: Metadata): Observable<Hero>
}

```

iangregsondev:
@stephenh Going to try and figure out what the bot is :-)

iangregsondev:
@stephenh It appears that was my fault - I was using a bot to keep the fork up to date from the main repo.

My bad, i removed the bot my end - it seems like a really good idea but it wasn't.

I can open up a new pull request. You cannot open old ones once they have been closed.

I will do it shortly, I will send concrete examples via ISSUES about the unit tests no problems.

Let me know when you have read the above and if I. have answered all your questions.

Thanks.

iangregsondev:
@stephenh  With regards to the metadata, it is now optional and yes it can be used from the client but doesn't need to be.

It can be used for passing anything but my specific use case was a bearer token.

I have an AuthGuard interceptor my end (server) to check to ensure that the bearer is valid otherwise it will return an error.

The server method (grpc method) can do something with the metadata or not.

In my case I already have an AuthGuard taking care of that - so I would technically do nothing..

Here is the AuthGuard info for being transparent.

I decorate this on the Grpc Method

```
@UseGuards(GrpcAuthGuard)
```

and use the following - untested :-) My next task :-)

```
import { Injectable, CanActivate, ExecutionContext } from "@nestjs/common"
import { Observable, throwError } from "rxjs"
import { JwtService } from "@nestjs/jwt"

@Injectable()
export class GrpcAuthGuard implements CanActivate {
  constructor(private jwtService: JwtService) {}

  canActivate(context: ExecutionContext): boolean | Promise<boolean> | Observable<boolean> {
    const type = context.getType()
    const prefix = "Bearer "
    let header
    if (type === "rpc") {
      const metadata = context.getArgByIndex(1) // metadata
      if (!metadata) {
        return false
      }
      header = metadata.get("Authorization")[0]
    }

    if (!header || !header.includes(prefix)) {
      return false
    }

    const token = header.slice(header.indexOf(" ") + 1)
    try {
      const valid = this.jwtService.verify(token)
      return valid
    } catch (e) {
      return false
    }
  }
}

```

I hope that answers that a little bit more.

PR #46, opened 2020-04-06
Enum as const

hamiltop:
Can't remember how to test this, would appreciate some guidance or just adding the tests for me 🤷‍♂

### DIFF
--- a/integration/vector-tile/__snapshots__/vector-tile-test.ts.snap
+++ b/integration/vector-tile/__snapshots__/vector-tile-test.ts.snap
@@ -73,11 +73,11 @@ function longToNumber(long: Long) {
 }
 
 export const Tile_GeomType = {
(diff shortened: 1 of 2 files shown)


stephenh:
@hamiltop nice! It's newly-written-up and also hidden a bit, but this section in the readme talks about the integration tests:

https://github.com/stephenh/ts-proto#building

That will re-run the codegen and so make the codegen changes show up in your PR/diff, which would be good enough for me.

If you wanted to also add a test somewhere in here:

https://github.com/stephenh/ts-proto/blob/master/integration/simple/simple-test.ts

But actually a new unit test is probably not very useful until we can use the `ts-expect-error` to ensure-but-not-fail-the-build that compile error you want. So could skip that for now.

stephenh:
Should be in v1.16.0. Thanks!

PR #48, opened 2020-04-07
Support bytes as base64 encoded strings.

fxb:
I took a quick stab at implementing support for bytes as base64 encoded strings in the `fromJSON` and `toJSON` methods.

It can be turned on by passing `--ts_proto_opt=forceJsonBase64Bytes=true`.

One limitation is that it currently relies on `window.atob` and `window.btoa` being available, meaning it would work in browsers, but not out-of-the-box in Node.

There are different approaches to solve this limitation:
  * Implement Node support using  [`Buffer`](https://nodejs.org/api/buffer.html)
  * Depend on a JavaScript base64 library (e.g. [base64-js](https://www.npmjs.com/package/base64-js))
  * Require the user to polyfill `window.atob` and `window.btoa`
  * Require the user to somehow supply base64 encode / decode functions

For now I have chosen to just support browsers, as this feature is opt-in.

Please let me know what you think :-)


src/main.ts
```diff
@@ -784,16 +828,20 @@ function generateToJson(
           from,
           defaultValue(field.type, options)
         );
-      } else {
```
fxb, line 787: Collapsing this `else` statement into an `else if` made this diff kind of hard to read. :-/

stephenh:
Thanks for the PR!

I'm pretty surprised that you're able to get this working in the browser, b/c I've put literally zero thought or effort into browser support so far. :-)

The diff looks great. I'd like to have an integration test, could you look at adding a new `integration/...pick a name...` directory and add like a minimal `*.proto` file that generates the code you expect? There are some `update-bins.sh` and `codegen.sh` helper files in the `integration/` directory that should just automatically pick up your new directory and generate the options for it. (And use `parameters.txt` for setting your force json param).

I suppose the biggest wrinkle for that is that right now the tests just run in node, we could either a) have the integration test just be "here is the generated code, make sure it looks fine, we won't actually run it, or b) somehow actually run the tests in a browser, or c) mock out the atoi/etc. functions for test.

I suppose c) is probably a combination of best + easiest at this point. Are you okay looking into that? I can poke around too if you'd rather not, but it won't be as quick as you doing it. :-) 
 
Just sanity checking, but it seems like eventually ts-proto could produce a single set of `.ts` files that works in both the browser or node by detecting window.atob vs. Buffer? Just hoping we won't have to consider something like producing separate build artifacts for each environment.

README.markdown
```diff
@@ -144,6 +144,7 @@ Supported options:
 
 * If you pass `--ts_proto_opt=context=true`, the Twirp services will have a Go-style `ctx` parameter, which is useful for tracing/logging/etc. if you're not using node's `async_hooks` api due to performance reasons.
 * If you pass `--ts_proto_opt=forceLong=long`, all 64 bit numbers will be parsed as instances of `Long` (using the [long](https://www.npmjs.com/package/long) library). Alternatively, if you pass `--ts_proto_opt=forceLong=string`, all 64 bit numbers will be outputted as strings.
+* If you pass `--ts_proto_opt=forceJsonBase64Bytes=true`, the `Message.fromJSON` and `Message.toJSON` methods will treat bytes as base64 encoded strings. This currently relies on `window.atob` and `window.btoa` being available.
```
stephenh, line 147: Hm, kind of thinking, does this need to be an option? I.e. could we just always output this code? I don't think that would break anyone's code/behavior today, because this is just not working at all AFAIU.

Granted, it's not going to work on node today, but at some point someone who is running this on node will file an issue, and we can fix the code then to support both node+browser...without any flags, I think?

fxb:
@stephenh Yes, I can add an integration test. I actually had a quick look but couldn't really find any existing  `toJSON` / `fromJSON` tests that I could copy and modify ;-)

For this I want to pre-generate a `.json` file instead of a `.bin` file, right?

Unfortunatley running the shell scripts also failed to due to some incompatibilities with the macOS version of bash (`shopt -s globstar`). I can probably solve that by installing a newer version via Homebrew.

README.markdown
fxb: I just assumed that multiple encodings for `bytes` could be chosen, but checking the [docs](https://developers.google.com/protocol-buffers/docs/proto3#json) base64 is the one and only encoding for the JSON mapping. I'll remove the option.

fxb:
@stephenh I force pushed a few commits adding an integration test, removing the option and fixing two little things I ran into.

README.markdown
stephenh: Great!

integration/update-bins.sh
```diff
@@ -1,10 +1,10 @@
-#!/bin/bash
+#!/usr/bin/env bash
 shopt -s globstar
 
 # To avoid running the protoc pipeline all the time, we capture the incoming Plugin
 # proto requests into .bin files that then unit tests can pull in directly as needed.
 
-for file in simple/simple.proto; do
+for file in **/*.proto; do
```
stephenh, line 7: Ah shoot, thanks for fixing that.

integration/bytes-as-base64/parameters.txt
```diff
@@ -0,0 +1 @@
+outputEncodeMethods=false,outputJsonMethods=true
```
stephenh, line 1: Pretty minor, but outputJsonMethods is already true by default, so maybe leave out?

fxb:
Thanks for getting this in! I might create some more PRs in the future :-)

stephenh:
> I might create some more PRs in the future 

Please do! :-) Fwiw I'm not actually using protobufs on a daily basis anymore (I changed jobs, and my new place isn't using them right now), so I'm happy to keep stewarding the project b/c I think it is generally a pretty pleasant way to work with TS & protobuf, but it'd be great to have some help maintaining / improving it.

stephenh:
(Also, I pushed release 1.18.0 with this change.)

PR #51, opened 2020-04-09
Removing generation of Timestamp conversion functions

alexrodriguez:


src/main.ts
```diff
@@ -153,7 +153,7 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     },
     options
   );
-  if (hasAnyTimestamps) {
+  if (hasAnyTimestamps && options.outputJsonMethods) {
```
fxb, line 156: Is this correct? I see `toTimestamp` and `fromTimestamp` being used in the `generateEncode` and `generateDecode` functions.

alexrodriguez: You are correct. I need to be more granular.
```suggestion
  if (hasAnyTimestamps && (options.outputJsonMethods || options.outputEncodeMethods) {
```
then in `addTimestampMethods(file: FileSpec, options: Options): FileSpec` add the `fromJsonTimestamp` conditionally.

PR #53, opened 2020-04-09
Only types

alexrodriguez:
Adding fix for when encode methods are requested in options but not json output.


integration/only-types/parameters.txt
```diff
@@ -0,0 +1 @@
+outputEncodeMethods=false,outputJsonMethods=false,outputClientImpl=false
```
stephenh, line 1: Huh, do you know why these `integration/only-types` files are showing up as new in this PR? They're already on master...

stephenh: Maybe I hadn't pushed out the `master` with your previously-merged PR. Maybe make sure you're rebased on top of the latest `master`?

src/main.ts
```diff
@@ -153,7 +153,7 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     },
     options
   );
-  if (hasAnyTimestamps) {
+  if (hasAnyTimestamps && (options.outputJsonMethods || options.outputEncodeMethods)) {
```
stephenh, line 156: This part looks great, thanks for the quick fix and great catch @fxb !

stephenh:
@alexrodriguez I'm forgetting what happened here, but it looks like this PR / code change is already in master? Can we close this one?

stephenh:
Assuming this is stale/already fixed. Happy to re-open if I'm missing something @alexrodriguez .

PR #54, opened 2020-04-09
Limiting generation of `fromJsonTimestamp` to cases when JSON is requested

alexrodriguez:
Timestamp encoding methods will be generated when `outputJsonMethods` and/or `outputEncodeMethods` options are true. Method `fromJsonTimestamp` will be generated only when `outputJsonMethods` is true.


stephenh:
Thanks!

src/main.ts
```diff
@@ -233,6 +233,25 @@ function addTimestampMethods(file: FileSpec, options: Options): FileSpec {
     secondsCodeLine = 'const seconds = (date.getTime() / 1_000).toString()';
   }
 
+  if (options.outputJsonMethods) {
+    file
```
stephenh, line 237: Er, sorry, this should have been `file = file.addFunction`, which is an unfortunate ts-poet 2.x ism that got copied over from the "thou shalt always use immutable builders" style from the original poet-style JVM library (and went away in ts-poet 3.x). I'll fix on master.

alexrodriguez: Cool. Thanks.

Issue #59, opened 2020-05-03
bug: stream should force observable


ToonvanStrijp:
Heey there, I'm quite new to GRPC and protobuffers. Great library btw! I'm using the nestjs feature and it's working great so far!

However I think when defining a `stream` parameter or return type it should be forced to be of type `Observable` right? Cause a `Promise` can't support a stream of data.

The proto file I have:
```
syntax = "proto3";

import "google/protobuf/empty.proto";

package statistic;

service StatisticService {
  rpc FindOne (StatisticById) returns (Statistic);
  rpc FindMany (stream StatisticById) returns (stream Statistic);
  rpc CreateEmptyRow (EmptyStatisticRow) returns (google.protobuf.Empty);
}

message EmptyStatisticRow {
  int32 id = 1;
  string field = 2;
}

message StatisticById {
  int32 id = 1;
  string field = 2;
}

message Statistic {
  int32 entityId = 1;
  string sales = 2;
  int32 revenue = 3;
  int32 revenueCash = 4;
  int32 revenueDigital = 5;
}
```

As you can see the `FindMany` call takes a `stream` as input and returns a `stream`.

Right now `ts-proto` outputs this function as followed:
```
  findMany(request: StatisticById, metadata?: Metadata): Promise<Statistic>;
```

You can force the return type by providing this option `--ts_proto_opt=returnObservable=true` but this only changed the return type not the parameter type. Also if you set this option to false and the `rpc` call returns a stream it should be forced to the `Observable` type right?

I'm currently looking at the `nestjs` documentation: https://docs.nestjs.com/microservices/grpc#subject-strategy

And as you can see they wrap the input in an `Observable` and the return type also.

# fix
I already implemented a fix and can create a pull request for this. But my main question is if we should do this if the `nestJs` option is true or not.

https://github.com/debugged-software/ts-proto/commit/dd5a808319cf5239b5ec9d3252e59d6731857ad5

stephenh:
Hi @ToonvanStrijp yeah, good call out. I've not actually used streams in protobuf on node; I did awhile ago in some protobuf on JVM but kinda forgot about streaming for node so far. :-) 

I think your initial fix is a good first step...

Since you're an NestJS user, can you extend the `nestjs-simple` integration test that @iangregsondev just added:

https://github.com/stephenh/ts-proto/tree/master/integration/nestjs-simple

To have an additional `.proto` message/method with `stream` in it?

[...]

PR #60, opened 2020-05-05
better nestjs support

ToonvanStrijp:
This pull request is discussed in #59, please see this issue for more context.

Let me know if anything needs to be changed. This pull request will be a breaking change, since we changed the output when the `nestJs=true` option is set.

I've changed/improved the following:
1. I added an `nestjs` section to the `readme` file: https://github.com/debugged-software/ts-proto#nestjs
2. It generated two interfaces, 1 for the client and 1 for the server. (only when `nestjs=true`)
3. I also added a naming convention for the interfaces in `nestjs`.
4. I created two dedicated functions for creating the server interface and the client interface. This way we can easily change the way how we build the interfaces for `nestjs`. (`generateNestjsService` and `generateNestjsServiceClient`).
5. when `nestjs=true` it also generates an `class decorator` for example `HeroServiceControllerMethods` (based on the name of the service). Normally in `nestjs` you would need to add `@GrpcMethod('SERVICE_NAME')` or  `@GrpcStreamMethod('SERVICE_NAME')`. This however would fail in runtime when you accidentally change your service name. But because we generate this `class decorator` for you, you won't have this risk anymore. The only thing you need to do is provide this `decorator` on the `controller` class. The decorator will take care of applying the `@GrpcMethod` and `@GrpcStreamMethod`
```typescript
@Controller('hero')
// Generated decorator that applies all the @GrpcMethod and @GrpcStreamMethod to the right methods
@HeroServiceControllerMethods()
export class HeroController implements HeroServiceController {
...
}
```
6. based on the `.proto` we'll generate a `const` for example `HERO_PACKAGE_NAME` and `HERO_SERVICE_NAME` this way your code breaks if you change your package or service name. (It's safer to have compiler errors than runtime errors)


README.markdown
```diff
@@ -53,6 +53,105 @@ This will generate `*.ts` source files for the given `*.proto` types.
 
 If you want to package these source files into an npm package to distribute to clients, just run `tsc` on them as usual to generate the `.js`/`.d.ts` files, and deploy the output as a regular npm package.
 
+NestJS
```
stephenh, line 56: What do you think about breaking this out as a separate `nestjs.markdown` file, and linking to it from the main readme?

README.markdown
```diff
@@ -53,6 +53,105 @@ This will generate `*.ts` source files for the given `*.proto` types.
...
+}
+```
+
+The controller interface name would be `HeroServiceController`.
+The client interface name would  `HeroServiceClient`.
+
+### implementation
+To implement the typescript file in your `nestjs` project you need to add the `controller` interface to your controller. We also generate a `decorator` for you controller. For example: `HeroServiceControllerMethods`, when you apply this to your controller we add all the method decorators you normally should do but doing it this way is safer.
+
+For the client we simply pass the `client` interface to the `client.getService<?>();`.
```
stephenh, line 88: Ah, I see you've got an example of this down below, maybe throw in "see below" i.e. "...pass the `client` interface to the `client...` (see below)" so readers know it's coming up.

src/main.ts
```diff
@@ -135,7 +142,25 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
...
+    );
+    if (options.nestJs) {
+      // generate nestjs grpc client interface
+      file = file.addInterface(generateNestjsServiceClient(typeMap, fileDesc, sInfo, serviceDesc, options));
+
+      // generate nestjs grpc service controller decorator
+      file = file.addFunction(generateNestjsGrpcServiceMethodsDecorator(serviceDesc, options));
+
+      let serviceConstName =  `${camelToSnake(serviceDesc.name)}_NAME`;
+      if(!serviceDesc.name.toLowerCase().endsWith('service')){
```
stephenh, line 158: Not quite following this? Would it hurt to just always export the const? Even if the user, should their service end with `service` doesn't technically need it, it's there for consistency?

src/main.ts
```diff
@@ -135,7 +142,25 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
...
+        serviceConstName = `${camelToSnake(serviceDesc.name)}_SERVICE_NAME`;
+      }
+
+      file = file.addCode(CodeBlock.empty().add(`export const %L = '%L'`, serviceConstName, serviceDesc.name));
```
stephenh, line 162: Minor but throw a `;` after the `'%'`?

src/main.ts
```diff
@@ -1106,6 +1129,176 @@ function generateServiceClientImpl(
   return client;
 }
 
+function generateNestjsService(
```
stephenh, line 1132: Maybe `generateNestjsControllerInterface`? Or maybe `generateNestjsServiceController`?

README.markdown
iangregsondev: I kind of agree. We don't want to alienate non nestjs users, actually maybe they will see it and migration over AH AH.

I don't think its a bad idea to separate it into its own markdown. Leaves us more control to write whatever needs to be said for Nestjs without disturbing non nestjs users.

I live the idea.

src/main.ts
iangregsondev: Ah I quite like that, @stephenh i think its explained here 

>  Note: Based on the `.proto` we'll generate a `const` for example `HERO_PACKAGE_NAME` and `HERO_SERVICE_NAME` this way your code breaks if you change your package or service name. (It's safer to have compiler errors than runtime errors)

src/main.ts
```diff
@@ -1304,6 +1500,13 @@ function maybeSnakeToCamel(s: string, options: Options): string {
   }
 }
 
+function camelToSnake(s: string): string {
+  console.log('convert to snake', s);
```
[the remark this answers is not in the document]
ToonvanStrijp: Oops! I'm working on a update right now with all the changes ;)

src/main.ts
ToonvanStrijp: @stephenh This way we offer more consistency in the naming. If the user happened to have a service inside their `.proto` named: `hero` then it would end up as `HERO_NAME`. We could do it without forcing the `SERVICE` in here but this also makes me think. If we go for this approach we also should force the `service` in the `controller` and `client` interface name?

@stephenh, @iangregsondev let me know what you guys think is a better approach.

src/main.ts
ToonvanStrijp: But the `add` function automatically adds this ending `;` right? Cause the code that it formats now isn't missing a `;`, and when looking at other parts of the code you're doing it the same. Or I don't get it hehe ;)

src/main.ts
ToonvanStrijp: Agree `generateNestjsServiceController` is a better name for this.

ToonvanStrijp:
@stephenh I fixed some of the changes you requested, and commented on the changes that need more info.

ToonvanStrijp:
@stephenh  @iangregsondev  please review this also: https://github.com/stephenh/ts-proto/pull/60/commits/93a4b28be13fc540c9132d254aa4b9ecfd8bcfa1

src/main.ts
stephenh: Hm, I'm pretty sure `add` doesn't suffix a `;` enough though some of these methods like `beginControlFlow`/`endControlFlow` do.

And the output doesn't have them now:

```
export const HERO_PACKAGE_NAME = 'hero'
export const HERO_SERVICE_NAME = 'HeroService'
```

src/main.ts
stephenh: Ah okay. 

> If we go for this approach we also should force the service in the
> controller and client interface name?

Eh, I could go either way. I think what you have now is fine (i.e. don't force `Service` into the interface names).

Most users probably do have `Service` as a suffix of their service names anyway? Or maybe they prefer something like `HeroRpc` (just making something up, but some other suffix that is "like service") so by assuming "oh you must want `Service` suffixed", we'd be doing the wrong thing.

Better to make less assumptions I think.

In that vein, I'm tempted to nudge the "does name end in service" check here out to be consistent, i.e. maybe they want would `HERO_RPC_NAME` :shrug: .

(I agree having `service Hero` is kind of odd, so I'm assuming most users won't do that.)

src/main.ts
```diff
@@ -1163,7 +1164,9 @@ function generateNestjsService(
     }
 
     // Return observable for interface only configuration, passing returnObservable=true and methodDesc.serverStreaming=true
-    if (options.returnObservable || methodDesc.serverStreaming) {
+    if (isEmptyType(methodDesc.outputType)) {
+      requestFn = requestFn.returns(TypeNames.anyType('void'));
```
stephenh, line 1168: Nice.

src/main.ts
ToonvanStrijp: @stephenh woow you're totally right! My bad! ;) will fix this

src/main.ts
ToonvanStrijp: Either `HeroRpc` or `HeroService` look fine to me but I prefer `HeroService` since it's a `service` inside the proto.

Also I think we should no force the user to much as you said.

So we can resolve this issue and leave it as it is now?

PR #61, opened 2020-05-06
nestjs package format fix

ToonvanStrijp:
Hello @stephenh,

I discovered a little bug inside the `nestjs=true` option when `exporting` the `consts` for the `packages`.

For example:
```
syntax = "proto3";

import "google/protobuf/empty.proto";
```

This would normally generate to this typescript code:
```
export interface Empty {
}

export const GOOGLE.PROTOBUF_PACKAGE_NAME = 'google.protobuf';
```

As you can see the `const` name contains a `.`. And well this shouldn't be there.

This pull request fixes that.

ps. sorry I missed this earlier ;)


stephenh:
Cool, np, I cherry-picked onto master and released as 1.20.1.

PR #66, opened 2020-05-21
Allow Uint8Arrays / Buffers to be passed to decode() directly

jcarver989:
This PR makes a small tweak to the `decode()` method so that it can accept an input type of `Uint8Array | Reader` (vs just `Reader` previously). 

You'll want to view this PR by commit. The 1st commit contains the actual changes. And the 2nd contains all the jitter caused by integration tests updating themselves to reflect the new method signature.

Also, I wasn't sure if your build process / CI auto ticks the package version, so I included a version bump in the PR as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-proto",
-  "version": "1.20.2",
+  "version": "1.21.0",
   "description": "",
   "main": "build/plugin.js",
   "repository": "github:stephenh/ts-proto",
--- a/src/main.ts
+++ b/src/main.ts
@@ -521,12 +521,13 @@ function generateDecode(
 ): FunctionSpec {
   // create the basic function declaration
   let func = FunctionSpec.create('decode')
-    .addParameter('reader', 'Reader@protobufjs/minimal')
+    .addParameter('input', TypeNames.unionType('Uint8Array', 'Reader@protobufjs/minimal'))
     .addParameter('length?', 'number')
     .returns(fullName);
 
   // add the initial end/message
   func = func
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ts-proto",
-  "version": "1.20.2",
+  "version": "1.21.0",
```
jcarver989, line 3: Technically this is a new feature vs a bugfix, so I went with 1.21.0 vs 1.20.3 -- I'm ambivalent though.

src/main.ts
```diff
@@ -521,12 +521,13 @@ function generateDecode(
 ): FunctionSpec {
   // create the basic function declaration
   let func = FunctionSpec.create('decode')
-    .addParameter('reader', 'Reader@protobufjs/minimal')
+    .addParameter('input', TypeNames.unionType('Uint8Array', 'Reader@protobufjs/minimal'))
```
jcarver989, line 524: `message` was already taken, so `input` seemed reasonable.

src/main.ts
```diff
@@ -521,12 +521,13 @@ function generateDecode(
...
     .addParameter('length?', 'number')
     .returns(fullName);
 
   // add the initial end/message
   func = func
+    .addStatement('const reader = input instanceof Uint8Array ? new Reader(input) : input')
```
jcarver989, line 530: In Node, `Buffer` is a sub-class of `Uint8Array`, and the browser doesn't have `Buffer` -- just `Uint8Array`, so this `instanceof` check seems ok to cover those use-cases.

stephenh: Nice. Wouldn't mind that as an in-source comment. I'd forgotten Buffer subclasses Uint8Array.

PR #68, opened 2020-05-21
fix(nestjs): fixes grpc controller decorator

ssilve1989:
First off let me say I am very grateful to see NestJS support here, so thanks! I've been doing similar things in a custom tool alongside your ts-proto package but glad to see it in here now.

This fixes the controller methods decorator. Prior to this fix it was using the wrong decorator for certain types of methods.

Given the following proto that demonstrates all possible combinations of rpcs:
```proto
service GreeterService {
    rpc SayHello (HelloRequest) returns (HelloReply) {}
    rpc SayHelloToAll (stream HelloRequest) returns (stream HelloReplyMessage) {}
    rpc MonitorHellos (MonitorHelloRequest) returns (stream MonitorHelloResponse) {}
    rpc MonitorHellosWithSnapshot (MonitorHelloRequest) returns (stream MonitorHelloResponseOrSnapshot) {}
    rpc BroadcastHellos (stream HelloRequest) returns (HelloCounter) {}
}
```

rpc calls that take in a regular request, regardless of if they return a stream, need to be decorated with `GrpcMethod()` in Nest. 

rpc calls that take in a stream, regardless of what they return should be decorated with `GrpcStreamMethod()`

## Note

There was no yarn.lock in this repository which i believe caused some issues when trying to initialize this repository and i've seen some tests fail as a result of that, probably. I also had to add a tsconfig rule to remove some TS errors I was seeing.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1261,12 +1261,12 @@ function generateNestjsGrpcServiceMethodsDecorator(
   let grpcServiceDecorator = FunctionSpec.create(`${serviceDesc.name}ControllerMethods`).addModifiers(Modifier.EXPORT);
 
   const grpcMethods = serviceDesc.method
-    .filter(m => !m.serverStreaming && !m.clientStreaming)
+    .filter(m => !m.clientStreaming)
     .map(m => `'${options.lowerCaseServiceMethods ? camelCase(m.name) : m.name}'`)
     .join(', ');
 
   const grpcStreamMethods = serviceDesc.method
-    .filter(m => m.serverStreaming || m.clientStreaming)
+    .filter(m => m.clientStreaming)
     .map(m => `'${options.lowerCaseServiceMethods ? camelCase(m.name) : m.name}'`)
     .join(', ');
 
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -1261,12 +1261,12 @@ function generateNestjsGrpcServiceMethodsDecorator(
   let grpcServiceDecorator = FunctionSpec.create(`${serviceDesc.name}ControllerMethods`).addModifiers(Modifier.EXPORT);
 
   const grpcMethods = serviceDesc.method
-    .filter(m => !m.serverStreaming && !m.clientStreaming)
+    .filter(m => !m.clientStreaming)
```
stephenh, line 1264: Yeah, this seems obviously more correct in retrospect.

stephenh: Well, maybe not obvious. :-) But still thanks!

stephenh:
@ssilve1989 thanks! I added a few more methods to the integration test to make sure the generated decorator looks good / doesn't break. Take a look and let me know if anything needs tweaks / you want to flush anything out:

https://github.com/stephenh/ts-proto/commit/6c5ec949ace9b61ca9250e0ac31c5748181b2396

I'm pushing your fix out as 1.21.1.

PR #72, opened 2020-05-27
support bytes as Buffer by Env option

### LINKED ISSUE #65: Support bytes as Buffer (optimal under node)
https://github.com/protobufjs/protobuf.js#usage

protobufjs parses protobuf `bytes` as `Buffer` on Node.js on default behavior.
So there is some incompatibility between imported ts interface created by `protoc-gen-ts_proto` and actual type parsed by protobufjs on node runtime.

I think that we need a option like `--env=node|browser`.

dolsup:
it resolves #65.

- added options.env (browser | node)
- protobuf `bytes` is `Buffer` on node or `Uint8Array` on browsers.

We need to discuss which of 'browser' and 'node' is proper to the default value for options.env.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -54,6 +54,11 @@ export enum LongOption {
   STRING = 'string'
 }
 
+export enum EnvOption {
+  NODE = 'node',
+  BROWSER = 'browser'
+}
+
 export type Options = {
   useContext: boolean;
   snakeToCamel: boolean;
@@ -65,6 +70,7 @@ export type Options = {
   returnObservable: boolean;
   lowerCaseServiceMethods: boolean;
   nestJs: boolean;
+  env: EnvOption;
 };
 
 export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, parameter: string): FileSpec {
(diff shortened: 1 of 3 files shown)


stephenh:
Hey @dolsup I checked this out locally, added a `bytes-node` integration test, and merged this into master. I did have to make one tweak to the output to add `as Buffer`, i.e. in the output like:

```
      switch (tag >>> 3) {
        case 1:
          message.data = reader.bytes() as Buffer;
          break;
```

Let me know if that doesn't seem right to you, but I think it makes sense...

cc @jcarver989 for his thoughts and also FYI about the new `env=node` option (which I actually made default) that will Buffer-ize `data` fields. 

Thanks @dolsup !

jcarver989:
I think the browser/node option and cast makes sense...at least after a cursory look. 

One interesting thing worth pointing out:  is you can generate an `env=browser` package and use it both in the browser and in node.

This is because `Uint8Array` is a superclass of  node's `Buffer`, and so even though the underlying type parsed by protobufjs in node is actually a `Buffer`, the "browser" interface's `Uint8Array` type still works.

stephenh:
Ah yeah, that's a good point. I'll note that in the readme...

Wonder if eventually there should be like a `env=node | browser | both` option where both uses only whats available on both platforms.

stephenh:
Yeah, I think I've changed my mind; I'm going to go ahead and add `both` now as the default, which makes this a non-breaking change, and you'll have to use `env=node` to get the `bytes`-use-`Buffer` output. Thanks @jcarver989 !

stephenh:
Released the `env=both` default+fix as v1.23.0.

dolsup:
Great! Thanks to your efforts ☺️

Issue #74, opened 2020-05-29
Immitate protobuf.js for oneof and non-scalar behavior re: field optionality


philikon:
I don't find ts-proto's generated types particularly helpful because it forces me to always provide all fields (even if I set it to `undefined`), which is especially annoying when you have huge nested `oneof`s.

As an example, consider the following proto:
```proto
message Person {
  string name = 1;
  uint32 age = 2;
  oneof legal_document {
    Passport passport = 3;
    DriversLicense dl = 4;
    GovtIssuedId govt_id  = 5;
  }
}
```

ts-proto translates this to
```ts
export interface Person {
  name: string;
  age: number;
  passport: Passport | undefined;
  dl: DriversLicense | undefined;
  GovtIssuedId: govtId | undefined;
}
```

If I wanted to encode one of these messages, I'd have specify *all* fields:
```ts
Person.encode({
  name: "Bob",
  age: 37,
  passport: undefined,
  dl: undefined,
  govId: undefined,
})
```
That is tedious and you can imagine with large messages, especially nested `oneof`s, it becomes entirely impractical. I'm also debating whether we should need to specify `age` and `name` at this stage. proto3 says that all fields are essentially optional, though `google.protobuf.*Value` is the de-facto way to explicitly declare an optional value.

(Also, I personally would also consider code that explicitly sets a value to `undefined` to be bad form. IMHO, `undefined` is a fallback value that the JS/TS language provides when the thing you're trying to access cannot be found or was not defined. Code that explicitly wants to provide a null value should use `null`. That's my interpretation of JS's built-in nullish types.)

I think pbts actually gets this right. It would translate the above to
```ts
                interface IPerson {
                    name?: (string|null);
                    age?: (number|null);
                    passport?: (Passport|null);
                    dl?: (DriversLicense|null);
                    govtId?: (GovtIssuedId|null);
                }
```
which is the interface that consumed by `Person.encode()`. This means I can write
```ts
Person.encode({name: "Bob"})
```
and it's perfectly valid. We can debate whether `name` and `age` should be optional and nullable (probably not), but all the fields in the `oneof` definitely should be! And for clarity, I would argue that any value that's not a basic type should be optional too, to save tedious typing, having to set them to `null` (or, worse, `undefined`, which as said above, is not a value I'd ever expect application code to have to set, it's something to check for).

I'm happy to help with the implementation, and I'm also happy to hide any changes behind a compiler flag if that's preferable.

stephenh:
[...]

PR #75, opened 2020-05-30
Compiler flag nullableOptionals=true turns message + oneof fields into optional and nullable properties

philikon:


philikon:
Closing in favor of https://github.com/stephenh/ts-proto/pull/76

PR #76, opened 2020-05-31
Compiler flag useOptionals=true turns non-scalar + oneof fields into optional properties

philikon:


philikon:
Closing this in favor of a future PR that will have different flags for non-scalar optionals and better oneof typing.

PR #82, opened 2020-06-06
feat: Simple Objects

cliedeman:
Hi there, thanks for the great library.

I ran into a corner when trying to adapt this for a new project. My issue is combining the generated objects with mobx. When mobx tries to create obserables from the generated types I get the following error:

```
Error: [mobx] The provided value could not be converted into an observable. If you want just create an observable reference to the object use 'observable.box(value)'
```

(An [obserable.box](https://mobx.js.org/refguide/boxed.html#observableboxvalue) is meant for primitives)

I solve this by adding a new option `nullPrototypes` (Better name very welcome) that changes the objects created by ts-proto to not have a custom prototype which fails this check in mobx:

https://github.com/mobxjs/mobx/blob/7f67c871ee2f669850eb7d85374274ede5604f60/src/v5/utils/utils.ts#L88

```
export function isPlainObject(value) {
    if (value === null || typeof value !== "object") return false
    const proto = Object.getPrototypeOf(value)
    return proto === Object.prototype || proto === null
}
```

In a nutshell it replaces this base method init

```
const message = Object.create(baseSimple) as Simple;
```

with this

```
const message = {...baseSimple} as Simple;
```


integration/simple-nullprototypes/simple-test.ts
```diff
@@ -0,0 +1,249 @@
...
+      blobs: []
+    };
+    expect(simple.name).toEqual('asdf');
+  });
+
+  it.only('can decode', () => {
+    const s1: ISimple = {
+      name: 'asdf',
+      age: 1,
+      createdAt: undefined,
```
cliedeman, line 44: I found some strange behaviour here that I don't understand enough about protobuf to explain.

In the original simple test createdAt, thing and the child.type values are all missing.
I am not sure if this is acceptable or a slightly variation in functionality that just needs to be documented in the options

cliedeman: Looks like some of these are fixed by https://github.com/stephenh/ts-proto/commit/7b839834dfa176850132c87322623f5348487704

src/main.ts
```diff
@@ -524,6 +525,14 @@ function visitServices(
   }
 }
 
+function generateObjectInitStatement(func: FunctionSpec, fullName: string, nullPrototypes: boolean) {
+  if (nullPrototypes) {
+    return func.addStatement('const message = {...base%L} as %L', fullName, fullName);
+  } else {
+    return func.addStatement('const message = Object.create(base%L) as %L', fullName, fullName);
```
philikon, line 532: nit: `else` after `return` is unnecessary noise

cliedeman: Fixed

stephenh:
@cliedeman ah interesting, I'm using mobx as well right now (but on a front-end that is using GQL instead of protobuf), so I know what you mean.

Fwiw/in retrospect, I wonder if the `baseMessage` approach is really even necessary.

I thought it was a good idea at first to "avoid a class" and yet "still apply defaults" via the prototype, but once I had to handle lists/maps, which can't use the prototype and need to have copies set on them anyway...eh, why bother?

wdyt? Maybe have this PR just do that instead? I.e. no config option, take out the `baseFoo` objects, and just set default primitives directly on each created object.

cliedeman:
@stephenh that sounds good to me. Is this what you had in mind: https://github.com/stephenh/ts-proto/pull/86

stephenh:
Ah, it's not exactly (I was thinking of assigning each property by hand), but that's better! :-)

Out of curiosity, I thought the github actions were passing/clean; do you know why these are showing up:

image

cliedeman:
@stephenh weird never seen those before and definitetly not something I added. 

Seems to be something new based on the beta tag
Screenshot 2020-06-07 at 21 27 30

README.markdown
```diff
@@ -234,6 +234,8 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
 * With `--ts_proto_opt=returnObservable=true`, the return type of service methods will be `Observable<T>` instead of `Promise<T>`.
 
+* With `--ts_proto_opt=nullPrototypes=true`, a custom prototype will not be used, this will increase memory usage.
```
stephenh, line 237: @cliedeman I think this approach looks good, maybe remove this config flag and the extra `integration/simple-nullprototypes`?

cliedeman: Its done like this in https://github.com/stephenh/ts-proto/pull/86 unless I misunderstood

stephenh: Yep, I think that commit looks good., but doesn't it mean we can remove this readme change? And the new `simple-nullprotobytes` example? Because now all output will work this way and doesn't need the flag.

stephenh: Oh, sorry, different PR. Yep, I get it.

stephenh:
Replaced by #86

PR #83, opened 2020-06-06
Add Github actions

cliedeman:
Sample build: https://github.com/cliedeman/ts-proto/runs/745864535


stephenh:
I'm generally ambivalent re github-actions vs. the existing circleci, but the node v12/v14 strategy thing is cool, so sure sgtm. Thanks!

cliedeman:
Its really powerful. You can even add a matrix on the os and run on windows too

PR #84, opened 2020-06-07
Upgraded prettier and added to CI

cliedeman:


cliedeman:
Hi @stephenh 

This is already in the project

```
{
  "printWidth": 120,
  "bracketSpacing": true,
  "singleQuote": true
}
```

You can see the full notes [here](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev) but the important bits are:

```
After a long and careful consideration, we decided to change the default values for the trailingComma, arrowParens, and endOfLine options
```

I paused on this because I am not sure how to handle the integration folder :/. The generated code does not conform to prettier which makes thinkgs a bit funky. I was tossing around the idea of adding a prettier flag to ts-poet to optionally run prettier on the output.

Otherwise we can just exclude the integration folder from prettier.

FYI: There is no way to mark an entire file as ignore - https://prettier.io/docs/en/ignore.html

stephenh:
> adding a prettier flag to ts-poet to optionally run prettier on the output

Ah yeah, fwiw the latest version of ts-poet (v3, which is a large breaking change from v2, which is why ts-proto is still on v2) does automatically run prettier on its output (and auto-detects `.prettierrc`s files, although we might have to copy/paste a prettierrc into each integration test for that to work).

 >  "singleQuote": true

Ah shoot! I forgot about that. Well. I'm tempted to flip that at some point, but can be a different PR, so what you're doing now lgtm.

stephenh:
@cliedeman in retrospect, let's not run prettier on the `integration` directory. I think that will mess up the workflow a bit for seeing "did my WIP ts-proto update correctly/incorrectly change the integration output" b/c then you'd have to both run the ts-proto codegen + prettier separately. Sound good?

cliedeman:
@stephenh sounds good to me. Updated PR to only change src and test

stephenh:
Sweet, thanks!

PR #86, opened 2020-06-07
Feat/remove prototype

cliedeman:
Replace Object.create with a simple spread operator. 

The interesting changes are in here: https://github.com/stephenh/ts-proto/pull/86/commits/2c01cbea398d0a915f099d8b74380e9d19123b5f

Slight bahviour change with some previously missing fields now being undefined

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -527,7 +527,7 @@ function generateDecode(
   func = func
     .addStatement('const reader = input instanceof Uint8Array ? new Reader(input) : input')
     .addStatement('let end = length === undefined ? reader.len : reader.pos + length')
-    .addStatement('const message = {...base%L} as %L', fullName, fullName);
+    .addStatement('const message = { ...base%L } as %L', fullName, fullName);
 
   // initialize all lists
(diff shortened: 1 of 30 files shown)


src/main.ts
```diff
@@ -541,7 +541,7 @@ function generateDecode(
   func = func
     .addStatement('const reader = input instanceof Uint8Array ? new Reader(input) : input')
     .addStatement('let end = length === undefined ? reader.len : reader.pos + length')
-    .addStatement('const message = Object.create(base%L) as %L', fullName, fullName);
+    .addStatement('const message = {...base%L} as %L', fullName, fullName);
```
stephenh, line 544: This is pretty minor, but can you change this to `{ ...base%L }` i.e. add parens after/before the parens. That is what prettier will eventually make this when we have prettier formatted output.

cliedeman: Fixed 👍 . Please squash this when you merge. Some bash script changes snuck into my commits

stephenh:
Released as v1.24.0. Thanks!

PR #87, opened 2020-06-07
Fix prepare step causing false github warnings 

stephenh:


.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ jobs:
             - v1-dependencies-{{ checksum "package.json" }}
             - v1-dependencies-
 
-      - run: yarn install || echo "ignore failure"
+      - run: yarn install
```
stephenh, line 14: @cliedeman fwiw this is where those false-positive failures were coming from; the github actions was picking up the stdout/stderr from this `yarn install` running `prepare` which was doing a `yarn build` "before we really wanted".

I was trying to use `prepare` as a "before publish" but _not_ "after install", which looks like is not possible? So just taking out `prepare` for now and will use a more manual `./publish.sh`.

cliedeman: Ah super weird. Did not know it did that. 

`prepublishOnly` might do what you want

stephenh: Ah I think so, thanks for pointing out `prepublishOnly`.

PR #88, opened 2020-06-07
Compiler flag useOptionals=true turns non-scalar fields into optional properties

philikon:
Also includes a change not to generate default entries in base objects when the value would be `undefined` anyway as a separate commit. Probably best to review each commit individually.


src/types.ts
```diff
@@ -318,13 +318,19 @@ export function valueTypeName(field: FieldDescriptorProto): TypeName {
...
   keepValueType: boolean = false,
   repeated: boolean = false
 ): TypeName {
-  // Watch for the wrapper types `.google.protobuf.StringValue` and map to `string | undefined`
+  // Watch for the wrapper types `.google.protobuf.*Value`. If we're mapping
+  // them to basic built-in types, we union the type with undefined to
+  // indicate the value is optional. Exceptions:
+  // - If the field is repeated, values cannot be undefined.
+  // - If useOptionals=true, all non-scalar types are already optional
+  //   properties, so there's no need for that union.
```
stephenh, line 330: I really appreciate these in-source comments, thanks!

src/types.ts
```diff
@@ -318,13 +318,19 @@ export function valueTypeName(field: FieldDescriptorProto): TypeName {
 export function messageToTypeName(
   typeMap: TypeMap,
   protoType: string,
+  options: (Options | null) = null,
```
stephenh, line 321: Hrm, can we make `options` required? I'd be worried about a caller getting back the wrong type name b/c they weren't passing in options... If this is a huge hassle, an in-source todo/separate follow up would be fine.

philikon: We can. The only callsites of `messageToTypeName` that don't currently have access to `options` are `requestType` and `responseType`. We could pass `options` into those, though, I think. I was just shying away from a bigger change.

philikon:
Rebased on top of master, fixed linter errors, and added another commit that makes `options` a required param in `messageToTypeName`.

philikon:
I'm not sure I like the two optional bools at the end of the `messageToTypeName` parameter list:
```
  keepValueType: boolean = false,
  repeated: boolean = false
```
IMHO bool params are an anti-pattern anyway because a call like `messageToTypeName(typeMap, field.typeName, options, true, false)` gives you little insight into what `true` and `false` are doing.

What do you think about taking an object there instead? Something like this:
```ts
export function messageToTypeName(
  typeMap: TypeMap,
  protoType: string,
  options: Options,
  typeOptions: {keepValueType?: boolean, repeated?: boolean}
): TypeName {
```
This would translate the above call to `messageToTypeName(typeMap, field.typeName, options, {keepValueType: true, repeated: false})` (or you could even leave out `repeated`).

stephenh:
Sure, the object type sgtm.

philikon:
> Sure, the object type sgtm.

Ok I'll send a separate PR for that some time since this PR is ready for review.

Issue #77, opened 2020-06-01
Angular: Build error "Cannot find name 'Buffer'" when using a bytes field.


timostamm:
First, thank you for this library, @stephenh. It has been a while since I looked around for a Typescript protobuf library and I was very happy to discover ts-proto last week. 

However, I did run into an issue with angular. The plugin generates the following code when a bytes field is present: 

```ts
interface WindowBase64 {
  atob(b64: string): string;
  btoa(bin: string): string;
}

const windowBase64 = (globalThis as unknown as WindowBase64);
const atob = windowBase64.atob || ((b64: string) => Buffer.from(b64, 'base64').toString('binary'));
const btoa = windowBase64.btoa || ((bin: string) => Buffer.from(bin, 'binary').toString('base64'));
```

The Angular compiler fails the build with the following error: 

```
error TS2591: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.
const atob = windowBase64.atob || ((b64: string) => Buffer.from(b64, 'base64').toString('binary'));
                                                    ~~~~~~
```

I don't think I should follow the compiler`s  suggestion to install node's type definitions because that would pollute the application with types from node that it will never be able to use. 

The following replacement fixes the issue:

```ts
interface WindowBase64 {
  atob(b64: string): string;
  btoa(bin: string): string;
}
declare class Buffer {
  static from(str: string, encoding?: 'base64'|'binary'): Buffer;
  toString(encoding:'base64'|'binary'):string;
}
const windowBase64 = (typeof globalThis !== 'undefined' ? globalThis : this) as unknown as WindowBase64;
const atob = windowBase64.atob || ((b64: string) => Buffer.from(b64, 'base64').toString('binary'));
const btoa = windowBase64.btoa || ((bin: string) => Buffer.from(bin, 'binary').toString('base64'));
```

Jest tests pass and the angular compiler is happy. As an added bonus this adds support for Edge 18 and iOS 12 (they don't know `globalThis`). 

More compact version - maybe better suited for generated code:
```ts
declare const global: any;
const globalObj = typeof globalThis !== 'undefined' ? globalThis : (typeof global !== 'undefined' ? global : this) as any;
const atob = globalObj.atob || ((b64: string) => globalObj.Buffer.from(b64, 'base64').toString('binary'));
[...]

PR #92, opened 2020-06-15
Adding simple angular example.

timostamm:
This PR adds a simple angular example in `integration/angular/` in response to #77.

To get this running:

```shell
cd integration

# generate bin for added .proto in integration/angular
sh update-bins.sh

# generate code for added .proto
sh codegen.sh

# install angular and all its dependencies
cd angular
npm install

# run the angular tests
# this should open a browser window that executes the tests
npm run test
```

This example is identical to a standard new angular app (scaffolded by `ng new`), with the following additions:
1. added `long` package as dependency
2. added workaround in `src/main.ts` to let protobufjs use long
3. added `src/app/simple-message.spec.ts` with a basic unit test for the generated ts-proto message

The build error "Cannot find name 'Buffer'" does not surface in this example because no `bytes` field is used as I don't want to commit a failing test.

I think this might rather serve as an example than an integration test, at least in the state it is right now. Please discuss.


stephenh:
Cool! Agreed this will be great for reproducing issues even if its not part of CI yet. Thanks!

stephenh:
(Wrong button. :-)

PR #95, opened 2020-06-19
Compiler flag oneof=unions generates type unions for oneof fields instead of individual properties

philikon:


philikon:
I can't see what the failure is, circleci wants me to log in. :(

cliedeman:
```
Summary of all failing tests
 
 FAIL  integration/simple/simple-oneofs-test.ts
 
  ● Test suite failed to run
 

 
    ENOMEM: not enough memory, read
 

 
      at ScriptTransformer._getTransformer (node_modules/@jest/transform/build/ScriptTransformer.js:355:19)
 
      at ScriptTransformer.transformSource (node_modules/@jest/transform/build/ScriptTransformer.js:444:28)
 

 
 FAIL  integration/bytes-as-base64/bytes-as-base64-test.ts
 
  ● Test suite failed to run
 

 
    ENOMEM: not enough memory, read
 

 
      at Object.<anonymous> (node_modules/bs-logger/dist/logger/index.js:36:15)
 
```

A transient failure

@philikon This looks great. Will give it a try on my project

cliedeman:
@philikon one thing I noticed

```
    PleaseChoose.fromPartial({
      name: 'Debbie',
      age: 37,
      choice: { aBool: true, aString: 'a' }, // Valid
      choice: { $case: 'aBool', aBool: true, aString: 'a' }, // Invalid
      eitherOr: { or: 'perhaps not' },
    });
```

the From Partial allows the user to omit the case. 

This seems to fix it. but here be dragons...
```
type DeepPartial<T> = T extends Builtin
  ? T
  : T extends Array<infer U>
  ? Array<DeepPartial<U>>
  : T extends ReadonlyArray<infer U>
  ? ReadonlyArray<DeepPartial<U>>
  : T extends { $case: string }
  // Try to make $case mandatory
  ? { [K in keyof Omit<T, '$case'>]?: DeepPartial<T[K]> } & { $case: T['$case'] }
  : T extends {}
  ? { [K in keyof T]?: DeepPartial<T[K]> }
  : Partial<T>;
```

This forces the user to allows provide a `$case`

cliedeman:
A more conceptual issue:

Proto
```
  oneof choice {

    // Use this if you want a number. Numbers are great. Who doesn't
    // like them?
    double a_number = 2;

    // Use this if you want a string. Strings are also nice. Not as
    // nice as numbers, but what are you going to do...
    string a_string = 3;

    Submessage a_message = 4;

    // We also added a bool option! This was added after the 'age'
    // field, so it has a higher number.
    bool a_bool = 6;

    bytes buncha_bytes = 10;
  }
```

Generated Code
```
  choice?:
    | { $case: 'aNumber'; aNumber: number }
    | { $case: 'aString'; aString: string }
    | { $case: 'aMessage'; aMessage: PleaseChoose_Submessage | undefined }
    | { $case: 'aBool'; aBool: boolean }
    | { $case: 'bunchaBytes'; bunchaBytes: Uint8Array };
```

I think that the generated code should actually be

```
  choice?:
    | { $case: 'aNumber'; aNumber: number }
    | { $case: 'aString'; aString: string }
    | { $case: 'aMessage'; aMessage: PleaseChoose_Submessage /* undefined removed */ }
    | { $case: 'aBool'; aBool: boolean }
    | { $case: 'bunchaBytes'; bunchaBytes: Uint8Array };
```

my reasoning is that these 2 cases are equivalent from a protobuf point of view
```
{ choice: { '$case': 'aMessage', aMessage: undefined }}
{ choice: undefined }
```

This also allows us to remove some [tedious](https://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgAoBsJwM4QMIAWA9kbgPoDKArgEYC2E22cA5igN4BQyyIcDALmTYwUUCwDcnAL6dOoSLEQoA8gAcwwIn3TIuPBMWBIA-AO49kAHz3IAJAhwQhAcjgA5KnRrQXE5B5ePlBCIEHQyLKW1rYOTq5wFKLifgFJYiAsQiIZLJEWPDbs9o64CQCyjMxsqXCVTKzOaJhOhCTk1PRVjTFUIAAmEDCgEP350UUl8chuAEIk6LXzROhCNAtYIOOWk3FlMzR9hnCzAJ6Q2KmHIMdnF0IAqgoAHACCUFBwp5FSspwwR002mQaigRCQTAAogA3CDgAAU-TgYDgQnUQJ0AEo9AVkMAYMhEci4AA6QxEYwQbH6aKWbAAd2AYEMhKRKLJRiQJL2VJxtP5yFKKDcnm8viENIFAqgEDAVCgWzZpPJlJJgTFUCkUuiUW1BicM0SyUyLgluL1PBlcoVyCVHIpXKNuS1Ft1eqFhvq1Qgpr5FuiAHoA8gABJwWHIMBEXhUdC6QwQBAAa2QBGgEHNFvxrOJ9tVdW6bGpmf9ltl8sVuZVjq9jRdpe2DatFZjcfr2rd2o9cwWvslpebNrt1YgauW6HbUs7Uu711u50YfZL2sHlfZI5Jc4IJwX2EnAunlk7f2iq9bE5knCAA) if statements

philikon:
> the From Partial allows the user to omit the case.

Yes, I think that's the whole point of `fromPartial` yes? :)

> I think that the generated code should actually be
>`    | { $case: 'aMessage'; aMessage: PleaseChoose_Submessage /* undefined removed */ }

Hmm, I need to think about this some more, but I think you might be right. Either you chose the `aMessage` case and then you need to provide the value, or you didn't want to provide a value at all, but then there's no point to provide a `$case`. Yeah, I think that makes sense, but before I go back into the code, I might want @stephenh and @timostamm to weigh in.

philikon:
(Btw, in your example you can do `switch (data.choice?.$case)` and avoid the outermost `if` that way.)

cliedeman:
> Yes, I think that's the whole point of fromPartial yes? :)

Definitely. I guess its just a bit of a footgun because erroneous code would compile. But this was not the goal of the PR I think, because the existing code has the same issue that you attempt to set multiple oneOf fields.

I think a runtime exception in this case might be good enough? I tried to spin up a comparative example in go but the generation style prevents settings 2 cases of a one of field.

> but then there's no point to provide a $case

I almost came to the same conclusion but then there is no clean way to [narrow](https://www.typescriptlang.org/play/?ssl=20&ssc=14&pln=18&pc=10#code/JYOwLgpgTgZghgYwgAgAoBsJwM4QMIAWA9kbgPoDKArgEYC2E22cA5igN4BQyyIcDALmTYwUUCwDcnAL6dOoSLEQoA8gAcwwIn3TIuPBMWBIA-AO49kAHz3I4AOSp0a0ISCcuoyWZeu24FKLiQiJiICzeFjw27HYAsozMbEIYWLiEJOTU9ImsKDZUIAAmEDCgEEWRvjF2AEIk6EI0DVggVZY1NIWGcLUAnpDYQgCqCgAcAIJQUHB93lKynDDdmtrIalBESEwAogBuEOAAFEVwYHBC6qs6AJR6UcjAMMgnZ3AAdIZExhB3+r6WJ4vU7nT5GJDvOAJJh5P4PAGWKAQMBUKBtEEfL4-SHQpIQKQIng+SyLXxIlFo3hUdDoBacIA) the type

> switch (data.choice?.$case)

Thanks :), I was being intentionally pedantic to make sure typescript was doing the right thing.

timostamm:
Wow, looks good, @philikon :)

```ts
{ choice: { '$case': 'aMessage', aMessage: undefined }}
```

I think `$case` should only point to `aMessage` when it has a value. But hey, this is miles better than before. The user should be well aware now that this is a oneof construct. Before, you would have to look at the .proto to notice. 

Regarding fromPartial(): I think @cliedeman is on the right track. The following seems to work out for me:

```ts
export type PartialMessage<T> =
      T extends (Date | Uint8Array | Long | boolean | string | number) ?
        T
    : T extends Array<infer U> ?
        Array<PartialMessage<U>>
    : T extends ReadonlyArray<infer U> ?
        ReadonlyArray<PartialMessage<U>>
    : T extends { oneofKind: string } ?
        T
    : T extends { oneofKind: undefined } ?
        T
    : T extends {} ?
        { [K in keyof T]?: PartialMessage<T[K]> }
    : Partial<T>;
```

You can read it like a if, else if, ... else  statement.

Don't think that this is a drop-in replacement, though. Have been working on yet another typescript plugin... Using the [compiler api](https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API) to generate .ts from .proto. Seems to work alright so far.

philikon:
I think I understand the `fromPartial` concern now. @cliedeman, your example illustrates it perfectly and I think your proposed solution

```
T extends { $case: string }
  ? { [K in keyof Omit<T, '$case'>]?: DeepPartial<T[K]> } & { $case: T['$case'] }
```

works great. I think it's fine to require that if `choice` is filled out, you *have to* provide `$case`, but you *don't* have to provide anything else. I'll update the PR accordingly.

I'll also remove the optionality of message payloads inside the `oneof` as discussed.

philikon:
@stephenh any chance you could look at this over the next few days?

philikon:
Ping?

stephenh:
@philikon yep, sorry, thanks for the patience; I spent my free cycles over the weekend hacking on something else. Will get to this this week.

philikon:
Awesome, thanks

src/types.ts
```diff
@@ -359,22 +359,34 @@ export function toTypeName(
...
+  // By default (useOptionals=false, oneof=properties), non-scalar fields
+  // outside oneofs and all fields within a oneof clause need to be unioned
+  // with `undefined` to indicate the value is optional.
+  //
+  // When useOptionals=true, non-scalar fields are translated to optional
+  // properties, so no need for the union with `undefined` here.
+  //
+  // When oneof=unions, we generate a single property for the entire `oneof`
+  // clause, spelling each option out inside a large type union. No need for
+  // union with `undefined` here, either.
```
stephenh, line 382: Great comment, thanks!

src/main.ts
```diff
@@ -273,7 +280,13 @@ function base64FromBytes(arr: Uint8Array): string {
   );
 }
 
-function addDeepPartialType(file: FileSpec): FileSpec {
+function addDeepPartialType(file: FileSpec, options: Options): FileSpec {
+  let oneofCase = '';
+  if (options.oneof === OneofOption.UNIONS) {
+    oneofCase = `
+  : T extends { $case: string }
+  ? { [K in keyof Omit<T, '$case'>]?: DeepPartial<T[K]> } & { $case: T['$case'] }`;
```
stephenh, line 288: Nice

src/main.ts
```diff
@@ -436,6 +465,52 @@ function generateInterfaceDeclaration(
...
+  );
+  let prop = PropertySpec.create(
+    maybeSnakeToCamel(messageDesc.oneofDecl[oneofIndex].name, options),
+    unionType,
+    true // optional
+  );
+
+  // Ideally we'd put the comments for each oneof field next to the anonymous
+  // type we've created in the type union above, but ts-poet currently lacks
+  // that ability. For now just concatenate all comments into one big one.
```
stephenh, line 496: Thanks again for the in-source comment.

src/main.ts
```diff
@@ -604,6 +679,9 @@ function generateDecode(
           .addStatement(`message.%L.push(%L)`, fieldName, readSnippet)
           .endControlFlow();
       }
+    } else if (isWithinOneOf(field) && options.oneof === OneofOption.UNIONS) {
+      let oneofName = maybeSnakeToCamel(messageDesc.oneofDecl[field.oneofIndex].name, options);
+      func = func.addStatement(`message.%L = {$case: '%L', %L: %L}`, oneofName, fieldName, fieldName, readSnippet);
```
stephenh, line 684: Huh, that wasn't too bad. Nice.

philikon:
Woot! Thank you!

PR #104, opened 2020-07-23
enable rest parameter to use decorators in nestjs

ToonvanStrijp:
I would like this pull request to be merged soon :) by adding the `addRestParameter=true` option you can use decorators while still benefiting the strong typed features `ts-proto` offers.

```
const CurrentUser = createParamDecorator(
  (data: unknown, ctx: ExecutionContext) => {
    return {id: 1, name: 'test'};
  },
);

@Controller('hero')
@HeroServiceControllerMethods()
export class HeroController implements HeroServiceController {

  findCurrentUser(request: Empty, @CurrentUser() user: User): User | Promise<User> | Observable<User> {
    return user;
  }  
}
```


ToonvanStrijp:
@stephenh Could you please take a look at this? It doesn't introduce any type of breaking change but it would be useful for all the nestjs users! :D

src/utils.ts
```diff
@@ -48,6 +48,7 @@ export function defaultOptions(): Options {
     outputClientImpl: true,
     returnObservable: false,
     addGrpcMetadata: false,
+    addRestParameter: false,
```
stephenh, line 51: Maybe rename this to `addNestjsRestParameter`? Since it seems really specific to nestjs's controllers/clients.

src/main.ts
```diff
@@ -1360,7 +1367,10 @@ function generateNestjsServiceClient(
 
     // Use metadata as last argument for interface only configuration
     if (options.addGrpcMetadata) {
-      requestFn = requestFn.addParameter('metadata?', 'Metadata@grpc');
+      requestFn = requestFn.addParameter(options.addRestParameter ? 'metadata' : 'metadata?', 'Metadata@grpc');
+    }
+    if (options.addRestParameter) {
+      requestFn = requestFn.addParameter('...rest', 'any');
```
stephenh, line 1373: I was thinking it'd be nice if could directly be generated as `user: User` here, but I get that these annotations are very specifically something that user adds one-by-one-sy to each call site as needed. So the `any` makes sense.

ToonvanStrijp: @stephenh I don't think we wanna generate this directly, cause it needs to be dynamic that's why the `...rest: any` comes in handy.

src/utils.ts
ToonvanStrijp: Totally agree! Changed it!

ToonvanStrijp:
@stephenh Ready to release ;) or do I need to change the version number in the `package.json`?

stephenh:
Oh right, I forgot releases aren't 100% automated yet. Yep, I'll merge and release.

stephenh:
Released as 1.27.1.

PR #105, opened 2020-07-23
Fix generated toJSON when an enum is in a oneOf

gyanta:
Hello! I found that in oneOf=properties mode the generated toJSON does not compile if the enum is in a oneOf. 
In oneOf=unions mode the code compiles but if the value is not present, it serializes as "false". This PR fixes both cases.


stephenh:
Looks great, thanks for the PR!

PR #106, opened 2020-08-01
Add support for generating service client implementations with JSON serialization

isherman:
First pass at https://github.com/stephenh/ts-proto/issues/7.

Doesn't include support for Batching yet. Any other features this interacts with that I should be aware of?

Thoughts on this approach?


README.markdown
```diff
@@ -248,6 +248,8 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
 - With `--ts_proto_opt=outputClientImpl=false`, the client implementations, i.e. `FooServiceClientImpl`, that implement the client-side (currently Twirp-only) RPC interfaces will not be output.
 
+- With `--ts_proto_opt=outputJsonClientImpl=true`, an additional client implementation will be output for each service, i.e. `FooServiceJsonClientImpl`, that uses JSON serialization. Has no effect if `outputClientImpl=false`.
```
stephenh, line 251: What do you think about doing something like `outputClientImpl = false | true | twirp | json`? I.e. false is off, true is a legacy value that means twirp, and json would be your new code?

I'm kinda thinking that shops will generally have a single preferred client? 

I dunno, maybe not I guess?

Maybe this is better.

Maybe `outputJsonClient=true/false`? I might deprecate the existing `outputClientImpl` and rename it to `outputTwirpClient`...

integration/simple-service/simple-service-test.ts
```diff
@@ -0,0 +1,69 @@
...
+  oldStates: [StateEnum.ON, StateEnum.OFF],
+  createdAt: new Date('1970-01-01T00:00:00.000Z'),
+  thing: undefined,
+  blobs: [],
+};
+
+describe('simple service', () => {
+  it('can echo with protobuf serialization', async () => {
+    const echoRpc = {
+      request: (_service: string, _method: string, data: Uint8Array) => {
```
stephenh, line 24: Ah sure, that makes sense. How about using `jest.fn`s though so we can assert the right values are passed? I.e. `expect(...).toBeCalledWith`?

integration/simple-service/simple-service-test.ts
```diff
@@ -0,0 +1,69 @@
...
+        return Promise.resolve(data);
+      },
+      requestJson: (_service: string, _method: string, data: unknown) => {
```
stephenh, line 27: I was initially confused by this `Rpc` having to have both `request` and `requestJson` ... kinda thinking we should just generate separate `Rpc` interfaces, i.e. the current `Rpc` stays as is and becomes `BinaryRpc` and the new one would be a `JsonRpc`?

src/main.ts
```diff
@@ -1596,22 +1611,29 @@ function generateCachingRpcMethod(
...
 function generateRpcType(options: Options): InterfaceSpec {
-  const data = TypeNames.anyType('Uint8Array');
-  let fn = FunctionSpec.create('request');
+  let rpc = InterfaceSpec.create('Rpc');
+  if (options.useContext) {
+    rpc = rpc.addTypeVariable(TypeNames.typeVariable('Context'));
+  }
+  rpc = rpc.addFunction(generateRpcRequestFn({ ...options, outputJsonClientImpl: false }));
+  if (options.outputJsonClientImpl) {
+    rpc = rpc.addFunction(generateRpcRequestFn(options));
```
stephenh, line 1620: I.e. here maybe just make a new `JsonRpc` that is separate from the existing `Rpc` type...

isherman:
Closing as stale, and in favor of the discussion about a more general approach happening here: https://github.com/stephenh/ts-proto/pull/106

Issue #94, opened 2020-06-18
useContext option seems not working. 


dyllanwli:
It seems `useContext` option is not working on latest version (1.25.0). Instead, set `context=true` works can enable context but the `DataLoaders` interface cannot be defined automatically like previous version:
```
interface DataLoaders {

  getDataLoader<T>(identifier: string, constructorFn: () => T): T;

}
```

my ts_proto_opt is `useContext=true,returnObservable=true,nestJs=true`

PR #107, opened 2020-08-09
Fix DataLoader interface not being added

djedlajn:
As per #94 which is closed. `useContext` didn't seem to work properly because it was not adding DataLoader interface. This makes sure that if flag is passed it will properly generate and insert interface into generated output.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -184,9 +184,6 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
 
   if (options.outputClientImpl && fileDesc.service.length > 0) {
     file = file.addInterface(generateRpcType(options));
-    if (options.useContext) {
-      file = file.addInterface(generateDataLoadersType());
-    }
   }
 
   if (options.useContext) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -189,6 +189,10 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     }
   }
 
+  if (options.useContext) {
+    file = file.addInterface(generateDataLoadersType());
+  }
+
   let hasAnyTimestamps = false;
   visit(
     fileDesc,


src/main.ts
```diff
@@ -189,6 +189,10 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     }
   }
 
+  if (options.useContext) {
+    file = file.addInterface(generateDataLoadersType());
+  }
```
stephenh, line 194: Hey, so this is also done a few lines above this:

```typescript
  if (options.outputClientImpl && fileDesc.service.length > 0) {
    file = file.addInterface(generateRpcType(options));
    if (options.useContext) {
      file = file.addInterface(generateDataLoadersType());
    }
  }
```

That in theory already does this.

Is there part of that `outputClientImpl && fileDesc.service.length` condition that is wrong for your setup? Do you have `outputClientImpl=true` set?

djedlajn: Hi, there is no issue with code above it’s just that with my particular needs. I don’t need the output client impl flag. Maybe I’m missing something but if I don’t want client-related code I can’t have DataLoaders generates properly. So I think it's more nuanced edge case that was not covered initially.

The following command fails to generate the DataLoaders interface prior to code change.

`protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto --ts_proto_opt=outputEncodeMethods=true,useEnumNames=false,asClass=false,outputJsonMethods=true,context=true,useContext=true,nestJs=true,outputClientImpl=false`

stephenh: > Maybe I’m missing something but if I don’t want client-related code I can’t have DataLoaders generates properly

Yeah, that is currently the case, which yeah we can pretty easily change.

Although, sanity checking my understanding here, if you have `outputClientImpl=false`, does anything in the ts-proto generated code even need / reference `DataLoaders`?

I'm ~80% sure that only the client impl code uses the `DataLoaders` stuff, so I'm just wondering/curious what you're using `DataLoaders` for? (Or am I forgetting something about ts-protos output?)

djedlajn: Correct, creating client impl is another step in the build process I'm currently having. You can think of it as implementation only being used by nestjs i.e and the other part of it is generating client code that gets bundled into the separate combined library. So if I use context which I need for the client to be able to tackle N+1 and want only the implementation part of it without the client code I can't have that currently. So short story is I need the client with data loader capability but not at that point in time.

Like I said its nuanced and not wrong approach by no means for you too only bundle data loaders if there is the flag to output the client impl too. Which I might assume is 90% of the time.

stephenh: Cool, that does sound nuanced :-) and sure sounds fine.

Can you remove the `addInterface(generateDataLoadersType())` from the `if (options.outputClientImpl && fileDesc.service.length > 0) {` condition right above this, as part of your PR?

As-is, `DataLoaders` with your change is being added twice if both `outputClientImpl` and `useContext` are set.

So, small tweak, but then will merge + release.

djedlajn: Hi, committed a new change.

PR #111, opened 2020-08-14
Add a karma test.

stephenh:
Attempting to get a "reproduce bugs that happen in the browser" setup.

### DIFF
--- a/integration/simple-long/karma.conf.js
+++ b/integration/simple-long/karma.conf.js
@@ -1,0 +1,18 @@
+module.exports = function (config) {
+  config.set({
+    basePath: '',
+    frameworks: ['jasmine', 'karma-typescript'],
+    files: ['**/*.ts', '*.js'],
+    preprocessors: {
+      '**/*.ts': 'karma-typescript',
+    },
+    reporters: ['progress', 'karma-typescript'],
+    port: 9876,
+    colors: true,
+    logLevel: config.LOG_INFO,
+    autoWatch: true,
+    browsers: ['Chrome'],
+    singleRun: false, // Continuous Integration mode, if true, Karma captures browsers, runs the tests and exits
+    concurrency: Infinity,
+  });
+};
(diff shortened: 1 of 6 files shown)


integration/simple-long/karma.conf.js
```diff
@@ -0,0 +1,18 @@
+module.exports = function (config) {
```
stephenh, line 1: Hey @timostamm / @philikon / anyone else looking for some help;  I'm trying to get this karma thing setup so that I can reproduce / TDD ts-proto bugs that happen on the client-side.

Seems like this "shouldn't be that hard", so I setup `karma-typescript` and am trying to run:

```
../../node_modules/.bin/karma start karma.conf.js
```

But keep getting ` Uncaught Error: Can't find ./pbjs`.

`pbjs.js` is a JS file, so I thought maybe `karma-typescript` only does `.ts` files, so I'd need "something else" for regular JS files.

Given that assumption, I've fought with setting up both `karma-commonjs` and `karma-browserify` and couldn't get either to work (I end up getting various other errors that basically amount to "cannot load everything you actually need to run").

Is there some obvious thing I'm missing that would make this work?

Thanks!

stephenh: I've got JS files working / being picked up now, so making some progress. Now it's not finding any tests to run...

timostamm: Hi @stephenh

sounds like [this issue](https://github.com/monounity/karma-typescript/issues/159). But if I try using "commonjs", it only leads further into the rabbit hole :/

Maybe it is an option for you to generate some test fixtures ahead of time?  
For example, I am using a `.ts` file like this: 
```typescript
/**
 * This data is generated using the `spec.LongsMessage`
 * using Grpc.AspNetCore v2.30.0 (C#).
 *
 * Each field has the appropriate minimum or maximum
 * value of the native type.
 */
export const msg_longs_bytes = new Uint8Array([17, 255, ...]);
```

Works fine for testing my decode and encode functions...

Issue #78, opened 2020-06-02
ts-proto expects Reader.int64() to always return Long, but protobufjs may return number


timostamm:
When protobufjs is used with a bundler, its Reader.int64() returns an unsafe `number` instead of `Long`. 

Can we add a "Gotcha: Module bundlers and long.js" section to the readme?

The following workaround forces protobufjs to use long.js:

```ts
import * as protobuf from 'protobufjs/minimal';
import * as Long from 'long';
protobuf.util.Long = Long;
protobuf.configure();
```
It should be included in the app´s bootstrap code, `main.ts` for angular.

This is required for every `forceLong` option. `forceLong=string` may not throw errors, but it will probably silently overflow the value.

stephenh:
Hm. Great find/point @timostamm . Fwiw ts-proto working in the browser is almost a complete accident given I never focused on it initially. But I love that people are using it there.

This is a large ask, but could you make a PR that adds an `./integration-test` that somehow replicates the issue? I know right now all of the integration-tests run their unit tests in jest on node, but that's exactly why this sort of stuff isn't caught.

I'm technically somewhat-capable in the frontend ecosystem, but most/all of the react tests that I use on a day-to-day basis still run in node with jsdom, and I don't think I've setup a "no really bundle everything to `.js` and run the tests against that".

Is that something you could setup? And then reproduce this?

I'm also wondering if the ^ incantation you have could be something that `ts-proto` just does on initialization? Like when the module is imported, a) detect if this is the browser, b) "know" what `forceLong` option is being used, and c) do this `protobuf.util.Long = Long` automatically?

If you think that's a good idea, that's what I like to have that `integration-test/longs-in-browser` test to cover/make sure it works.

Thanks!

stephenh:
@timostamm so, would it be useful to do something like:

* Add a global variable like `global.tsProtoHasInitializedProtobuf`
* In any file that ts-proto generates, if `forceLong` is on, include that snippet you have
  * The snippet will only be run if `global.tsProtoHasInitializedProtobuf` is unset, and then will set it, so that we only do it once even if you have multiple files generated by ts-proto
* (Optional) Add a `configureProtobuf` config flag which defaults `true` but can be set `false` (if for some reason projects don't want this "we'll just call configure for you" behavior)

If that makes sense to you, do want to work on a PR that does that?

timostamm:
A `global.tsProtoHasInitializedProtobuf` might be unnecessary? The `if` body below runs only once:

[...]

PR #112, opened 2020-08-14
Adds the option `configureProtobuf` for issue #78

timostamm:
This PR adds a new option flag `configureProtobuf`. 

When `true`, the following code will be appended to each generated file:
```typescript
if (protobuf.util.Long !== Long) {
  protobuf.util.Long = Long;
  protobuf.configure();
}
```

With the following imports (added with `ts-poet`): 
```typescript
import * as protobuf from 'protobufjs/minimal';
import * as Long from 'long';
```

No test coverage. Maybe a use case for #111 ?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -264,6 +264,9 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
   Note that `addGrpcMetadata`, `addNestjsRestParameter` and `returnObservable` will still be false.
 
+- With `--ts_proto_opt=configureProtobuf=true`, the binary reader of `protobufjs` will be explicitly configured to use the package `long`. This option may be necessary if you use `ts-proto` in the browser.  
+
+
 ### Only Types
 
 If you're looking for `ts-proto` to generate only types for your Protobuf types then passing all three of `outputEncodeMethods`, `outputJsonMethods`, and `outputClientImpl` as `false` is probably what you want, i.e.:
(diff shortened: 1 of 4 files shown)


stephenh:
@timostamm hey, I think this generally makes sense, although I'm wondering if we should just always do it, i.e. if `forceLong=long` is on? I suppose being able to turn it _off_ makes sense (i.e. maybe we should keep the config value), but defaulting on seems like a good idea?

stephenh:
I'd also like to do this only once somehow, instead of once-per-generated file. Like have some sort of `global.tsProtoHasInitialized` or something like that, that we first check. Seems like that'd be a good idea? Maybe not a huge deal if the `.configure` call is cheap I guess, so might be prematurely optimizing...wdyt?

timostamm:
Hey @stephenh , 

Only emitting the snippet when `forceLong=long` would not work. The `configure()` is also necessary for `forceLong=string` because `Long.js` is required to convert from [LongBits](https://github.com/protobufjs/protobuf.js/blob/master/src/util/longbits.js) to `string`. 

`protobuf.configure()` will run only once. The if statement above makes sure of that. 

Inverting the option `configureProtobuf` should work fine. Is `dontConfigureProtobuf` a fitting name?

stephenh:
> protobuf.configure() will run only once

Ah right, right, sorry I missed that.

> Is dontConfigureProtobuf a fitting name

I like `configureProtobuf` as the name, I'm just thinking it should default to `true` if `forceLong` is set to `string` or `long`? I.e. it's only if `forceLong=number` that we could skip this? Is that right?

timostamm:
@stephenh, thanks for your feedback and sorry for my late response. 
I have updated the PR accordingly. 

I have noticed that PR #125 overlaps with this PR. Please feel free to accept or drop my PR as you see fit. No bad feelings at all if you decide PR #125 is the better way to go.

stephenh:
Fixed through #125; thanks @timostamm and sorry again for not realizing the overlap sooner!

Issue #110, opened 2020-08-13
Flag to generate enums as TypeScript enums


philikon:
What's the reason for generating protobuf enums the way they're currently being generated, e.g.

```proto
enum StateEnum {
  UNKNOWN = 0;
  ON = 2;
  OFF = 3;
}
```

becomes
```ts
export const StateEnum = {
  UNKNOWN: 0 as const,
  ON: 2 as const,
  OFF: 3 as const,
  UNRECOGNIZED: -1 as const,
}
export type StateEnum = 0 | 2 | 3 | -1;
```

rather than

```ts
enum StateEnum {
  UNKNOWN = 0,
  ON = 2,
  OFF = 3,
  UNRECOGNIZED = -1,
}
```

Would it be ok if we made the native TypeScript enum available as a flag? It's handy for a variety reasons, including the ability to do reverse lookups (find name by value).

stephenh:
Yeah, they are not `enum`s b/c of this issue:

https://github.com/stephenh/ts-proto/issues/40

Where I was "adding" methods to the enum like `StateEnum.encode` / `StateEnum.decode` by using a `namespace StateEnum`, which doesn't work in babel.

I had also looked at putting the enum `encode` / `decode` methods "somewhere else", but having them be "like static methods on `StateEnum`" makes it really convenient when using enums across different proto files...

Granted, could fix this, and then go back to actual-`enum`s.

philikon:
Ah, of course. The `encode`/`decode` thing makes sense. I mostly care about the reverse lookup, not having them be actual enums. Any objections to adding the reverse lookup? Should it be behind a flag? My inclination is no, since regular TS enums support reverse lookup out of the box.

stephenh:
Yeah, I agree would be great to avoid yet-another-flag for this...

I'm torn. I'm tempted to just go back to `enum`s and just figure out where to put the `encode`/`decode` methods.

Part of the wrinkle is that it can't be something generic like:

```
export class Enums {
  public static decodeStateEnum(...) ...
  public static encodeStateEnum(...) ...
}
```

B/c then `import { Enums } from file-a` and `import { Enums } from file-b` will conflict (technically ts-poet 3.x solves this and would do `Enums as Enums1` and `Enums as Enums2`...so could just bite the bullet and upgrade to that...

Or could use top-level functions, i.e. just `export function decodeStateEnum`...

stephenh:
[...]

PR #113, opened 2020-08-16
Go back to use `enum` keyword for enums.

stephenh:
Previously we used an enum-like pattern so that we could attach
methods like toJSON/fromJSON to the enum itself, i.e.
StateEnum.fromJSON.

This was done out of a) convenience b/c we already knew where the
`StateEnum` came from (i.e. had it imported) and b) in-theory
developer ergonomics for calling the `StateEnum.fromJSON` method.

However, a) is easy to fix with just trying harder, i.e. a new
getEnumMethod that can find top-level functions for a given enum,
i.e. stateEnumFromJSON is in module x/y/z.

And b) is not really a factor b/c developers should rarely serialize
an enum directly and are probably/basically always working with
MessageFoo.to/fromJSON, and it's only the internals of
MessageFoo.to/fromJSON that have to know where the enum functions
live/how to call them.

So, this PR negates a) and ignores b), especially because developers
really do expect the protobuf enums to TS enums, i.e. for doing
things like Object.entries(StateEnum).

### DIFF
--- a/integration/codegen.sh
+++ b/integration/codegen.sh
@@ -1,10 +1,25 @@
 #!/usr/bin/env bash
 shopt -s globstar
 
-# Runs the local code generator for each of our .bin test files (as created/kept
+# Runs the local code generator for the .bin test files (as created/kept
 # up-to-date by ./update-bins.sh). Good for local iteration of WIP changes.
+#
+# Usage:
+#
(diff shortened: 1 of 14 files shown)


stephenh:
Fixes #110

PR #125, opened 2020-09-05
Ensure protobuf uses the long library if 64-bit types are used.

### LINKED ISSUE #78: ts-proto expects Reader.int64() to always return Long, but protobufjs may return number

stephenh:
Fixes #78.


Raiondesu:
Hey, @stephenh, thanks for submitting this!
I just wanted to add a quick note:
```ts
if (protobuf.util.Long !== Long) {
  protobuf.util.Long = Long;
  protobuf.configure();
}
```
This code errors with some configs, as util.Long and Long types don't overlap correctly in TS-terms.

Please, consider implementing a following change:
https://github.com/Raiondesu/ts-proto/blob/c7a208275314fdbf14065e3a04c80deb1a346b33/src/main.ts#L280

timostamm:
@stephenh , @Raiondesu: The PR #112 has been updated with the suggested change (casting Long to any). 

But maybe this PR is the better solution? 

Not sure If I read the update to main.ts correctly, but if you do the "configureProtobuf" automatically for .proto files using 64 bit integral types, that sounds like a smart way to tackle the issue. 

Please decide as you see fit @stephenh.

stephenh:
>  "configureProtobuf" automatically for .proto files using 64 bit integral types

Yep, that's what this is doing...I was hesitant at first about calling `configure()` b/c I've never had to really do that before, and so I think I steered your PR down the "well, let's do this behind a config flag", but I think I've come around that this should be always safe...

I.e. if anything a user could call `configure` on their own first, with whatever esoteric options they might haven, and then this code would not run (assuming they also setup long).

@Raiondesu thanks for the `as any` suggestion, I've added it to this PR.

Issue #129, opened 2020-09-09
Add observable support to grpc-web clients


PhilipMantrov:
` --ts_proto_opt=outputClientImpl=grpc-web,returnObservable=true` 
All methods (except stream methods) still promise.

stephenh:
Yeah, that combination of options is not supported. `returnObservable` only works with the nestjs output.

PhilipMantrov:
@stephenh this might be useful to you
```
import { Observable } from "rxjs";
import { grpc } from "@improbable-eng/grpc-web";
import { share, take } from "rxjs/operators";

const DEFAULT_TIMEOUT_TIME: number = 3 /* seconds */ * 1000 /* ms */;

export function grpcUnaryToObservable<T>(host: string, def: any, request: any): Observable<T> {
  return new Observable<T>(observer => {
    grpc.unary(def, {
      host,
      request,
      onEnd: (next) => {
        if (next.status !== 0) {
          observer.error({
            code: next.status,
            message: next.statusMessage,
          });
        } else {
          observer.next(next.message as any);
          observer.complete();
        }
      },
    });
  }).pipe(take(1));
}

export function grpcInvokeToObservable<T>(host: string, def: any, request: any): Observable<T> {
  return new Observable<T>((observer) => {
    const upStream = (() => {
      grpc.invoke(def, {
        host,
        request,
        transport: grpc.WebsocketTransport(),
        onMessage: (next) => {
          observer.next(next as any);
        },
        onEnd: () => {
          setTimeout(() => {
            upStream();
          }, DEFAULT_TIMEOUT_TIME);
        },
      });
    });
```

    upStream();
  }).pipe(share());
}

PhilipMantrov:
After some investigation i realize what wrong with Observable for stream methods: In interface we have forced `Observable` (and i find why you doing this (hello nestjs implementation and past issue (https://github.com/stephenh/ts-proto/issues/59)) , but in method we have `grpc.unary` with Promise returned. 

For stream methods we need to use `grpc.invoke`, as I indicated in the example above.

stephenh:
Pretty this is fixed, and just forgot to close the issue.

Issue #128, opened 2020-09-09
[Suggestion] export const for generateGrpcServiceDesc and generateGrpcMethodDesc


PhilipMantrov:
Need do generateGrpcServiceDesc and generateGrpcMethodDesc const's exportable for additional customs unary methods. 
Example: I have `export function grpcInvokeToObservable<T>(host: string, def: any, request: any): Observable<T>` with custom factory and some additional checkers and i need to provide "def" but i cant because they not exportable.

stephenh:
Looks like this was already fixed in a separate PR.

PR #131, opened 2020-09-10
returnObservable, grpc.invoke(), exports const's interfaces

PhilipMantrov:
returnObservable option implementation for grpc-web, 
grpc.invoke() streaming support added,
exports const's interfaces for grpc-web

https://github.com/stephenh/ts-proto/issues/129
https://github.com/stephenh/ts-proto/issues/128


stephenh:
@PhilipMantrov great! Fwiw it'll probably be a few days before I can fully take a look. Just after a quick scan, two asks:

1) can you add a new/made-up streaming method to the existing `integration/grpc-web/example.proto` file, rerun `update-bins.sh` and `codegen.sh grpc-web` and then include the updated/generated `example.ts` code output for the now-can-be-streaming method(s)?

2) this is a larger ask, and assumes golang knowledge, so perfectly fine to skip, but would be great if you updated the `integration/grpc-web/main.go` file to have a matching server impl for the dummy streaming impl (i.e. just stream back like 2-3 dummy responses and then close the stream), such that running `./client-ts.ts` would show the streaming working. Disclaimer that this is not technically part of the CI test running, and also the go project setup is pretty "I just copy/pasted this in" and may not be canonical go...so...ymmv, not really how sure it easy it is to update...

src/generate-grpc-web.ts
```diff
@@ -231,5 +250,154 @@ return new Promise((resolve, reject) => {
...
+          request,
+          metadata: maybeCombinedMetadata,
+          transport: grpc.WebsocketTransport(),
+          debug: this.options.debug,
+          onMessage: (next) => {
+            observer.next(next as any);
+          },
+          onEnd: () => {
+            setTimeout(() => {
+              upStream();
```
stephenh, line 282: Curious about this, why do we recall `upStream` and so recall `grpc.invoke` when `onEnd` is called?

PhilipMantrov: The standard behavior for a grps stream is to always be active and close on the backend side, but since all http 1.1 proxies have a ttl for streams, we must reconnect automatically. (By ther way, we can put the possibility of reconnection in the arguments as an option, why not).

stephenh: Ah cool, that makes sense / I'm good with this as-is. Thanks!

PhilipMantrov:
> @PhilipMantrov great! Fwiw it'll probably be a few days before I can fully take a look. Just after a quick scan, two asks:
> 
> 1. can you add a new/made-up streaming method to the existing `integration/grpc-web/example.proto` file, rerun `update-bins.sh` and `codegen.sh grpc-web` and then include the updated/generated `example.ts` code output for the now-can-be-streaming method(s)?
> 2. this is a larger ask, and assumes golang knowledge, so perfectly fine to skip, but would be great if you updated the `integration/grpc-web/main.go` file to have a matching server impl for the dummy streaming impl (i.e. just stream back like 2-3 dummy responses and then close the stream), such that running `./client-ts.ts` would show the streaming working. Disclaimer that this is not technically part of the CI test running, and also the go project setup is pretty "I just copy/pasted this in" and may not be canonical go...so...ymmv, not really how sure it easy it is to update...

1. Completed. 
2. I have some problem with --gogo_out plugin and i use standard --go_out and create classic grpc service, with classic rpc implementation, if you use my implementation pls use grpcwebproxy or simular http1.1 proxy (i use for testing bloomRPC because i dont have enough free time for testing all properly).

PhilipMantrov:
Added a stream reopening handler for certain status response codes. 
`[2, 4, 8, 9, 10, 13, 14, 15]`
These codes are may be discussed.
https://developers.google.com/maps-booking/reference/grpc-api/status_codes

PhilipMantrov:
>i am use for testing bloomRPC because i dont have enough free time to test everything properly 
upd: checked `integration/grpc-web/main.go` with grpc-web angular based frontend, working flawlessly.

PhilipMantrov:
>i am use for testing bloomRPC because i dont have enough free time to test everything properly

upd: checked `integration/grpc-web/main.go` with grpc-web angular based frontend, working flawlessly.

src/generate-grpc-web.ts
```diff
@@ -150,30 +159,42 @@ export function addGrpcWebMisc(options: Options, _file: FileSpec): FileSpec {
...
+        : TypeNames.PROMISE.param(TypeNames.ANY)
+    );
+  fnI = fnI
     .addTypeVariable(t)
     .addParameter('methodDesc', t)
     .addParameter('request', TypeNames.ANY)
     .addParameter('metadata', TypeNames.unionType(TypeNames.anyType('grpc.Metadata'), TypeNames.UNDEFINED))
-    .returns(TypeNames.PROMISE.param(TypeNames.ANY));
-  rpc = rpc.addFunction(fn);
+    .returns(TypeNames.anyType('Observable@rxjs').param(TypeNames.ANY));
```
stephenh, line 191: I think technically this is going to add an `rxjs` dependency for anyone using the grpc-web output, i.e. even if they don't use streaming at all, their `Rpc` interface is going to have this `invoke` type in it...

Can we conditionally add the `invoke` method, and only output if we know they have methods that are streaming?

src/generate-grpc-web.ts
```diff
@@ -231,5 +252,160 @@ return new Promise((resolve, reject) => {
...
+  }).pipe(%T(1));
+`,
+            BrowserHeaders,
+            grpc,
+            take
+          )
+        )
+    )
+    .addFunction(
+      FunctionSpec.create('invoke')
```
stephenh, line 366: Is this just a copy/paste of the `invoke` from the `generateGrpcWebImplPromise` method? Can we share these with some sort of `.addFunction(createInvokeMethod())`?

stephenh (commented):
Sorry for not reviewing this very quickly, but it looks great/functional; I have a few questions/asks if you don't mind. Thanks!

integration/grpc-web/generated/example_pb.d.ts
```diff
@@ -0,0 +1,213 @@
+// package: rpx
```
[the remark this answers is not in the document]
PhilipMantrov: Ok, then I just ignore these files in pr?

integration/only-types/google/protobuf/timestamp.ts
```diff
@@ -1,18 +1,16 @@
...
- *  Gregorian calendar backwards to year one.
- *
- *  All minutes are 60 seconds long. Leap seconds are "smeared" so that no leap
- *  second table is needed for interpretation, using a [24-hour linear
- *  smear](https://developers.google.com/time/smear).
- *
- *  The range is from 0001-01-01T00:00:00Z to 9999-12-31T23:59:59.999999999Z. By
- *  restricting to that range, we ensure that we can convert to and from [RFC
- *  3339](https://www.ietf.org/rfc/rfc3339.txt) date strings.
+ *  A Timestamp represents a point in time independent of any time zone
```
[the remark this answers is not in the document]
PhilipMantrov: i think its because this a newer version of protobuf

integration/grpc-web/generated/example_pb.d.ts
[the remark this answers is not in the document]
stephenh: Yep, that'd be great

PhilipMantrov:
I would like to see this pr asap in master (and in npm), can you write what else me need to do to make this pr be accepted?

stephenh:
Hey @PhilipMantrov sure, it looks like there are a few file conflicts; can you fix those, then I'll merge and release. Thanks for your patience on this...

stephenh:
Alright, released as v1.40.0. Thanks!

src/generate-grpc-web.ts
```diff
@@ -148,33 +157,50 @@ function methodDescName(serviceDesc: ServiceDescriptorProto, methodDesc: MethodD
...
   return rpc;
 }
 
 /** Implements the `Rpc` interface by making calls using the `grpc.unary` method. */
-function generateGrpcWebImpl(): ClassSpec {
+function generateGrpcWebImplPromise(): ClassSpec {
   const maybeMetadata = TypeNames.unionType(TypeNames.anyType('grpc.Metadata'), TypeNames.UNDEFINED);
   const optionsParam = TypeNames.anonymousType(
-    ['transport?', TypeNames.anyType('grpc.TransportFactory')],
+    ['unaryTransport?', TypeNames.anyType('grpc.TransportFactory')],
```
stephenh, line 202: @PhilipMantrov out of curiosity, do the `unaryTransport` and `invokeTransport` have to be different? 

...also, this is a per-request override anyway, so I think it'd be simpler to do just `transport` and have the user not care about if its supposed to be `unary` or `invoke` under the hood.

I'm planning on changing this to just `transport` like it was before, but let me know if there is some rationale here that I don't know about.

integration/grpc-web/main.go
```diff
@@ -29,49 +25,50 @@ var (
...
 
-	grpcServer := grpc.NewServer()
-
-	rpx.RegisterDashStateServer(grpcServer, &stateService{})
-	rpx.RegisterDashAPICredsServer(grpcServer, &credsService{})
-
-	grpclog.SetLogger(log.New(os.Stdout, "exampleserver: ", log.LstdFlags))
+type stateService struct{}
 
-	wrappedServer := grpcweb.WrapServer(grpcServer)
```
stephenh, line 44: @PhilipMantrov I'm confused about this, why did you remove this? Doesn't this mean the go server no longer speaks grpc-web?

src/generate-grpc-web.ts
PhilipMantrov: Often they use websocket transport for streams, and standard transport for unary. We should be able to use different transports for unary and invoke methods.

integration/grpc-web/main.go
PhilipMantrov: Don't remember why I deleted this, sorry.
I think we need return this part.

src/generate-grpc-web.ts
stephenh: Ah okay. Just for my own understanding, could the websocket transport be used for both unary and invoke, if a user wanted to do that?

Fwiw I also missed that this is a breaking change to the api. I'm thinking about keeping the old transport name for the unary and also invoke by default, and then adding a new 'streamingTransport', that would be for the invoke usage (but invoke would use the 'transport' if 'streamingTranspory' isn't set).

Kind of minor, but I like steamingTransport as a less "implementation details" name instead of the user knowing about invoke vs unary.

Sound okay?

integration/grpc-web/main.go
stephenh: Cool, np, I'll work on adding it back. Give me an excuse to re-learn some go. :-)

integration/grpc-web/example.ts
```diff
@@ -180,6 +190,36 @@ export class GrpcWebImpl implements Rpc {
...
+              onEnd: () => {
+                setTimeout(() => {
+                  upStream();
+                }, DEFAULT_TIMEOUT_TIME);
+              },
+            });
+          });
+
+          upStream();
+        }).pipe(share());
```
stephenh, line 220: @PhilipMantrov Just for my own understanding, what does this `pipe(share())` do? I haven't done a lot with rxjs observables and so am just curious to learn. Thanks! Fwiw I read the docs about it becoming a hot observer...something...dunno, maybe just need to read more.

src/generate-grpc-web.ts
PhilipMantrov: >Just for my own understanding, could the websocket transport be used for both unary and invoke, if a user wanted to do that?

Yes he can, but don't forget about active websocket connection limitations for modern browsers (6 per page and 255 per browser).

integration/grpc-web/example.ts
PhilipMantrov: share() move observable from cold to hot used subject fabric (other words we create subject for this observable). It's useful for request (especially streams).

PR #133, opened 2020-09-15
Allow maps with enum values

ahmadj-levelbenefits:
This patch adds a message containing a map with enum values that cannot be compiled on the trunk. The error message:
```
FAILED!No type found for Error: No type found for 
    at Object.fail (/Users/ahmadjarara/ts-proto/build/utils.js:23:11)
    at toModuleAndType (/Users/ahmadjarara/ts-proto/build/types.js:332:46)
    at messageToTypeName (/Users/ahmadjarara/ts-proto/build/types.js:326:28)
    at Object.basicTypeName (/Users/ahmadjarara/ts-proto/build/types.js:96:20)
    at readSnippet (/Users/ahmadjarara/ts-proto/build/main.js:673:68)
    at /Users/ahmadjarara/ts-proto/build/main.js:698:133
    at Array.forEach (<anonymous>)
    at generateFromJson (/Users/ahmadjarara/ts-proto/build/main.js:626:23)
    at /Users/ahmadjarara/ts-proto/build/main.js:75:35
    at /Users/ahmadjarara/ts-proto/build/main.js:414:9--ts_proto_out: protoc-gen-ts_proto: Plugin failed with status code 1.
```

The proposed fix is to pass in the full valueType instead of just passing in the type of the valueType. This has the metadata ts-proto needs to construct the enum. Included are some changes needed to make the resulting code compile. An assumption this PR makes is that all enums ts-proto generates are numeric enums, that is, they can be indexed by a number instead of a string or some other literal rep.


stephenh:
@ahmadj-levelbenefits I poked around in the git history and I can't tell / don't recall why the `FieldProtoDescriptor.create` call was done like that.

I've pulled down your change locally, re-run the codegen for all of the examples, and AFAICT it doesn't break anything, so I think we should be good to do that.

I'll make some inline comments on your PR but otherwise lgtm! Thanks!

integration/map-with-enums/map-with-enums.proto
```diff
@@ -0,0 +1,13 @@
...
+package mapWithEnums;
+
+enum Enums {
+  UNKNOWN = 0;
+  GOOD = 1;
+  BAD = 2;
+}
+
+message MapWithEnums {
+  map<int32, Enums> enums_by_id = 1;
```
stephenh, line 12: Given this is such a simple scenario, can you edit `integration/simple/simple.proto` and add a `mapOfEnums` example to the `SimpleWithMap` message? That way this will live next to the other "map of ..." tests.

(And then similarly move your test into `simple-test.ts`.

Normally I like smaller repro tests like you did, but since each new `integration/...foo...` directory requires another codgen run / another pbjs run / etc., I've been nudging towards sharing `*.proto` files more often than not.

Sound okay?

stephenh (approved):
Small ask about moving the test file around, but otherwise looks great. Thanks!

integration/map-with-enums/map-with-enums.proto
ahmadj-levelbenefits: Sounds good, but this PR still has some more work coming. Map-with-enums.ts did not compile before the latest round of changes, and now that this proto has moved to simple, simple.ts no longer compiles. I'm working on figuring out what's up.

integration/simple/simple.ts
```diff
@@ -1834,6 +1851,130 @@ export const SimpleWithSnakeCaseMap_EntitiesByIdEntry = {
...
+      }
+    }
+    return message;
+  },
+  fromJSON(object: any): SimpleWithMapOfEnums {
+    const message = { ...baseSimpleWithMapOfEnums } as SimpleWithMapOfEnums;
+    message.enumsById = {};
+    if (object.enumsById !== undefined && object.enumsById !== null) {
+      Object.entries(object.enumsById).forEach(([key, value]) => {
+        message.enumsById[Number(key)] = StateEnum(value);
```
ahmadj-levelbenefits, line 1887: here's one compile error

integration/simple/simple.ts
```diff
@@ -1834,6 +1851,130 @@ export const SimpleWithSnakeCaseMap_EntitiesByIdEntry = {
...
+    }
+    return message;
+  },
+  fromPartial(object: DeepPartial<SimpleWithMapOfEnums>): SimpleWithMapOfEnums {
+    const message = { ...baseSimpleWithMapOfEnums } as SimpleWithMapOfEnums;
+    message.enumsById = {};
+    if (object.enumsById !== undefined && object.enumsById !== null) {
+      Object.entries(object.enumsById).forEach(([key, value]) => {
+        if (value !== undefined) {
+          message.enumsById[Number(key)] = StateEnum(value);
```
ahmadj-levelbenefits, line 1898: ibid

integration/simple/simple.ts
```diff
@@ -1834,6 +1851,130 @@ export const SimpleWithSnakeCaseMap_EntitiesByIdEntry = {
...
+      })
+    }
+    return message;
+  },
+  toJSON(message: SimpleWithMapOfEnums): unknown {
+    const obj: any = {};
+    obj.enumsById = {};
+    if (message.enumsById) {
+      Object.entries(message.enumsById).forEach(([k, v]) => {
+        obj.enumsById[k] = simpleWithMapOfEnums_EnumsByIdEntryToJSON(v);
```
ahmadj-levelbenefits, line 1909: This one will take some work.

ahmadj-levelbenefits:
These failures seem to be the same failures on master. I'm happy to work with you on fixing them in a separate PR.

As for this PR: in order to get map value enums working I had to make the assumption that the enums ts-proto generates are [numeric enums](https://www.typescriptlang.org/docs/handbook/enums.html#numeric-enums), and that their wire format was their indices. Do you think that's an okay assumption to commit to?

stephenh:
> These failures seem to be the same failures on master.

Ah shoot, I had some out-of-date files locally so didn't notice those. Thanks! I pushed a fix to master for them.

> enums ts-proto generates are numeric enums, and that their wire format was their indices.

Yep, I think that's good/fine. Especially the wire format. We might eventually support string enums but we can upgrade this code to handle that when/if needed.

stephenh:
I've released this as v1.35.1. Thanks!

ahmadj-levelbenefits:
Sounds great -- thanks for fixing/looking/merging.

Issue #36, opened 2020-03-07
Import declaration conflicts with local declaration


jeffpyke:
file1.proto
```proto
syntax = "proto2";

package common.a;
message Foo {}
```
file2.proto
```proto
syntax = "proto2";
import "file1.proto";

package common.b;
message Foo {}

message Bar {
  optional common.a.Foo foo = 1;
}
```

Generated Typescript (file2.ts) (error):
```ts
import { Foo } from 'file1';
export interface Foo {
}
export interface Bar {
  foo: Foo | undefined;
}
```
I think this error is caused by loss of the package information.

stephenh:
Thanks for the report; kind of surprised b/c ts-proto generally tries to understand packages, although the mapping between a proto package and JS/TS module and file paths are all kind of murky. I'll try to take a look at this soon. I might get distracted reorganizing some of the test projects to make this easier to repro...

stephenh:
Oh wait, I see. I missed that local `message Foo` when I first scanned this issue this morning.

Yeah, that will be a little bit more involved.

In theory, the ts-poet library that ts-proto uses is supposed to solve exactly this sort of "fancy auto-import" type thing, but it doesn't realize there could be local vs. remote conflicts.

So, ideally getting ts-poet fixed would transitively fix this issue, although then there is a small wrinkle that ts-poet has a 3.x API out that is a fairly big breaking change from the 2.x API that ts-proto is currently using.

Basically, yeah, I see the issue, and will (likely slowly but hopefully steadily) work towards a fix.

stephenh:
As an update here, I implemented import collision avoidance in the latest/3.x API for ts-poet, see this test case:

https://github.com/stephenh/ts-poet/blob/master/src/index-tests.tsx#L163

Which is cool, but now means either: a) backporting that feature to the ts-poet 2.x branch that ts-proto uses (...probably easier?), or b) migrating ts-proto onto ts-poet 3.x (best for the long term and should not be hard but definitely tedious work).

If this is a big pain point/blocker for you, you could look into either of those, otherwise I'll get around to b) at some point.

jeffpyke:
Thanks, that's neat! It's not a huge pain point right now, because we can just delete the (unused) conflicting import manually. Looking forward to the fix! I think it will be useful once we generate a larger number of proto ts interfaces (right now 10-20), but eventually have 1000s of proto files in a mono repo we would like to generate ts interfaces for. Thanks, this project is great!

[...]

PR #137, opened 2020-09-17
Resolve import collisions

ahmadj-levelbenefits:
It is common to have a file that imports two files with exports that have symbol collisions. Here is a minimal reproduction of that use case: see import-collision.ts for a demonstration.

I'm not sure how to fix this: ts-proto uses ts-poet for import generation, but the dependency hasn't been updated in a while: v1.0.1 is completely different from ts-poet at v3.2.

How should this be resolved?

The only real option I see is updating, to 3.2: https://github.com/stephenh/ts-poet/blob/master/src/Code.ts#L123


integration/import-collision/import-collision.ts
```diff
@@ -0,0 +1,127 @@
+import { Job } from './aws/aws';
+import { Job } from './gcp/gcp';
```
ahmadj-levelbenefits, line 2: Here's the issue.

stephenh:
Hi @ahmadj-levelbenefits , yep, your diagnosis is exactly correct: this is a known-issue in ts-poet 1.x, it's solved "for free" in ts-poet 3.x, and the solution for ts-proto is to just bite the bullet and covert ts-proto over to use ts-poet 3.x.

In general ts-poet 3.x is a much lighter-weight / "less Java" API, so the migration shouldn't be hard/complex, just tedious. :-)

I've been generally procrastinating this for "someday". Would be awesome if you wanted to take a crack at it, and I'd be happy to help with any pointers/etc.

ahmadj-levelbenefits:
To get the migration to be gradual, would it be possible to alias the old version, and use the new version from now on? https://classic.yarnpkg.com/en/docs/cli/add/#toc-yarn-add-alias

This way we can integration test at many steps of the migration instead of one big one at the end, at the risk of the migration taking net longer, but this specific problem being fixed much quicker (this is the major blocker towards us using it internally).

This might not be possible to begin with: since 3.2 expects to work with its notion of nodes for interleaving.

stephenh:
@ahmadj-levelbenefits "since 3.2 expects to work with its notion of nodes for interleaving." right, the two ts-poets can't really interop, so the only way to move gradually would be `file-a.ts`-is-generated-by-v1 and `file-b.ts`-is-generated-by-v2. However, ts-proto's open is generally a single admittedly huge file, so I think that effectively keeps an all-at-once migration.

ahmadj-levelbenefits:
I see this issue: https://github.com/stephenh/ts-proto/issues/36 effectively tracks this suggestion. 

I unfortunately can't commit to this effort, so I'll close this PR. Thanks for your words, work, and time.

PR #143, opened 2020-10-01
Add way to pass data loader options

felixmo:
Adds an optional `rpcDataLoaderOptions` field to `ctx` which would allow us to pass options to the data loader during runtime to, for instance, disable caching for a particular request.

### DIFF
--- a/src/generate-services.ts
+++ b/src/generate-services.ts
@@ -200,7 +200,7 @@ function generateBatchingRpcMethod(typeMap: TypeMap, batchMethod: BatchMethod): 
     .addParameter(singular(inputFieldName), inputType)
     .addCode('const dl = ctx.getDataLoader(%S, () => {%>\n', uniqueIdentifier)
     .addCode(
-      'return new %T<%T, %T>(%L, { cacheKeyFn: %T });\n',
+      'return new %T<%T, %T>(%L, { cacheKeyFn: %T, ...ctx.rpcDataLoaderOptions });\n',
       dataloader,
       inputType,
       outputType,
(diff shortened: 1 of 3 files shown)


src/generate-services.ts
```diff
@@ -288,5 +288,9 @@ export function generateDataLoadersType(): InterfaceSpec {
     .addParameter('identifier', TypeNames.STRING)
     .addParameter('constructorFn', TypeNames.lambda2([], TypeNames.typeVariable('T')))
     .returns(TypeNames.typeVariable('T'));
-  return InterfaceSpec.create('DataLoaders').addModifiers(Modifier.EXPORT).addFunction(fn);
+  return InterfaceSpec.create('DataLoaders').addModifiers(Modifier.EXPORT).addFunction(fn).addProperty('rpcDataLoaderOptions', 'DataLoaderOptions', { optional: true });
```
felixmo, line 291: `DataLoaderOptions` is actually https://github.com/graphql/dataloader/blob/6592183a83025c6cbe9b040ea57faf7a2b81bb50/src/index.d.ts#L78 but I'm not sure how to use this type here

stephenh: You can refer to other types like:

```
const dataloader = TypeNames.anyType('DataLoader*dataloader');
```

At the top of this file.

stephenh (commented):
Lgtm, there is a failing lint check, can you take a look at this?

src/generate-services.ts
felixmo: Looking at this again, the options type is defined as `Options<K, V>` where `K` & `V` are the request input and output. But since we're setting the options at the context-level I don't think we can really provide these types.

If it's OK, I think I'll just keep this as-is with `DataLoaderOptions` providing a subset of `Options` that can be set at the context-level.

PR #144, opened 2020-10-03
Generate protobuf package constant for each file

### LINKED ISSUE #142: Lookup Decoder by google.protobuf.Any typeUrl
We have a use case where we'd like to send a batch of protobuf messages from a server to a client that uses ts-proto for its version of the protobuf definitions. They'll be embedded in Any objects. Rather than having to manually maintain a map of typeUrl to Decoder, is there any way to (possibly reflectively) access all available Decoders and index them by typeUrl?

Thanks!

andrewparmet:
Here's a first pass at allowing code to see the protobuf package associated with generated types. I see some NestJS-associated package names as well, but they lack a uniform name.

Resolves #142.

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -9,6 +9,8 @@ const baseSimpleMessage: object = {
   numberField: 0,
 };
 
+export const protobufPackage = 'angular'
+
 export const SimpleMessage = {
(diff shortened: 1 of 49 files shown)


stephenh (approved):

stephenh:
Released as 1.36.0

PR #145, opened 2020-10-06
Make Timestamp/Date mapping optional

gribbet:
The mapping of google.protobuf.Timestamp to Date in the generated interfaces does not seem compatible with NestJS which does not serialize these as Dates.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -277,6 +277,8 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
   Note that `addGrpcMetadata`, `addNestjsRestParameter` and `returnObservable` will still be false.
 
+- With `--ts_proto_opt=useDate=false`, fields of type `google.protobuf.Timestamp` will not be mapped to type `Date` in the generated types.
+
 ### Only Types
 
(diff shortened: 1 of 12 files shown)


stephenh:
@gribbet cool, makes sense. Can you add the `useDate` param to the readme?

Additional, could you change one of the `integration/...` examples that uses nestjs (in its `parameters.txt` file) to have a `Timestamp` field in its proto file, and then include in the PR the generated code output that shows it's now the `Timestamp` type instead of the built-in `Date` type.

This would also ensure (within the ts-proto test suite) that having `useDate=false` set still generates working/it-at-least-compiles TS code.

gribbet:
Added a Timestamp field to the nestjs-simple test
Added useDate=false to the README

stephenh:
@gribbet looks great, thanks!

mo4islona:
Maybe be useful for someone who uses NestJS and wants Timestamp as Date

```ts
import protobuf from 'protobufjs';

interface Timestamp {
  seconds: number;
  nanos: number;
}

protobuf.wrappers['.google.protobuf.Timestamp'] = {
  fromObject(date: Date) {
    const millis = date.valueOf();
    return this.create({
      seconds: millis / 1_000,
      nanos: (millis % 1_000) / 1_000_000,
    });
  },
  toObject({ seconds, nanos }: Timestamp): Date {
    return new Date(seconds * 1_000 + nanos / 1_000_000);
  },
} as any;
```

PR #146, opened 2020-10-06
Add option for string enums

### LINKED ISSUE #119: Support string enums option
I.e. generate

```typescript
export enum StateEnum {
  UNKNOWN = "UNKNOWN",
  ON = "ON",
  OFF = "OFF",
  UNRECOGNIZED = "UNRECOGNIZED",
}
```

bastianeicher:
This adds an option `stringEnums=true` to generate TypeScript string enums instead of int enums. This is useful if you want "only types" and are using a gRPC REST Gateway configured to serialize enums as strings.

This inspired by #38 but reimplemented because the master has diverged significantly since then.

This resolves #119.

I initially also started implementing code for generating `...EnumFromInt` and `...EnumToInt` functions to allow `stringEnums=true` to work with `encode` and `decode` but it turned out to be a bit more complicated than I thought. Since I'd like to keep my first PR in this project small I decided to take the easy way out and enforce that `stringEnums=true` must be combined with `outputEncodeMethods=false`.


stephenh:
@bastianeicher cool, thanks for the PR. I think the initial "only if not encoding" pass sgtm.

I would ask, can you make a new `integration/simple-string-enums` type directory that has a proto file and a `parameters.txt` with this flag turned on, and then include the generated code in your PR?

That will basically serve as a regression test to make sure the compiled output for this feature doesn't break going forward.

(I believe there are some notes about the `integration/` directory setup in the readme, but let me know if you have questions.)

bastianeicher:
@stephenh Just added a `integration/simple-string-enums` dir as you proposed. Does the structure look right?

stephenh:
Looks great, thanks!

Issue #148, opened 2020-10-08
Allow disabling UNRECOGNIZED enum value


ShakedH:
I'm using this tool in order to create a client for defined values expected in a database.
I expect the data in the DB to be consistent with enums defined.

Hence, the `UNRECOGNIZED` option which is added automatically to enums causes noise and is unnecessary for me.

I think that adding an option to disable this value generation will be helpful.

stephenh:
Published in 1.38.0. Thanks!

PR #149, opened 2020-10-08
feat: add option to remove UNRECOGNIZED enum value

ShakedH:
Resolves https://github.com/stephenh/ts-proto/issues/148


ShakedH:
@stephenh I wasn't sure what do to in [`generateEnumFromJson`](https://github.com/stephenh/ts-proto/pull/149/files#diff-8cfead41d88ad47d44509a8ab0a109adR442) when there's no `UNRECOGNIZED` value and the value in the JSON is not recognised...
Perhaps it's better to allow `unrecognizedEnum=false` option only when `outputJsonMethods=false`?

stephenh:
> Perhaps it's better to allow unrecognizedEnum=false option only when outputJsonMethods=false

@ShakedH what you're seeing as "only as an issue in `generateEnumFromJson`" is actually more pervasive, i.e. you're going to run into the same boundary case in `decode`, where "the external system has enum value 10 ... and I don't, what do I do?".

I think what should probably happen, in both `generateEnumFromJson` for JSON and also `decode` for binary, is that if you want to turn off `UNRECOGNIZED` as an explicit value, then unrecognized enum values should cause runtime exceptions.

Sound okay?

ShakedH:
@stephenh Yeah, I thought of this course of action at first as well, but I saw there aren't many cases where the generated code contains an error throw.
I'll make the changes. Do you think it's something that needs to be added to the `README`?

stephenh:
> I don't see where the UNRECOGNIZED value of the enum is used in decode

Yeah, I think this is a bug in `decode` that it never learned to support that. :-/ We can ignore that for this PR.

Issue #122, opened 2020-08-27
google.protobuf.Int64Value is being generated as number on interface regardless of forceLong=long


idlebot:
@stephenh given then following proto:

 ```proto
syntax = "proto3";

package loft.bug.v1;

import "google/protobuf/wrappers.proto";

message Bug {
    google.protobuf.Int64Value value = 2;
}
```

The interface being generated is:

```typescript
export interface Bug {
  value: number | undefined;
}
```

regardless of the forceLong being `long` or `string`

stephenh:
@idlebot ah yeah, you're right, looks like the wrapper types don't look at the `forceLong` value. I've tagged this as a bug. If you could work on a PR, that'd be great.

PR #151, opened 2020-10-09
Apply forceLong to wrapper types

gribbet:
https://github.com/stephenh/ts-proto/issues/122


stephenh:
This is great, thanks!

PR #154, opened 2020-10-15
Add option for omitting "DONE" log.

TheTonsOfCode:
There should be an option for skipping that log. When I am using "prototool" it results in an error while everything is ok.

```
2020-10-15T11:34:45.555+0200    WARN    protoc returned a line we do not understand, please file this as an issue at https://github.com/uber/prototool/issues/new       {"protocLine": "DONE"}
<input>:1:1:DONE
```

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -85,6 +85,7 @@ export type Options = {
   addNestjsRestParameter: boolean;
   returnObservable: boolean;
   lowerCaseServiceMethods: boolean;
+  printDoneInfo: boolean;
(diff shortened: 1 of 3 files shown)


stephenh:
Thanks for the PR; fwiw I'd rather just remove the log output entirely than add yet-another-config flag. Can you update the PR to do that instead?

TheTonsOfCode:
@stephenh done

stephenh (approved):
Thanks!

Issue #155, opened 2020-10-15
google.protobuf.BytesValue generates invalid code for fromJSON method


ebakoba:
Example message like:

```proto

message SomeMessage = {
  google.protobuf.BytesValue someBytes = 1;
}
```

generates following line the fromJSON method for that message:

```javascript
message.someBytes = Uint8Array(object.someBytes);
```

For valid code, keyword called `new` should be added, resulting in:

```javascript
message.someBytes = new Uint8Array(object.someBytes);
```

stephenh:
Fixed in v1.54.0. Thanks!

PR #156, opened 2020-10-16
Fix BytesValue handling in FromJSON method

ebakoba:
Resolves the issue raised in issue #155


ebakoba:
Can we pealse merge this @stephenh ?

PR #158, opened 2020-10-23
Process proto2 optional label

actuosus:
`optional` label for proto2 should be optional property in Typescript.


stephenh:
A lot of proto2 optional work just landed in #1007 , so going to close this out.

PR #166, opened 2020-11-15
Fix left over globalThis.Error scoping when using unrecognizedEnum=false

zapo:
There seemed to be one left over unscoped when `unrecognizedEnum=false`.


stephenh:
Thanks! Released in 1.38.1.

PR #171, opened 2020-11-30
Add support for meta-typings

Vilsol:
This PR adds a new parameter `outputMetaTypings` which when set to true will generate meta objects that store information about the original protos which can be accessed at runtime.

Using this it is possible to feed down into other generators, for example, generating a GraphQL facade for several services.

It supports messages, services and packages.

At the end of each generated meta block it generates a "package" level meta object that contains references to all other meta objects in the file.

If the parameter is not set, there is no difference in the output.

Examples:

---

Message:

```proto
message Simple {
  // Name field
  string name = 1;
  /* Age */
  int32 age = 2;
  google.protobuf.Timestamp created_at = 9; // This comment will also attach
  Child child = 3;
  StateEnum state = 4;
  repeated Child grand_children = 5;
  repeated int32 coins = 6;
  repeated string snacks = 7;
  repeated StateEnum old_states = 8;
  // A thing (imported from thing)
  ImportedThing thing = 10;
  repeated bytes blobs = 11;
  google.type.Date birthday = 12;
  bytes blob = 13;
}
```

```ts
export const metaSimple: { [key in keyof Required<Simple>]: MetaI | string } = {
  name: {meta:'builtin', type:'string', original:'string'} as MetaB,
  age: {meta:'builtin', type:'number', original:'int32'} as MetaB,
  createdAt: {meta:'union', choices: [undefined, {meta:'builtin', type:'Date', original:'.google.protobuf.Timestamp'} as MetaB]} as MetaU,
  child: {meta:'union', choices: [undefined, {meta:'object', type:'.simple.Child', name:'Child'} as MetaO]} as MetaU,
  state: {meta:'object', type:'.simple.StateEnum', name:'StateEnum'} as MetaO,
  grandChildren: {meta:'array', type:{meta:'object', type:'.simple.Child', name:'Child'} as MetaO} as MetaA,
  coins: {meta:'array', type:{meta:'builtin', type:'number', original:'int32'} as MetaB} as MetaA,
  snacks: {meta:'array', type:{meta:'builtin', type:'string', original:'string'} as MetaB} as MetaA,
  oldStates: {meta:'array', type:{meta:'object', type:'.simple.StateEnum', name:'StateEnum'} as MetaO} as MetaA,
  thing: {meta:'union', choices: [undefined, {meta:'object', type:'.simple.ImportedThing', name:'ImportedThing'} as MetaO]} as MetaU,
  blobs: {meta:'array', type:{meta:'builtin', type:'Uint8Array', original:'bytes'} as MetaB} as MetaA,
  birthday: {meta:'union', choices: [undefined, {meta:'object', type:'.google.type.Date', name:'DateMessage'} as MetaO]} as MetaU,
  blob: {meta:'builtin', type:'Uint8Array', original:'bytes'} as MetaB,
}
```

---

Service:

```proto
service PingService {
  rpc ping(PingRequest) returns (PingResponse);
}
```

```ts
export const metaPingService: { [key in keyof PingService]: MetaS<any, any> } = {
  ping: {request: {meta:'object', type:'.simple.PingRequest', name:'PingRequest'} as MetaO, response: {meta:'object', type:'.simple.PingResponse', name:'PingResponse'} as MetaO, clientStreaming: false, serverStreaming: false, encodeRequest: PingRequest.encode, decodeResponse: PingResponse.decode} as MetaS<PingRequest, PingResponse>,
}
```


README.markdown
```diff
@@ -287,6 +287,8 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
 - With `--ts_proto_opt=useDate=false`, fields of type `google.protobuf.Timestamp` will not be mapped to type `Date` in the generated types.
 
+- With `--ts_proto_opt=outputMetaTypings=true`, meta typings will be generated that can later be used in other code generators.
```
stephenh, line 290: What do you think about `outputSchema=true`? I.e. using the term "schema" instead of "meta typing". Maybe pedantically a "typing" in the TS sense typically isn't available at runtime, and what you're building actually is, hence a slight nudge away from the term "typing". If you really want to use the term "meta" we could do `outputMetadata=true`. Wdyt?

integration/meta-typings/google/type/date.ts
```diff
@@ -0,0 +1,123 @@
...
+  day: number;
+}
+
+const baseDateMessage: object = {
+  year: 0,
+  month: 0,
+  day: 0,
+};
+
+export interface MetaI {
```
stephenh, line 41: I like what you're doing here, but would prefer some slightly more clear names, i.e.:

* `MetaI` -> `MetaBase` ...or just `Meta`?
* `MetaO` -> `MetaMessage`?
* `MetaA` -> `MetaArray`
* `MetaM` -> `MetaMap`
* `MetaU` -> `MetaUnion`
* `MetaB` -> `MetaPrimitive` (I prefer primitive over builtit, is that okay?)
* `MetaS` -> `MetaService`

Sound okay?

integration/meta-typings/google/type/date.ts
```diff
@@ -0,0 +1,123 @@
...
+  },
+};
+
+export const metaDateMessage: { [key in keyof Required<DateMessage>]: MetaI | string } = {
+  year: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+  month: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+  day: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+}
+export const metaPackageGoogleType: { [key: string]: ['service', string, any, { [key: string]: MetaS<any, any> }] | ['enum', string, any, any] | ['message', string, any, { [key: string]: MetaI | string }] } = {
+  DateMessage: ['message', '.google.type.DateMessage', DateMessage, metaDateMessage],
```
stephenh, line 122: Cool.

integration/meta-typings/meta-typings-test.ts
```diff
@@ -0,0 +1,27 @@
+import { Child_Type, Simple, StateEnum } from './simple';
+
+const jan1 = new Date('1970-01-01T00:00:00.000Z');
+
+describe('simple', () => {
+  it('generates types correctly', () => {
```
stephenh, line 6: Can you add a test, maybe even an `toMatchInlineSnapshot` test, that asserts against the generated `metaPackageGoogleType`? To make sure all of the values match what you expect.

integration/meta-typings/google/type/date.ts
```diff
@@ -0,0 +1,123 @@
...
+    return message;
+  },
+};
+
+export const metaDateMessage: { [key in keyof Required<DateMessage>]: MetaI | string } = {
+  year: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+  month: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+  day: {meta:'builtin', type:'number', original:'int32'} as MetaB,
+}
+export const metaPackageGoogleType: { [key: string]: ['service', string, any, { [key: string]: MetaS<any, any> }] | ['enum', string, any, any] | ['message', string, any, { [key: string]: MetaI | string }] } = {
```
stephenh, line 121: Hm, Maybe just `metadata`? `schema`? for the name of this `const`.

src/generate-meta.ts
```diff
@@ -0,0 +1,295 @@
...
+} from 'ts-poet';
+import { maybeSnakeToCamel } from './case';
+import { OneofOption, Options } from './main';
+import { google } from '../build/pbjs';
+import DescriptorProto = google.protobuf.DescriptorProto;
+import FileDescriptorProto = google.protobuf.FileDescriptorProto;
+import ServiceDescriptorProto = google.protobuf.ServiceDescriptorProto;
+import FieldDescriptorProto = google.protobuf.FieldDescriptorProto;
+
+export function getMetaInterfaces() {
```
stephenh, line 34: This is a good approach for this PR, and maybe even indefinitely. Wdyt about maybe eventually having like a separate `ts-proto-schemas` or some sort of separate package that we could put these types into and import?

As a disclaimer, I know I've shied away from runtime dependencies, i.e. with things like `DeepPartial`, but those seemed smaller, and were also used by all clients of ts-proto, and not just ones who are opting into something specific/special like your runtime metadata/schema.

And yeah, let's keep this for this PR, just thinking of maybe later.

integration/meta-typings/google/type/date.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+const baseDateMessage: object = {
+  year: 0,
+  month: 0,
+  day: 0,
+};
+
+export interface MetaI {
+  readonly meta: 'object' | 'array' | 'map' | 'union' | 'builtin';
```
stephenh, line 42: What do you think about renaming this `meta` field to `kind`? I've generally seen `kind` used as a common name for the discriminator key in type unions in the typescript community.

stephenh (commented):
Generally looks good to me; mostly had questions around some naming that is personally more intuitive to me. Let me know what you think. Thanks!

isherman:
Maybe I'm misunderstanding here, but we already have [protobuf descriptors](https://developers.google.com/protocol-buffers/docs/techniques#self-description) as a standardized way of describing message structure and supporting reflection. 

Couldn't we just add a ts-proto [Descriptor](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto) message to the generated code instead?

stephenh:
@isherman interesting, that's a good point.

Fwiw I kind of like the more "TS idiomatic" approach (i.e. ADTs & kinds) that @Vilsol used (and some of the descriptors are a little crufty/showing their age).

That said, I admit what you describe is probably the "more correct" way to do it. I think it may come down to just a matter of which approach has been volunteered/actually implemented in a PR :-). 

@Vilsol I think up to you at this point, if you want to try @isherman 's approach. Eventually if someone implements the descriptor-style approach, we could either move to that (and deprecate this current kind-based version) or even just support both, especially given that ts-proto's lot in life seems to be having a config option per user to support _exactly_ their given use case/desired output. :-)

src/main.ts
```diff
@@ -228,6 +230,56 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
     file = addTimestampMethods(file, options);
   }
 
+  if (options.outputMetaTypings) {
```
stephenh, line 233: Just thinking ahead (...and also in retrospect), `main.ts`  is already too big, can you move the contents of this `if` to a new `generate-metadata.ts` (or `generate-schema.ts`)?

Vilsol:
@stephenh I have made all the requested changes.

I don't completely understand how the solution suggested by isherman would work. The issue this is solving that the proto schema is available at runtime, not just compile-time.

stephenh:
> The issue this is solving that the proto schema is available at runtime, not just compile-time.

@Vilsol right, isherman's suggestion was that the data you're making available at runtime could, instead of defining your own new types (the `MetaArray`, `MetaMap`, etc.) you could use the existing protobuf schema types, i.e. general code that does something like:

```
const importedThingSchema: DescriptorProto = {
  name: "importedThing",
  field: FieldDescriptorProto[] = [
    ...
  ]
}
```

With isherman's point being that the out-of-the-box `DescriptorProto` / `FieldDescriptorProto` / etc. defacto already define very that could possible be described in a proto file/schema, and also are already understood by users who've done `Descriptor`-based runtime/reflection.

What's interesting is that, when ts-proto runs, we get exactly this data structure handed to us (as binary input on stdin)...it would be cute if we could do something like:

```typescript
const descriptor = DescriptorProto.decode(stdin);
const output = ...regular output in main.ts...
if (options.outputMetadata) {
  output += `export const descriptor: DescriptorProto = ${descriptor}`
}
```

Because in theory that `descriptor` POJO/object literal already matches the `DescriptorProto` type, and we'd just want the codegen library (ts-proto) to generically ".toString()" the descriptor in a way that was valid TS output.

I've got an example of that in the new version of ts-poet (that I just upgraded ts-proto to use):

https://github.com/stephenh/ts-poet/blob/master/src/index-tests.tsx#L160

The upshot is that this could be a pretty simple feature to add, if done this way, and also match the canonical protobuf descriptor types; the wrinkle is that they are kinda old school/crufty looking and not as pleasant as your custom/TS idiomatic TS types that use ADTs/etc.

But I'm tempted to try this "just output the descriptor" approach.

Vilsol:
@stephenh I have re-written to use the proto-builtin types. There is now a root-level object called `protoMetadata` that contains the full proto dump (in JS form), all references to objects within the current file, and all references to dependency `protoMetadata`'s.

This makes it extremely easy to automate parsing as you only need to pass the top level `protoMetadata` and all the rest are referenced in a tree downwards.

If you wish I can split out the code for meta generation into a separate file as before, but because the code is a lot smaller I thought it wouldn't be too encroaching in the `main.ts` file.

src/main.ts
```diff
@@ -182,6 +184,72 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
...
+
+    values.push(code`
+      references: {${joinCode(references, { on: ',\n' })}}
+    `);
+
+    if (fileDesc.dependency) {
+      const dependencies = fileDesc.dependency.map((dep) => {
+        const mod = dep.replace('.proto', '');
+        const localName = mod.replace(/\//g, '_') + '_protoMetadata';
+        return code`${new ImportsName(localName, './' + mod, 'protoMetadata')}`;
```
stephenh, line 240: I think this can be:

```
return code`${imp(`protoMetadata@./${mod}`)}`
```

Disclaimer that this _just_ started working in the latest ts-poet that ts-proto started using ~yesterday, where ts-poet will notice the overlapping `protoMetadata` import symbols, and auto-suffix/rename them with `import { protoMetadata as protoMetadata1 } from ./mod`;

src/main.ts
```diff
@@ -182,6 +184,72 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
     chunks.push(generateDataLoadersType());
   }
 
+  if (options.outputSchema) {
+    chunks.push(code`
+      export interface ProtoMetadata {
+        fileDescriptor: ${imp('IFileDescriptorProto@protobufjs/ext/descriptor')};
+        references: { [key: string]: any };
```
stephenh, line 191: How about instead of a single `references` field, we have `messages: { [key: string]: any`, and then `enums` and `clients`.

Just thinking it might be nice to keep those separate in case someone wanted to, say, loop over the clients or loop over just the messages.

Wdyt?

src/main.ts
```diff
@@ -182,6 +184,72 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
     chunks.push(generateDataLoadersType());
   }
 
+  if (options.outputSchema) {
+    chunks.push(code`
+      export interface ProtoMetadata {
+        fileDescriptor: ${imp('IFileDescriptorProto@protobufjs/ext/descriptor')};
```
stephenh, line 190: Nice! Glad this approach worked out.

Fwiw I've been meaning to publish ts-proto-generated versions of `FileDescriptorProto` to npm, like `ts-proto-protobuf-descriptors` or something like that, and then use that for parsing the `protoc` input messages, instead of the current protobufjs types (which I had to use when originally writing ts-proto b/c ts-proto didn't exist yet :-) ).

But, just mentioning that, b/c then these type could be the ts-proto-generated `FileDescriptorProto` would might be slightly more pleasant to use (or at least somewhat different).

We don't have to do that for this PR, but just mentioning it as something that I might get around to doing.

stephenh (commented):
Cool, thanks for this updated version! Few questions but I think looks really good.

src/main.ts
Vilsol: I think that considering that the metadata gets generated for everything, if you want to have the list of just enums, you can use `protoMetadata.fileDescriptor.enumType` and the same for messages and services.

The reason I would like to keep this as a single large object is so that if you are given a `ProtoMetadata` instance, and a string reference of `.simple.Nested.InnerMessage.DeepMessage`, you don't have to check 3 different objects for its existence.

src/main.ts
Vilsol: That could add some convenience if you are only generating the typescript typings and metadata, however, if you are also generating the json methods or client implementations, it instantly adds the dependency to `protobufjs` which would sort of defeat the purpose. (and I also consider the official typings a bit more stable than ones generated from protos)

stephenh: > you are also generating the json methods or client implementations

True, it wouldn't need the json methods... There would be no client impls b/c the descriptor `*.proto` files only have messages and no services.

Re the `protobufjs` dependency, ts-proto already has this (at build-time) to decode the `protoc` plugin's stdin messages, via the currently-generated-by-pbjs messages, so just switching to ts-proto-generated messages would use the same `protobufjs` dependency for doing the raw binary decoding. So wouldn't/shouldn't be a new dependency.

> official typings a bit more stable than ones generated from protos

AFAIU these protobufjs typings are themselves generated from `*.proto`s, so I don't know that they are any more or less official than what ts-proto would similarly generate from the [descriptor](https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.descriptor.pb) proto file.

All said, I'm fine leaving this to later, it just means the caveat of a slight breaking change if/when `fileDescriptor` moves to use a ts-proto-generated message type.

Vilsol: Ok, yeah, that makes sense, and if the official typings are also generated from the protos, then there shouldn't really be any serious breaking changes when transferring over.

src/main.ts
```diff
@@ -182,6 +184,72 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
...
+    `);
+
+    const outputFileDesc: IFileDescriptorProto = {
+      ...fileDesc,
+      sourceCodeInfo: null,
+    };
+
+    if (fileDesc.sourceCodeInfo) {
+      outputFileDesc.sourceCodeInfo = {
+        location: fileDesc.sourceCodeInfo.location.filter((loc) => loc['leadingComments'] || loc['trailingComments']),
```
stephenh, line 203: Just curious, but what is the intent of this code? My guess is only outputing comments and not the `path` / `span` / etc. aspects of the source code location?

stephenh: Ah, I get it now, n/m.

src/main.ts
```diff
@@ -182,6 +184,72 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
...
+      } as any;
+    }
+
+    const values = [code`fileDescriptor: ${JSON.stringify(outputFileDesc)} as any`];
```
stephenh, line 207: Fwiw I really wanted to remove this `as any`. Just as is (using `JSON.stringify`), it looks like enums are coming out as strings instead of numbers, which doesn't match the type.

Using `FieldDescriptorProto.toObject` fixes that, but then the problem is that `proto3Optional` is a field in the output, but is not something in the protobufjs `IFieldDescriptorProto` yet (although it should be).

So makes sense to keep the `as any` .

stephenh:
Okay, this is released as v1.60.0. Thanks for the feature!

I was poking around a bit before merging, let me know if I broke anything. :-)

src/main.ts
stephenh: Fwiw I went ahead and moved over to ts-proto-generated messages for both the metadata in this PR and also just the core stdin/stdout message plugin parsing. I don't think there should be any significant differences but let me know if so.

PR #178, opened 2020-12-18
add outputMessageNames option

marcushultman:
I don't know at all if this is desired. Maybe there already is a way to, given a message, get the name of the proto message as a string?


marcushultman:
@stephenh 🙏🏻

stephenh:
@marcushultman thanks for the PR! Stepping back a bit, this seems like a subset of the functionality that's in #171 . If that PR got merged (or something like that), would that effectively replace what you need here? Or maybe with a tweak to #171 that each `Message` had a reference to its metadata?

marcushultman:
I skimmed through it, but it only seem to contain metadata about message fields. The proposed metadata object also lacks structure that allow any metadata about the message itself (if the message has a field named 'name' it would conflict). Even if the metaobject there could be changed to allow information about the message itself,  you'd still, given a message object, have no way of looking up the metaobject.

marcushultman:
If the proposed meta object was wrapped in an object under the 'field' hash, and the object was added to some meta() accessor it could work. So that `Simple.meta().name` is 'Simple', and `Simple.meta().fields` is the meta object in the other PR.

stephenh:
Yeah, that makes sense. One of the proposals on that PR is to mirror the existing protobuf descriptors, so to address your ask of mapping a message to its metadata, something like:

```
const descriptor = SimpleMessage.descriptor;
descriptor.fields // basically whatever is in the proto Descriptor type
descriptor.name // etc.
```

> given a message object, have no way of looking up the metaobject.

Fwiw I think this is also true of your current PR, and a general limitation of ts-proto today? I.e.:

```
const message = ...get something off the wire...
callSomethingThatWantsMetadata(message);
```

The `callSomethingThatWantsMetadata` doesn't have a way to look at just the `message` instance and map it back to the message it came from. Because of the ts-proto message instances being just object literals. So you'd always need to do something like:

```
callSomethingThatWantsMetadata(SimpleMessage, message);
````

Which is probably fine? But just checking that that's the case, since you seemed to be explicitly call that out as a use case (given a message instance, linking back to its metadata)

marcushultman:
For my use case I would only need to go from Message object to name (not from a plain object instance), as my PR would solve. The descriptor object you propose  would be enough.

stephenh:
Hey @marcushultman sorry for not following through on this, but in related good news, a PR just got merged that adds a `@type` field to messages, which I think is what you're looking for. It's the `outputTypeRegistry=true` option courtesy of @aikoven .

Issue #183, opened 2021-01-18
Deprecated properties in proto should be marked as deprecated in JSDoc comments 


ShakedH:
Marking a field as deprecated is [supported in proto](https://developers.google.com/protocol-buffers/docs/proto#options).

According to the docs, when generating Java code from the proto files, the property is marked with `@Deprecated` annotaion.

When generating Typescript code, the field should be marked as deprecated in [JSDoc](https://jsdoc.app/tags-deprecated.html)

stephenh:
Fixed in v1.42.0.

PR #184, opened 2021-01-18
WIP: feat: add @deprecated to JSDoc for deprecated fields

ShakedH:
Add JSDoc comment with `@deprecated` for deprecated fields.

Closes https://github.com/stephenh/ts-proto/issues/183

No tests added since JSDoc comments can't be tested.


ShakedH:
Hey @stephenh,

I added this fix but I have a problem in case of deprecated field with a comment marked with `/**/`.

You can see the result in [this property](https://github.com/stephenh/ts-proto/pull/184/files#diff-6607a70d9befea199bf07a9ff9e6a429bd1004da879305d8081a550d0ac5a4beR20)

Any idea how to fix this?

stephenh:
> I added this fix but I have a problem in case of deprecated field with a comment marked with /**/

Hey, just saw you merged master, thanks for doing that!

Specifically for comments, how they are handled changed pretty dramatically in the latest ts-poet upgrade. I want to nudge them around a little bit (i.e. I think we're still escaping `%%` twice and don't need to do that anymore), but maybe this upgrade also just solved the `/**` issue for you?

Even if not, it should be much easier to fix it with the new "just string literals" ts-poet API.

stephenh:
I pulled this down locally, made some tweaks, and have merged it. The functionality is in v1.42.0. Thanks!

PR #185, opened 2021-01-19
Explicitly define globalThis

### LINKED ISSUE #97: globalThis is not defined (in node < v12)
After using optionals, ie by importing Google's wrapped types:

```
google.protobuf.StringValue email = 1;
```

TS-Proto plugin adds additional ts files in Google folder. This contains file: _wrappers.ts_

This file uses _globalThis_ which is only available in node 12+ 

Could you please suggest right way of using it in lower node versions ? 
For now I included https://www.npmjs.com/package/globalthis polyfil and by force added
import statement to top of generated wrappers.ts file:
`var globalThis = require('globalthis')();`

willclarktech:
Closes #97 

I had to include the declarations because you can't refer to undeclared variables even when using `typeof`. The `self` and `window` declarations could be `let`s but somehow `var` seemed more appropriate since their scope isn't supposed to be limited to any particular block.

Let me know if this code should be located somewhere else or if anything else is required.

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -1,6 +1,14 @@
 import { Writer, Reader } from 'protobufjs/minimal';
 
 
+function getGlobalThis() {
+  if (typeof globalThis !== "undefined") return globalThis;
+  if (typeof self !== "undefined") return self;
+  if (typeof window !== "undefined") return window;
+  if (typeof global !== "undefined") return global;
+  throw new Error("Unable to locate global object");
+}
+
 export interface SimpleMessage {
   numberField: number;
 }
@@ -8,6 +16,12 @@ export interface SimpleMessage {
 const baseSimpleMessage: object = {
   numberField: 0,
 };
+
+declare var self: any | undefined;
+
+declare var window: any | undefined;
+
+var globalThis = getGlobalThis();
 
 export const protobufPackage = 'angular'
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -106,6 +106,17 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
   const moduleName = fileDesc.name.replace('.proto', '.ts');
   let file = FileSpec.create(moduleName);
 
+  file = file.addCode(CodeBlock.empty().addStatement('declare var self: any | undefined'));
+  file = file.addCode(CodeBlock.empty().addStatement('declare var window: any | undefined'));
+  // Make sure globalThis is defined
+  let getGlobalThis = FunctionSpec.create('getGlobalThis');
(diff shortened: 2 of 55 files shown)


willclarktech:
@webmaster128

src/main.ts
```diff
@@ -106,6 +106,20 @@ export function generateFile(typeMap: TypeMap, fileDesc: FileDescriptorProto, pa
...
+  file = file.addCode(CodeBlock.empty().addStatement('declare var window: any | undefined'));
+  // Make sure globalThis is defined
+  let getGlobalThis = FunctionSpec.create('getGlobalThis');
+  ['globalThis', 'self', 'window', 'global'].forEach((candidate) => {
+    getGlobalThis = getGlobalThis
+      .beginControlFlow('if (typeof %L !== "undefined")', candidate)
+      .addStatement('return %L', candidate)
+      .endControlFlow();
+  });
+  getGlobalThis.addStatement('throw new Error("Unable to locate global object")');
```
webmaster128, line 119: Why is this not added to the output code?

integration/angular/simple-message.ts
```diff
@@ -1,6 +1,21 @@
...
+  }
+  if (typeof self !== "undefined") {
+    return self;
+  }
+  if (typeof window !== "undefined") {
+    return window;
+  }
+  if (typeof global !== "undefined") {
+    return global;
+  }
```
webmaster128, line 16: Would it be possible to write in a less verbose way? Like

```ts
function getGlobalThis() {
  if (typeof globalThis !== "undefined") return globalThis;
  if (typeof self !== "undefined") return self;
  if (typeof window !== "undefined") return window;
  if (typeof global !== "undefined") return global;
}
```

Or does the code generator not support this?

webmaster128 (commented):
🚀

src/main.ts
willclarktech: Ah whoops, I need to reassign this here.

integration/angular/simple-message.ts
willclarktech: You can add anything you like, but not using the nice `.beginControlFlow`/`.endControlFlow` methods.

webmaster128: I don't have strong opinions here, just a slight tendency to avoid polluting people's code bases with many lines of boilerplate

stephenh:
@willclarktech looks great!

Fwiw I'm in the middle of upgrading ts-proto to the latest / not-as-Java-builder-all-the-things version of the ts-poet codegen library, so I want to land that first, and then will get this PR in. Shooting for ~this weekend-ish if that sounds good.

stephenh:
I redid this on master to avoid the conflicts, and released as 1.43.0. Thanks!

PR #189, opened 2021-01-22
fix: add deprecated comment when there are not other comments

ShakedH:
Currently, deprecated fields without any comments are not marked as deprecated.

This PR fixes this by adding a comment regardless of existing comments in proto - if necessary.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,5 +1,5 @@
 import ReadStream = NodeJS.ReadStream;
-import { EnvOption, LongOption, OneofOption, Options } from './main';
+import { Options, LongOption, EnvOption, OneofOption } from './main';
 import { SourceDescription } from './sourceInfo';
 import { code, Code } from 'ts-poet';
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -168,20 +169,21 @@ export function maybeAddComment(
...
-    if (lines.length === 1) {
-      comment = code`/** ${content} */`;
-    } else {
-      comment = code`/**\n * ${lines.join('\n * ')}\n */`;
-    }
+    lines = content.split('\n').map((l) => l.replace(/^ /, '').replace(/\n/, ''));
+  }
+  // Deprecated comment should be added even if no other comment was added
+  if (deprecated) {
+    lines.push('');
```
stephenh, line 176: Wdyt about only pushing this empty string if lines is non-empty? 

That would make `testField` look like:

```
  /** @deprecated */
  testField: string;
```

stephenh (approved):
Ah yeah, thanks makes sense!

src/utils.ts
ShakedH: Yeah I thought about it as well at first. Will add it now

stephenh:
Released in v1.42.1. Thanks!

PR #201, opened 2021-02-02
Fix typo in generate-services > generateRegularRpcMethod

willclarktech:
I just hit against this typo. Haven't looked into the test structure yet, but happy to add one if you point me to the right location.


stephenh:
Ah yep, thanks! I probably introduced that recently in a refactoring and missed it go by. Fixed in v1.53.0.

PR #203, opened 2021-02-03
Fix bytes value from json

ebakoba:
Resolves the issue raised in issue #155


stephenh:
Thanks for the PR!

Could you change `integration/simple/simple.proto`'s `SimpleWithWrappers` to includes a `BytesValue`, rerun `update-bins.sh`, then rerun `./codegen.sh simple`, to reproduce and show the generated code works now?

Also adding a unit test in `simple-json.test.ts` that shows decoding bytes from JSON would be even better.

ebakoba:
@stephenh 
After the changes newly generated code adds the ```new``` keyword before the ```Uint8Array```, which is the correct behavior.
image

I also updated the ```SimpleWithWrapper``` in ```integration/simple/simple.proto``` to include ```google.protobuf.BytesValue``` and updated tests to reflect that change.

Issue #200, opened 2021-02-01
fromTimestamp function is not generated


mharsat:
When generating a proto file with
```proto
import "google/protobuf/timestamp.proto";

message Metadata {
    google.protobuf.Timestamp last_edited = 1;
}
```
with the following options - `useOptionals=true, outputEncodeMethods=false, stringEnums=true, esModuleInterop=true`

I get a ts file with the `fromJsonTimestamp` but in which `fromTimestamp` and `Timestamp` are missing.

![Screen Shot 2021-02-01 at 12 36 54](https://user-images.githubusercontent.com/12008392/106448209-204bc380-648b-11eb-85a7-2ab88d67a8b6.png)

This started reproducing in [v1.42.0](https://github.com/stephenh/ts-proto/blob/master/CHANGELOG.markdown#v1420). Probably related to `ts-poet` upgrade.
Can you please help me understand what is wrong?

mharsat:
Also @stephenh, when running with `useDate=false` `Timestamp` functions are still generated, causing errors.
```
error TS2739: Type 'Date' is missing the following properties from type 'Timestamp': seconds, nanos

775       message.updatedDate = fromJsonTimestamp(object.updatedDate);
          ~~~~~~~~~~~~~~~~~~~

error TS2322: Type '{ seconds?: number | undefined; nanos?: number | undefined; }' is not assignable to type 'Timestamp'.
  Types of property 'seconds' are incompatible.
    Type 'number | undefined' is not assignable to type 'number'.
      Type 'undefined' is not assignable to type 'number'.

795       message.updatedDate = object.updatedDate;
          ~~~~~~~~~~~~~~~~~~~

error TS2339: Property 'toISOString' does not exist on type 'Timestamp'.

808         message.updatedDate !== undefined ? message.updatedDate.toISOString() : null);
                                                                    ~~~~~~~~~~~

error TS2304: Cannot find name 'fromTimestamp'.

830     return fromTimestamp(Timestamp.fromJSON(o));
               ~~~~~~~~~~~~~
```

stephenh:
Fixed in v1.55.0. Thanks @jessebutterfield !

mharsat:
Hey @stephenh @jessebutterfield thanks for the fix!
The issue I described in the second comment is still reproduced in v1.55.0 with the `useDate=false` option

stephenh:
@mharsat can you file a separate issue?

mharsat:
Sure @stephenh it's #208

PR #205, opened 2021-02-09
Add missing fromTimestamp function 

jessebutterfield:
fromTimestamp function definition was missing when outputJsonMethods is true but outputEncodeMethods is false.


jessebutterfield:
Fixes https://github.com/stephenh/ts-proto/issues/200

PR #206, opened 2021-02-09
Fix json value parsing for Long when forceLong is not set to long.

jessebutterfield:
I don't know if you want safer parsing when parsing json to Number, but currently when outputJsonMethods=true and forceLong is Number or String, it's producing Number.fromValue and String.fromValue which are undefined.


stephenh:
Huh, good catch; can you find ideally and existing `integration/` test that has a `parameters.txt` combination with what was breaking for you, make a change to its `*.proto` file to produce, re-run `update-bins.sh`, then `codegen.sh <name of integration directory>`, and show a) a reproduction and b) that your change fixes it?

I.e. https://github.com/stephenh/ts-proto/tree/master/integration/simple-long-string

Your PR is a pretty easy fix, so I'm confident it's right, but just want to have a test case to prevent regression.

Thanks!

jessebutterfield:
Before fix
`
    if (object.guint64 !== undefined && object.guint64 !== null) {
      message.guint64 = String.fromValue(object.guint64);
    } else {
      message.guint64 = undefined;
    }
    return message;
`
After fix
`
    if (object.guint64 !== undefined && object.guint64 !== null) {
      message.guint64 = String(object.guint64);
    } else {
      message.guint64 = undefined;
    }
    return message;
`

stephenh:
Okay, should be released in v1.58.0. Thanks!

PR #207, opened 2021-02-09
Allow generating JSON methods without the fromPartial method

mharsat:
For a large schema defined with many proto messages, generated code becomes rather large and increases the bundle size in the web environment. This allows generating the JSON methods and opting out of the `fromPartial` method in a non breaking way.


README.markdown
```diff
@@ -261,6 +261,10 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
 
   This is also useful if you want "only types".
 
+- With `--ts_proto_opt=outputPartialMethods=false`, the `Message.fromPartial` method for working with partial JSON-coded data will not be output.
```
stephenh, line 264: Fwiw the `fromPartial` methods are not for partially code JSON-data, they are for converting objects / object literals that programmers might have typed-out / made that look "basically like the protobuf message" but don't have every little thing right (like the default values applied for primitives).

That's still fine that you don't want them, but maybe word this as "the `Message.fromPartial` methods for accepting partially-formed objects/object literals will not be output".

README.markdown
```diff
@@ -261,6 +261,10 @@ protoc --plugin=node_modules/ts-proto/protoc-gen-ts_proto ./batching.proto -I.
...
+
+  Also, if `outputJsonMethods=false` is provided this method will not be generated.
```
stephenh, line 266: I know historically this was the case, but maybe we should just fully decouple `outputJsonMethods` and `outputPartialMethods` now that you've added this 2nd flag.

Maybe just remove this line, and let's have `outputJsonMethods` be completely separate from `outputPartialMethods`.

This will be a breaking change b/c users will have to also start setting `outputPartialMethods=false` in addition to `outputJsonMethods=false` but I think that's fine.

src/main.ts
```diff
@@ -116,7 +116,9 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
         }
         if (options.outputJsonMethods) {
           staticMethods.push(generateFromJson(ctx, fullName, message));
-          staticMethods.push(generateFromPartial(ctx, fullName, message));
+          if (options.outputPartialMethods) {
```
stephenh, line 119: Per other comment in the readme, let's move this check to be a sibling of the `if outputJsonMethods` block.

README.markdown
mharsat: I agree this should be decoupled, done

stephenh:
Okay, released as v1.67.0. Thanks!

PR #210, opened 2021-02-10
Add support for 'json_name' annotation.

fxb:
This PR adds support for the `json_name` annotation.

I went through calls to `maybeSnakeToCamel` and replaced it with calls to a new function called `getFieldName`, which will additionally inspect the `jsonName` field of `FieldDescriptorProto`.

A simple integration test is added as well.

I decided to not add a configurable option for this right now, but it would be trivial to do so.

According to the [proto3 language guide](https://developers.google.com/protocol-buffers/docs/proto3#json) parsers should to handle different field names:

> Parsers accept both the lowerCamelCase name (or the one specified by the json_name option) and the original proto field name.

As this currently does not seem to be implemented, I also skipped it for the `json_name` support.


marcushultman (approved):
🎉

integration/simple-json-name/simple.ts
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable */
+import { Writer, Reader } from 'protobufjs/minimal';
+
+export const protobufPackage = 'simple';
+
+export interface Simple {
+  other_name: string;
```
stephenh, line 7: Hm, my interpretation of `json_name` attribute would be that this particular name would still be `name` and then `other_name` would only show up on the JSON on the wire (i.e. in `toJson` methods) and then looked for when data is parsed in from the wire (i.e. in the `fromJson` methods).

Can we do that instead?

stephenh:
Fixed in #408

PR #211, opened 2021-02-11
Add support for useDate=false

### LINKED ISSUE #208: Timestamp functions cause issues with useDate=false option
When running with `useDate=false`, the `Timestamp` functions are still generated, causing errors.
```
error TS2739: Type 'Date' is missing the following properties from type 'Timestamp': seconds, nanos

775       message.updatedDate = fromJsonTimestamp(object.updatedDate);
          ~~~~~~~~~~~~~~~~~~~

error TS2322: Type '{ seconds?: number | undefined; nanos?: number | undefined; }' is not assignable to type 'Timestamp'.
  Types of property 'seconds' are incompatible.
    Type 'number | undefined' is not assignable to type 'number'.
      Type 'undefined' is not assignable to type 'number'.

795       message.updatedDate = object.updatedDate;
          ~~~~~~~~~~~~~~~~~~~

error TS2339: Property 'toISOString' does not exist on type 'Timestamp'.

808         message.updatedDate !== undefined ? message.updatedDate.toISOString() : null);
                                                                    ~~~~~~~~~~~

error TS2304: Cannot find name 'fromTimestamp'.

830     return fromTimestamp(Timestamp.fromJSON(o));
               ~~~~~~~~~~~~~
```

ethanfrey:
Closes #208 

- [x] build example `use-date-false` that demonstrates this case
- [x] write unit tests 
- [x] fix codegen

Note for devs:

* First you must run the setup as described in https://github.com/stephenh/ts-proto/blob/master/README.markdown#building
* After changing the `*.proto` file you need `cd integrations && ./update-bins.sh` 
  * You want update `protoc --version` to be `3.14.0` for this to run properly
* After changing the codegen logic, you need `cd integrations && ./codegen.sh`
  * `../node_modules/.bin/ts-node ./codegen.ts ./use-date-false ./use-date-false/metadata.bin useDate=false` is faster to generate just one file, and works on macos


src/main.ts
```diff
@@ -576,7 +588,9 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
       readSnippet = code`${type}.decode(reader, reader.uint32()).value`;
     } else if (isTimestamp(field)) {
       const type = basicTypeName(ctx, field, { keepValueType: true });
-      readSnippet = code`${utils.fromTimestamp}(${type}.decode(reader, reader.uint32()))`;
+      readSnippet = options.useDate
```
stephenh, line 591: Hm, I think it might be a simpler change to leave this line as it was, and then change line 589's "if isTimestamp(field)" to include "&& useDate". 

That way we'll only get into this `else if` if its timestamp & use date is true, otherwise we'll go into the `isMessage(field)` block next, which I think should handle `Timestamp`'s just fine, i.e. it will basically do that `${type}.decode` that currently you have re-typed-out as a the `:` case.

...could probably do that for most of these `if isTimestamp(field)` changes?

src/main.ts
```diff
@@ -883,7 +899,9 @@ function generateToJson(ctx: Context, fullName: string, messageDesc: DescriptorP
           ? code`${from} !== undefined ? ${toJson}(${from}) : undefined`
           : code`${toJson}(${from})`;
       } else if (isTimestamp(field)) {
-        return code`${from} !== undefined ? ${from}.toISOString() : null`;
+        return options.useDate
+          ? code`${from} !== undefined ? ${from}.toISOString() : null`
+          : code`${from} !== undefined ? ${utils.fromTimestamp}(${from}).toISOString() : null`;
```
stephenh, line 904: Ah cool, keeping ISO strings on the wire...

integration/use-date-false/metadata.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+import "google/protobuf/timestamp.proto";
+
+message Metadata {
+    google.protobuf.Timestamp last_edited = 1;
```
stephenh, line 6: This is a great small repro, thanks!

src/main.ts
stephenh: I wonder if `isTimestamp` should be renamed to `isTimestampAsDate`, and have it take both `options` & `field`, because (...in theory...?) everywhere that `isTimestamp` is called today, we really mean "...so that we can do special `Date` handling".

I guess the JSON serialization is the exception to that rule, where we want to detect `Timestamp` and do something special even if its not a `Date`.

ethanfrey:
Thanks for the review, it will probably be Monday or Tuesday til we get to the fixes, but happy to make them

src/main.ts
willclarktech: OK, I've simplified the code where I can and updated the codegen. Since we need `isTimestamp` to be separable from `options.useDate` for the JSON serialisation, I guess we could add a wrapper function that checks both, but I don't think that adds much value.

ethanfrey:
Thanks for the updates Will. Is this blocked on anything @stephenh or just time to review it?

Issue #213, opened 2021-02-11
Omit scalar fields with default value


webmaster128:
According to the proto3 Language Guide, [default values should be omitted](https://developers.google.com/protocol-buffers/docs/proto3#default) during encoding:

> Also note that if a scalar message field is set to its default, the value will not be serialized on the wire.

This behaviour is implemented in a variety of protobuf encoders and it would be great to have it here as well.

The encoder from code generated by protobuf.js looks like this:

```ts
SignDoc.encode = function encode(m, w) {
    if (!w) w = $Writer.create();
    if (m.bodyBytes != null && Object.hasOwnProperty.call(m, "bodyBytes"))
    w.uint32(10).bytes(m.bodyBytes);
    if (m.authInfoBytes != null && Object.hasOwnProperty.call(m, "authInfoBytes"))
    w.uint32(18).bytes(m.authInfoBytes);
    if (m.chainId != null && Object.hasOwnProperty.call(m, "chainId")) w.uint32(26).string(m.chainId);
    if (m.accountNumber != null && Object.hasOwnProperty.call(m, "accountNumber"))
    w.uint32(32).uint64(m.accountNumber);
    return w;
};
```

This does not conform to the spec as well, but allows applications to implement the ommiting themselves.

After the migration to ts-proto the same message type is encoded like this:

```ts
export const SignDoc = {
  encode(message: SignDoc, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
    writer.uint32(10).bytes(message.bodyBytes);
    writer.uint32(18).bytes(message.authInfoBytes);
    writer.uint32(26).string(message.chainId);
    writer.uint32(32).uint64(message.accountNumber);
    return writer;
  },

  // ...
}
```

In contrast to the previous version, we don't have `| null | undefined` types in the fields anymore, which is good. But for scalar values we can and should check for the default values.

- `double`/`float`/`int32`/`int64`/`uint32`/`uint64`/`sint32`/`sint64`/`fixed32`/`fixed64`/`sfixed32`/`sfixed64`: 0
- `bool`: false
- `string`: ""
- `bytes`: 0-length

stephenh:
Yep, this sgtm, I just haven't historically cared about this enough to implement (since it was optional). A PR would be great.

webmaster128:
[...]

PR #214, opened 2021-02-12
Remove redundant default check for messages

webmaster128:
I came across this when working on #213.

Since `${defaultValue(ctx, field)}` is always `undefined` for messages, this generates the ame test code twice.


stephenh:
Ah thanks! That'd been bothering me for awhile but not enough to track it down. Glad to have it fixed!

PR #215, opened 2021-02-15
Omit default values

### LINKED ISSUE #213: Omit scalar fields with default value

webmaster128:
Closes #213

This unfortunately got a little bigger than I anticipated because pbjs does not support decoding map entries with a key but no value.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -9,6 +9,7 @@ import {
   basicLongWireType,
   basicTypeName,
   basicWireType,
+  notDefaultCheck,
   defaultValue,
   detectMapType,
   getEnumMethod,
@@ -642,6 +643,7 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
     encode(
       ${messageDesc.field.length > 0 ? 'message' : '_'}: ${fullName},
       writer: ${Writer} = ${Writer}.create(),
+      forceDefaultSerialization = false,
     ): ${Writer} {
   `);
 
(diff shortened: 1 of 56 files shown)


src/main.ts
```diff
@@ -649,38 +652,48 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
-      writeSnippet = (place) => code`${type}.encode(${place}, writer.uint32(${tag}).fork()).ldelim()`;
+      writeSnippet = (place, forceDefaultSerialization) =>
+        code`${type}.encode(${place}, writer.uint32(${tag}).fork(), ${forceDefaultSerialization}).ldelim()`;
     } else {
       throw new Error(`Unhandled field ${field}`);
     }
 
     if (isRepeated(field)) {
       if (isMapType(ctx, messageDesc, field)) {
+        // protobuf.js does not support map entries with keys set and values unset.
```
stephenh, line 680: Hm, just curious, but how does protobuf.js not support unset values? I.e. what would break if we did serialize this with a key set but value unset?

webmaster128: I'm glad you ask. It felt like a bug in pbjs to me. I tried to reproduce it and it turns out I cannot reproduce it anymore. Maybe outdated pbjs code during development 🤷‍♂️

I was able to remove all the special case handling, making all of this much nicer.

stephenh:
Awesome, thanks! This should be released in v1.64.0.

Issue #219, opened 2021-02-17
Exclude nestjs decorator


ardyfeb:
I currently generating typescript from my proto files, but i only use it as grpc client. How to exclude decorator when outputing to nestjs ?

ardyfeb:
And `returnObservable` is not work

Issue #157, opened 2020-10-21
returnObservable option not being honored


tonyneel923:
When I pass returnObservable=false into the options flag it still uses observable. 

Example command:

`protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_opt=returnObservable=false,outputEncodeMethods=false,outputJsonMethods=false,outputClientImpl=false  --ts_proto_out=typescript ${path}`

tonyneel923:
After more digging this only happens when return types use stream. Is there any wah to not use an external lib to represent stream in typescript? Sorry new to grpc so not sure.

stephenh:
@tonyneel923 "not use an external lib to represent stream" right, there is not a built-in stream type, so we have to pull in a 3rd party library. Also since ts-proto just does codegen, it's not ts-proto really deciding whether to use rxjs or not, it's the nestjs and/or grpc-web runtimes that choose what streaming library they use, and then ts-proto just outputs code that uses the appropriate type.

PR #220, opened 2021-02-17
returnObservable should not depends on grpc-web

ardyfeb:
since ts-proto just does codegen, it should follow user configuration

https://github.com/stephenh/ts-proto/issues/157
https://github.com/stephenh/ts-proto/issues/219#issuecomment-780834036


stephenh:
I'm pretty sure this limitation is on purpose because the non-grpc-web code currently does not support `returnObservable`, so even though your PR "turns it on", the generated code very likely does not work.

Can you confirm that is true? Or does this actually produce working code?

If it does produce working code, it'd be great to make a new `integration/return-observable` directory with a minimal `*.proto` and `parameters.txt` that shows what the output is / that it works.

ardyfeb:
Hy thank you for your response, yes some case returning Observable will be usefull even user is not on grpc-web context

I have test it on my internal code it's works. but sorry i forgot to write the test. i will push the test ASAP

PR #221, opened 2021-02-18
feat: Add support for useDate=string

eqyiel:
This PR builds on the best parts of #211 and #152 and makes the `useDate` flag have three states, like suggested [here](https://github.com/stephenh/ts-proto/pull/152#discussion_r503011417).

My use case is a bit unusual because I use this library to generate _types only_ for use with a [gRPC ↔ ️JSON gateway](https://github.com/grpc-ecosystem/grpc-gateway) (no client implementation!) that follows the [canonical JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json), and *what I really want is the option to have timestamps be strings in interfaces*.

Because of that, _I don't know if these encode and decode methods for `useDate=string` make sense_. I'd love to get some feedback from @actuosus and @jessebutterfield about this (because of their involvement in #152)!

> Why absorb the changes in #211?

There are a few reasons:

1. I couldn't touch this "tri-state" thing without worrying about breaking other use cases.
2. I really want this library to support `useDate=string` :grin:

If #211 gets merged first, I'll update this. 🤞

Otherwise, I put appropriate `@co-authored-by` attributions.


integration/use-date-true/google/protobuf/timestamp.ts
```diff
@@ -0,0 +1,210 @@
...
+ * http://www.joda.org/joda-time/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTime%2D%2D
+ * ) to obtain a formatter capable of generating timestamps in this format.
+ */
+export interface Timestamp {
+  /**
+   * Represents seconds of UTC time since Unix epoch
+   * 1970-01-01T00:00:00Z. Must be from 0001-01-01T00:00:00Z to
+   * 9999-12-31T23:59:59Z inclusive.
+   */
+  seconds: number;
```
eqyiel, line 97: Is this value wrong? According to [this](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#timestamp), `seconds` can be `int64` which can be `string` or `Long` (depending on other options). I think it's not related to this change though 🤔

stephenh:
Ah shoot, sorry, I forgot about this PR and already merged #211; can you rebase on top of master?

> no client implementation!

Fwiw just checking that for this to work, you can't use primitives w/o default values (like `0` / empty string) being "lost" and showing up as `undefined`, per my comment here: 

https://github.com/stephenh/ts-proto/pull/152#issuecomment-779356557

Are you able to handle/work around that?

integration/use-date-true/google/protobuf/timestamp.ts
jessebutterfield: forceLong defaults to number so I think this is correct according to https://github.com/stephenh/ts-proto#number-types

src/types.ts
```diff
@@ -225,6 +225,9 @@ export function defaultValue(ctx: Context, field: FieldDescriptorProto): any {
         return 'new Uint8Array()';
       }
     case FieldDescriptorProto_Type.TYPE_MESSAGE:
+      if (isTimestamp(field) && options.useDate === DateOption.STRING) {
```
jessebutterfield, line 228: See my comment about the test, I think this is incorrect and can just be removed so it still defaults to undefined.

integration/use-date-string/use-date-string.ts
```diff
@@ -0,0 +1,253 @@
...
+    message.repeatedTimestamp = [];
+    message.mapOfTimestamps = {};
+    while (reader.pos < end) {
+      const tag = reader.uint32();
+      switch (tag >>> 3) {
+        case 1:
+          message.id = reader.string();
+          break;
+        case 2:
+          message.timestamp = reader.string();
```
jessebutterfield, line 54: I think this should decode the timestamp using Timestamp.decode(reader, reader.uint32()) and then convert it to string.

src/main.ts
```diff
@@ -572,9 +584,11 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
     } else if (isValueType(ctx, field)) {
       const type = basicTypeName(ctx, field, { keepValueType: true });
       readSnippet = code`${type}.decode(reader, reader.uint32()).value`;
-    } else if (isTimestamp(field)) {
+    } else if (isTimestamp(field) && options.useDate === DateOption.DATE) {
       const type = basicTypeName(ctx, field, { keepValueType: true });
       readSnippet = code`${utils.fromTimestamp}(${type}.decode(reader, reader.uint32()))`;
+    } else if (isTimestamp(field) && options.useDate === DateOption.STRING) {
+      readSnippet = code`reader.${toReaderCall(FieldDescriptorProto_Type.TYPE_STRING)}()`;
```
jessebutterfield, line 591: I think this needs a utility like above to convert from the protobuf message Timestamp type to string. Maybe just modify the fromTimestamp function so that it returns a string if useDate===DateOption.String and just leave this code as it was.

src/main.ts
```diff
@@ -655,12 +669,16 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
       const tag = ((field.number << 3) | basicWireType(field.type)) >>> 0;
-      writeSnippet = (place) => code`writer.uint32(${tag}).${toReaderCall(field)}(${place})`;
-    } else if (isTimestamp(field)) {
+      writeSnippet = (place) => code`writer.uint32(${tag}).${toReaderCall(field.type)}(${place})`;
+    } else if (isTimestamp(field) && options.useDate === DateOption.DATE) {
       const tag = ((field.number << 3) | 2) >>> 0;
       const type = basicTypeName(ctx, field, { keepValueType: true });
       writeSnippet = (place) =>
         code`${type}.encode(${utils.toTimestamp}(${place}), writer.uint32(${tag}).fork()).ldelim()`;
+    } else if (isTimestamp(field) && options.useDate === DateOption.STRING) {
```
jessebutterfield, line 678: Same as above, this needs to convert from a string to a protobuf timestamp format rather than just writing the string.

jessebutterfield (commented):
@eqyiel Thanks so much for working on this. Looks good. I had a couple of suggestions for fixing the encode/decode methods and the fromJson method. Let me know if you have any questions about my suggestions.

integration/use-date-false/google/protobuf/timestamp.ts
```diff
@@ -56,16 +56,7 @@ export const protobufPackage = 'google.protobuf';
...
- *
- *     Instant now = Instant.now();
- *
- *     Timestamp timestamp =
- *         Timestamp.newBuilder().setSeconds(now.getEpochSecond())
- *             .setNanos(now.getNano()).build();
- *
- *
- * Example 6: Compute Timestamp from current time in Python.
+ * Example 5: Compute Timestamp from current time in Python.
```
eqyiel, line 59: I guess this file (and `observable.bin` and `metadata.bin`) changed because I used a different version of `protoc` than when they were compiled. I checked in these changes because after `./update-bins.sh && ./codegen.sh && npm run test -- -u` there are no other unstaged files. It seems that this makes it consistent 😇 

```console
~/git/ts-proto/integration feat/support-date-as-string*
❯ protoc --version
libprotoc 3.13.0
```

src/main.ts
```diff
@@ -380,51 +380,68 @@ function makeTimestampMethods(options: Options, longs: ReturnType<typeof makeLon
...
-      function toTimestamp(date: Date): ${Timestamp} {
-        const seconds = ${seconds};
-        const nanos = (date.getTime() % 1_000) * 1_000_000;
-        return { seconds, nanos };
-      }
-    `
+    options.useDate === DateOption.STRING
+      ? code`
+          function toTimestamp(dateStr: string): ${Timestamp} {
+            const date = new Date(dateStr);
```
eqyiel, line 386: I felt this was not smart because the variable named `date` can't be changed due to stuff outside of this function: https://github.com/eqyiel/ts-proto/blob/c4977a2a19ad565766d581c4e3d733d1667de150/src/main.ts#L371-L379

I don't think it's _that_ much worse though, because that was already the case 😇

eqyiel (commented):
@stephenh @jessebutterfield thanks for reviewing, please take another look when you get time 😁

jessebutterfield (approved):
Looks good to me. Suggested one possible simplification but overall looks great.

src/main.ts
```diff
@@ -910,10 +937,12 @@ function generateToJson(ctx: Context, fullName: string, messageDesc: DescriptorP
         return isWithinOneOf(field)
           ? code`${from} !== undefined ? ${toJson}(${from}) : undefined`
           : code`${toJson}(${from})`;
-      } else if (isTimestamp(field)) {
-        return options.useDate
-          ? code`${from} !== undefined ? ${from}.toISOString() : null`
-          : code`${from} !== undefined ? ${utils.fromTimestamp}(${from}).toISOString() : null`;
+      } else if (isTimestamp(field) && options.useDate === DateOption.DATE) {
+        return code`${from} !== undefined ? ${from}.toISOString() : null`;
```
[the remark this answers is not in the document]
eqyiel: https://github.com/stephenh/ts-proto/pull/221#discussion_r605528580

src/main.ts
```diff
@@ -910,10 +937,12 @@ function generateToJson(ctx: Context, fullName: string, messageDesc: DescriptorP
...
+      } else if (isTimestamp(field) && options.useDate === DateOption.STRING) {
+        return code`${from}`;
+      } else if (isTimestamp(field) && options.useDate === DateOption.TIMESTAMP) {
```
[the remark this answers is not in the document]
eqyiel: https://github.com/stephenh/ts-proto/pull/221#discussion_r605528580

Issue #230, opened 2021-03-02
Add an option for const enums generation


lxgreen:
As Typescript docs say, `const enum` members are inlined at use sites. For example, the following TS code

```
export enum DividerData_DividerType {
  DOUBLE = 'DOUBLE',
  SINGLE = 'SINGLE',
  DASHED = 'DASHED',
  DOTTED = 'DOTTED',
  UNRECOGNIZED = 'UNRECOGNIZED',
}

const divider = DividerData_DividerType.DASHED;
```

is translated to the following JS:

```
export var DividerData_DividerType;
(function (DividerData_DividerType) {
    DividerData_DividerType["DOUBLE"] = "DOUBLE";
    DividerData_DividerType["SINGLE"] = "SINGLE";
    DividerData_DividerType["DASHED"] = "DASHED";
    DividerData_DividerType["DOTTED"] = "DOTTED";
    DividerData_DividerType["UNRECOGNIZED"] = "UNRECOGNIZED";
})(DividerData_DividerType || (DividerData_DividerType = {}));

const divider = DividerData_DividerType.DASHED;
```

while the same example using `const enum`:

```
export const enum DividerData_DividerType {
  DOUBLE = 'DOUBLE',
  SINGLE = 'SINGLE',
  DASHED = 'DASHED',
  DOTTED = 'DOTTED',
  UNRECOGNIZED = 'UNRECOGNIZED',
}

const divider = DividerData_DividerType.DASHED;
```

is translated to just
```
const divider = "DASHED";
```

As you can see, this can have a dramatic effect on bundle size. While there are some cases when the enums have to be preserved in JS, they are not required in many cases.

stephenh:
Sure, makes sense. I'd be tempted to just make enums always `const` but then at some point I'll get an issue filed that someone wants them _not_ `const`. :-)

Can you submit a PR with the new flag and an example of a mini proto file + `parameters.txt` in the `integration/` directory?

stephenh:
Released as v1.71.0. Thanks!

PR #231, opened 2021-03-02
constEnums option

lxgreen:
This PR adds an integration test case for `const enum` generation option (#230)

### DIFF
--- a/integration/const-enum/const-enum.proto
+++ b/integration/const-enum/const-enum.proto
@@ -1,0 +1,12 @@
+syntax = "proto3";
+
+message DividerData {
+    enum DividerType {
+        DOUBLE = 0;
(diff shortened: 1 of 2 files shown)


stephenh:
Looks great! I realize in retrospect my comment was ambiguous, but could you go ahead and take a crack at changing `main.ts` to output the enum with the `const` keyword?

It'll be somewhere around here:

https://github.com/stephenh/ts-proto/blob/master/src/enums.ts#L22

And then `options.ts` for adding the new `constEnums` option.

lxgreen:
sure! you just had to tell me, just fix it! 😄

lxgreen:
full solution in #232

stephenh:
Looked great, thanks!

PR #238, opened 2021-03-11
Add missing bytes fields to base objects

### LINKED ISSUE #237: bytes fields are missing in base objects
For the proto message

```proto
message MsgInstantiateContractResponse {
  // Address is the bech32 address of the new contract instance.
  string address = 1;
  // Data contains base64-encoded bytes to returned from the contract
  bytes data = 2;
}
```

The following base bessage is generated:

```ts
const baseMsgInstantiateContractResponse: object = { address: "" };
```

Since the `.decode` method only overrides field numbers that are serialized, the `data` field is not set when it is not serialized.

E.g. the serialization `0a2b7761736d317674796d7a6b6b65753777387578666e687961676a6b6b70653278676a303776396371707574`:

```sh
$ echo 0a2b7761736d317674796d7a6b6b65753777387578666e687961676a6b6b70653278676a303776396371707574 | xxd -r -p | protoc --decode_raw
1: "wasm1vtymzkkeu7w8uxfnhyagjkkpe2xgj07v9cqput"
```

results in the object

```js
{
  address: 'wasm1vtymzkkeu7w8uxfnhyagjkkpe2xgj07v9cqput',
}
```

instead of the expected

```js
{
  address: 'wasm1vtymzkkeu7w8uxfnhyagjkkpe2xgj07v9cqput',
  data: new Uint8Array(),
}
```

which is defined in the interface

```ts
export interface MsgInstantiateContractResponse {
  /** Address is the bech32 address of the new contract instance. */
  address: string;
  /** Data contains base64-encoded bytes to returned from the contract */
  data: Uint8Array;
}
```

webmaster128:
Closes #237

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -13,6 +13,7 @@ const baseMessage: object = {};
 export const Message = {
   fromJSON(object: any): Message {
     const message = { ...baseMessage } as Message;
+    message.data = new Uint8Array();
     if (object.data !== undefined && object.data !== null) {
       message.data = bytesFromBase64(object.data);
     }
--- a/integration/bytes-node/point.ts
+++ b/integration/bytes-node/point.ts
@@ -22,6 +22,7 @@ export const Point = {
     const reader = input instanceof Uint8Array ? new Reader(input) : input;
     let end = length === undefined ? reader.len : reader.pos + length;
     const message = { ...basePoint } as Point;
+    message.data = Buffer.alloc(0);
     while (reader.pos < end) {
       const tag = reader.uint32();
       switch (tag >>> 3) {
@@ -38,6 +39,7 @@ export const Point = {
 
   fromJSON(object: any): Point {
     const message = { ...basePoint } as Point;
+    message.data = Buffer.alloc(0);
     if (object.data !== undefined && object.data !== null) {
       message.data = Buffer.from(bytesFromBase64(object.data));
     }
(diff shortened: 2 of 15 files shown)


integration/bytes-as-base64/message.ts
```diff
@@ -5,7 +5,7 @@ export interface Message {
   data: Uint8Array;
 }
 
-const baseMessage: object = {};
+const baseMessage: object = { data: new Uint8Array() };
```
stephenh, line 8: @webmaster128 is `Uint8Array` immutable? I'd assumed it wasn't, and being in the `baseMessage` would mean this this is shared across multiple `Message`s and so they'd end up sharing each other's `data` instance?

webmaster128: Nice catch, you're right. Uint8Array is mutable. I forgot about that because we never mutate Uint8Arrays.

Are you okey with turning baseMessage into a function that creates instances on demand because of this? Or do you have a better idea?

stephenh: Cool. Maybe keep `baseMessage` as-is and add a new line  in the various `decode` / `fromJSON` methods that does `message.data = new Uint8Array()`?

Similar to what we do for lists and maps, where there are a few lines at the start of `decode` that initialize them to empty values.

willclarktech: @stephenh I updated this PR to remove the mutable `Uint8Array`s and `Buffer`s from all the base objects (like how lists are treated), which I'm not completely sure is what you meant here. I also adjusted the `fromPartial` methods for consistency. Happy to update if any of this is wrong.

stephenh: @willclarktech cool! Fwiw did your changes to `main.ts` not get pushed up? Looking at the diff + the 3rd commit you pushed to the branch, it looks like maybe only the integration test changes got pushed? Or maybe I'm missing it.

willclarktech: @stephenh Whoops! I've updated, and actually changed my mind about `fromPartial`, so that's no longer included in this change.

PR #239, opened 2021-03-11
Fix setup script in package.json

webmaster128:
and run it


stephenh:
Cool! Thanks for fixing the `setup` thing.

Tangentially, I think there is some system/version-of-`protoc`-dependent behavior in terms of the `.bin` files that get generates, i.e. all of them changing, so I'll probably re-run `setup` locally before merging to selfishly have it use whatever version of `protoc` I happen to have (3.13.0 looks like), and my assumption is that will undo most of these diffs.

Not sure the best way to isolate the `protoc` thing ... I guess docker? Haven't gotten to that yet.

PR #240, opened 2021-03-12
Use qualified imports for grpc-web types

asarium:
I had some issues with using the generated code in combination with
babel-loader in my webpack setup. The grpc-web output was producing
references to .d.ts files within @improbable-eng/grpc-web. This broke
compilation with babel-loader since that has no concept of type
definition files.

This fixes that by simply using the fully qualified types which are
accepted by babel-loader.


stephenh:
Released as v1.74.0. Thanks!

PR #247, opened 2021-03-16
Fix probobufjs long initialisation 

### LINKED ISSUE #246: protobufjs may not be initialised if forceLong=long
### Expected Behaviour
A snippet of code is appended to generated files that conditionally initialises `util.long` in protobufjs if it has not been set to use `long`. This happens as expected when `forceLong` is set to `number`, `string` or omitted. 

### Observed Behaviour
The snippet of code is omitted if ts-proto is set to use the `long` type. In certain deployments, protobufjs can default to use the `number` type, which is then  type-asserted to `long` using the `as` operator by ts-proto, making such issues difficult to detect. I personally encountered this issue while working on a react app. 

### How to Reproduce
Compile any protobuf message containing an int64 field with the `forceLong=long` option set. The snippet will not be present at the bottom of the file. If the `forceLong` option is set to `string`, `number` or omitted altogether, the snippet will be present as expected. 

### Impact
Precision can be lost in a difficult-to-detect manner, certain methods expected by ts-proto might not exist on the resulting type even though the type appears to be correct (due to the aforementioned type assertion). For example, if a Protobuf message containing an optional int64 is decoded and then encoded again, the `isZero()` method will be called on the `number` type, which does not exist. 

### Fix 
Include the snippet of code regardless of which type is used to represent numbers (it's required in every scenario). I will submit a pull request shortly.

daw1012345:
Fixes #246

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -200,7 +200,7 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
       } else if (v instanceof Code) {
         return code`${v}`;
       } else {
-        return code ``;
+        return code``;
       }
     })
   );
--- a/src/main.ts
+++ b/src/main.ts
@@ -58,6 +58,7 @@ import { visit, visitServices } from './visit';
 import { EnvOption, LongOption, OneofOption, Options } from './options';
 import { Context } from './context';
 import { generateSchema } from './schema';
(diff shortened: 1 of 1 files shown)


stephenh:
Cool, thanks for the PR, can you add a new directory in `integration/` that is a test case with a minimal `*.proto` that reproduces the issue? Just to prevent regressions. Thanks!

PR #262, opened 2021-04-02
chore: Link to the grpc-js tests.

stephenh:


integration/grpc-js/grpc-js-test.ts
```diff
@@ -2,17 +2,20 @@
...
 describe('grpc-js-test', () => {
   it('compiles', () => {
-    expect(TestService).not.toBeUndefined();
+    expect(TestServiceService).not.toBeUndefined();
   });
 
   it('can create a server and a client', async () => {
     const server = new Server();
 
-    const impl: TestServer = {
```
stephenh, line 15: @aikoven I was adding a link to these tests in the readme, and just while reading the code again, thought that `TestServer` didn't really jump out at me as "the server-side implementation of our rpc service", vs. some sort of mocking/testing utility.

So I renamed the `service Test` in the proto file to `service TestService`, which afaiu is also a ~somewhat common naming convention for services (correct me if wrong though).

With this change, everything still works fine, but now the definition is `TestServiceService` which is kinda odd.

wdyt about changing the definition from `${name}Service` to `${name}Definition`?

Then some users would see `TestServiceDefinition` and others would see `TestDefinition`, depending on their preferred naming convention.

I don't want to do that in this PR, but just got me thinking about it.

aikoven: I don't see that it's common to add a suffix to service names in `.proto`. See examples in grpc repo:

https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto
https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1alpha/reflection.proto

See also the protobufs of etcd:

https://github.com/etcd-io/etcd/blob/master/api/etcdserverpb/rpc.proto

Moreover, I'm not sure that it's fully correct to add a `Service` suffix to the service definition constant name in generated code (just like they do for grpc-js in [grpc-tools](https://www.npmjs.com/package/grpc-tools)), since there would be a conflict e.g.:

```proto
service Foo {}
message FooService {}
```

Though this looks like a marginal case.

stephenh: > I don't see that it's common to add a suffix to service names

Ah nice! Thanks for looking around to get actual data points; I think we had done this at a prior job I worked at, so that biased my original thought, but sounds like we were a minority.

> Moreover, I'm not sure that it's fully correct to add a Service suffix to the service

Do you mean what we're doing here:

https://github.com/stephenh/ts-proto/blob/main/src/generate-grpc-js.ts#L58

This is the line I was proposing to change to:

```
export const ${def(`${serviceDesc.name}Definition`)}
```

I guess fwiw the grpc-web code currently does something similar but uses `${service}Desc`:

https://github.com/stephenh/ts-proto/blob/main/src/generate-grpc-web.ts#L74

So we could do `${service}Desc` as short for descriptor here too. Not a huge fan of abbreviations though. `${service}Desc`, `${service}Descriptor` (it's not really the full descriptor from the proto file), `${service}Definition`, ${service}Def`? Dunno, what do you think?

I'm fine keeping it as-is, I was just trying to make the test a little more newbie-friendly to read, since it's our only form of docs right now :-), but we can accomplish that w/o doing the `Test` -> `TestService` rename in the proto, so I'll close this out.

aikoven: I would prefer no suffix at all, just like there's no suffix for message names.

In my PR for `grpc-js` support, I've added the `Service` suffix to more closely match the output of `grpc-tools`. Though now I acknowledge that this makes little sense, so we could go with any suffix or no suffix at all. If choosing between suffixes, I like `Definition` the most.

stephenh: > I would prefer no suffix at all, just like there's no suffix for message names.

Yep, I agree with that rationale.

I think there is a slight wrinkle here, insofar as this output isn't service itself (like with a message, the `interface Message` really is the message itself), but instead kinda some metadata about it, so lean towards the `Definition` suffix.

That said, I'm cool leaving it as-is, and then when/if we get around to it, moving over to the `Definition` suffix since that sounds good to you.

PR #263, opened 2021-04-02
chore: Add JSON.parse test case.

stephenh:


integration/use-date-string/use-date-string-test.ts
```diff
@@ -45,4 +45,24 @@ describe('useDate=string', () => {
...
+      repeatedTimestamp: [jan1.toISOString(), feb1.toISOString()],
+      mapOfTimestamps: {
+        jan1: jan1.toISOString(),
+        feb1: feb1.toISOString(),
+      },
+    };
+    const json = JSON.stringify(t1);
+
+    // The client-side can JSON.parse it
+    const t2 = JSON.parse(json);
```
stephenh, line 63: @eqyiel after much delay, I finally merged/released your `useDate` PR. Thanks for your patience on that one!

While going through it, I wanted to type out this test case, b/c AFAIU this is what you're trying to accomplish? I.e. get a JSON string off the wire, do a `JSON.parse`, and immediately/already have a `Todo`-compliant instance?

eqyiel: @stephenh thanks for asking, it looks good to me 😊

PR #268, opened 2021-04-04
fix: improve type registry types

aikoven:
I started integrating `ts-proto`, employing the new type registry feature, and discovered that the `$type` field was missing from `MessageType` type. This PR adds the missing field.


stephenh:
Nice, looks pretty cool!

Issue #273, opened 2021-04-16
Wrong generation when package not set


DK99:
The protobuf specs allow to not specify a package. Your generator assumes that a package is always defined.
e.g. there is this line of code in the grpc web generator `serviceName: "${fileDesc.package}.${serviceDesc.name}",` No matter the package it always puts a dot in front of the service name. This doen't work when dealing with an empty package name as the service name would be `.ExampleService`.

stephenh:
Should be out in the next release, thanks!

PR #274, opened 2021-04-16
fix: handle empty protobuf package

DK99:
See #273, the generators don't work correctly on empty package names. I fixed that by checking  if the package name is empty and using it when needed.


stephenh:
@DK99 thanks for the PR!

Can you add a new integration-test for this, i.e. make a new directory in `integration/` with a name like `no-proto-package`, and put a ~as-minimal-as-possible proto file that reproduces your issue?

Then include the generated output in the PR, to both review + function as a regression test.

Thanks!

DK99:
I have added one minimal integration test. I am unsure if that test covers all possibilities. To cover all possibilities it would be necessary to also compile this test for grpc-web.

stephenh:
Thanks @DK99 ! I wanted to make a small tweak and haven't learned how to push to existing PRs, so I created #285 and will release from there. Thanks!

PR #276, opened 2021-04-18
fix: Support string enum in repeated enum

kanziw:
In repeated enum type with `stringEnums` option, it didn't work in typescript.

e.g.
ts_error

I don't know if this is the right way to do it, but please check.


stephenh:
@kanziw great, thanks for the PR!

I think this fixes looks good.

Just to make sure, and prevent future regressions, can you include a change to `integration-simple-string-enums/simple.proto` to add a new repeated enum field to that test message?

Then rerun update-bins and codegen.sh to see the updated output, and include it in this PR. 

Thanks again!

kanziw:
Thanks, please check last commit again! @stephenh

src/main.ts
```diff
@@ -789,6 +796,16 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
             ${writeSnippet('v!')};
           }
         `);
+      } else if (isEnum(field) && options.stringEnums) {
+        const tag = ((field.number << 3) | basicWireType(field.type)) >>> 0;
```
stephenh, line 800: @kanziw I'm pretty sure that `| basicWriteType(...)` is wrong here, because we're encoding the repeated field as a packed repeated field, so that tag should just be `| 2`:

https://developers.google.com/protocol-buffers/docs/encoding

I've made this change in my #284 version of this fix, but let me know if I'm missing something.

stephenh:
I've merged/released this in #284 but let me know if it doesn't work as expected. Thanks!

PR #278, opened 2021-04-19
feat: Support bytes as base64 strings

danwsong:
Currently, `bytes` fields are converted to JavaScript `Uint8Array` types. So, given the following schema:
```protobuf
message Message {
    bytes data = 1;
}
```
The following TypeScript interface would be produced:
```typescript
interface Message {
    data: Uint8Array;
}
```
This pull request adds support for base64 strings in addition to the already supported Uint8Array representation:
```typescript
interface Message {
    data: Uint8Array | string;
}
```

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -5,7 +5,7 @@ import * as Long from 'long';
 export const protobufPackage = '';
 
 export interface Message {
-  data: Uint8Array;
+  data: Uint8Array | string;
 }
 
 const baseMessage: object = {};
@@ -61,7 +61,10 @@ function bytesFromBase64(b64: string): Uint8Array {
 
 const btoa: (bin: string) => string =
   globalThis.btoa || ((bin) => globalThis.Buffer.from(bin, 'binary').toString('base64'));
-function base64FromBytes(arr: Uint8Array): string {
+function base64FromBytes(arr: Uint8Array | string): string {
+  if (typeof arr === 'string') {
+    return arr;
+  }
   const bin: string[] = [];
   for (let i = 0; i < arr.byteLength; ++i) {
     bin.push(String.fromCharCode(arr[i]));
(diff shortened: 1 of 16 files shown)


stephenh:
Hey @danwsong , thanks for the PR!

Thinking about this some more, I don't think having:

```
interface Message {
    data: Uint8Array | string;
}
```

As output is a good thing, because it will compliant readers who will now have to `if` check several thing `message.data` access with "is this `Uint8Array` or `string`?". Which seems boilerplately/not fun.

I think I'd rather have something where writers could write "either `Uint8Array` or a `string`" to give them flexibility in writing, but readers would have the simplicity of "only `Uint8Array`".

The way to do this would be to not change `Message.data` directly, but instead probably change `Message.fromPartial` to allow any `byte` fields to be `Uint8Array | string`, and then inside of `fromPartial` realize which values are base64s and convert to the `Uint8Array`s.

Dunno, that seems tricky, but I think is clearer.

The other alternative would be to have _only_ `data: string` and have `byte` just always represented as base64 when in memory / in messages. And so reading bytes off the wire would be immediately base64'd, and writing bytes to the write would be-base64 and then write the bytes.  Dunno. That doesn't seem great either, b/c base64 version of things will be more expensive to keep in memory.

I dunno. Kinda wondering if just have `message.byte = fromBase64(...)` helper methods would be easier and just keep the interfaces/types as is. Sorry to come to that conclusion after you've already done the PR, but dunno, just thinking through things.

What do you think?

webmaster128:
> because it will compliant readers who will now have to if check several thing message.data access with "is this Uint8Array or string?". Which seems boilerplately/not fun.

I agree with this. Can we close here?

Opening up for `Uint8Array | string` inputs is a different story.

stephenh:
@webmaster128 yep, agreed, will close this out for now. Good point about accepting `Uint8array | string` in `fromPartial`.

PR #280, opened 2021-04-22
fix: long=string handling of timestamps

### LINKED ISSUE #248: using --ts_proto_opt=forceLong=string with google timestamp produces invalid Dates()
Generated code made this
```
function toTimestamp(date: Date): Timestamp {
  const seconds = (date.getTime() / 1_000).toString();
  const nanos = (date.getTime() % 1_000) * 1_000_000;
  return { seconds, nanos };
}
```
followed by
```
Writer.prototype.uint64 = function write_uint64(value) {
    var bits = LongBits.from(value);
    return this._push(writeVarint64, bits.length(), bits);
};
```
the seconds = part keeps the fraction in the string.
but LongBits.From(...) expects sanitized input, so just deletes the decimal point, and you end up with a seconds that's (potentially MUCH) higher than it should be.

```
  let seconds: string | Code = 'date.getTime() / 1_000';
  let toNumberCode = 't.seconds';
  if (options.forceLong === LongOption.LONG) {
    toNumberCode = 't.seconds.toNumber()';
    seconds = code`${longs.numberToLong}(date.getTime() / 1_000)`;
  } else if (options.forceLong === LongOption.STRING) {
    toNumberCode = 'Number(t.seconds)';
    seconds = '(date.getTime() / 1_000).toString()';
  }
```

Needs to change to throw away the decimal part before calling .toString()

Psykar:
When a second value on a timestamp is > 8 characters as a string, and
has a fractional component, the calculated Long value is incorrect.

Fixes this by truncating the fractional component before converting to
long (keeps the fractional component on the nanoseconds key instad)

I suspect I may have a different version of protobuf to what the maintainer is using, `npm run setup` modified a LOT of files. I've committed only those changes relevant, but you might want to run `npm run setup` with the correct versions and update this PR.

Closes #248


src/main.ts
```diff
@@ -409,7 +409,10 @@ function makeTimestampMethods(options: Options, longs: ReturnType<typeof makeLon
     seconds = code`${longs.numberToLong}(date.getTime() / 1_000)`;
   } else if (options.forceLong === LongOption.STRING) {
     toNumberCode = 'Number(t.seconds)';
-    seconds = '(date.getTime() / 1_000).toString()';
+    // Must discard the fractional piece here
+    // Otherwise the fraction ends up on the seconds when parsed as a Long
+    // (note this but would only occurs when the string is > 8 characters)
```
stephenh, line 414: Great comment, thanks!

stephenh (approved):
Awesome, thanks!

src/main.ts
Psykar: Aww man, you commented on it and now I noticed the horrible grammar.

Psykar: Resolved grammar issues in https://github.com/stephenh/ts-proto/pull/281/files

stephenh: Ha, I hadn't actually noticed :-) but thanks!

PR #282, opened 2021-04-23
fix: ignore `$type` field in `DeepPartial`

aikoven:
This enables using `fromPartial` with objects of different but structurally equivalent types.


stephenh:
Nice!

PR #283, opened 2021-04-23
fix: Only generate files in fileToGenerate

emil-e:
According to doc comments in `plugin.proto` from Google:

```
// The .proto files that were explicitly listed on the command-line.  The
// code generator should generate code only for these files.  Each file's
// descriptor will be included in proto_file, below.
repeated string file_to_generate = 1;
```

### DIFF
--- a/integration/codegen.ts
+++ b/integration/codegen.ts
@@ -8,6 +8,7 @@ import { prefixDisableLinter } from '../src/utils';
 import { getTsPoetOpts, optionsFromParameter } from '../src/options';
 import { Context } from '../src/context';
 import { generateTypeRegistry } from '../src/generate-type-registry';
+import { protoFilesToGenerate } from '../src/plugin';
 
 /**
  * Generates output for our integration tests from their example proto files.
@@ -31,7 +32,7 @@ async function generate(binFile: string, baseDir: string, parameter: string) {
   const options = optionsFromParameter(parameter || '');
   const typeMap = createTypeMap(request, options);
 
-  for (let file of request.protoFile) {
+  for (let file of protoFilesToGenerate(request)) {
     // Make a different utils per file to track per-file usage
     const utils = makeUtils(options);
     const ctx: Context = { options, typeMap, utils };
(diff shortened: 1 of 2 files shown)


integration/codegen.ts
```diff
@@ -31,7 +31,7 @@ async function generate(binFile: string, baseDir: string, parameter: string) {
   const options = optionsFromParameter(parameter || '');
   const typeMap = createTypeMap(request, options);
 
-  for (let file of request.protoFile) {
+  for (let file of request.fileToGenerate) {
```
stephenh, line 34: This causes compile errors due to the `generateFile` call on line 38 still wanting the `FileDescriptorProto` but `fileToGenerate` is just a `string[]`.

stephenh (changes requested):

integration/codegen.ts
emil-e: I will take a look!

emil-e:
@stephenh How about that?

src/plugin.ts
```diff
@@ -20,8 +20,12 @@ async function main() {
...
 
   const files = await Promise.all(
-    request.protoFile.map(async (file) => {
-      const [path, code] = generateFile(ctx, file);
+    request.fileToGenerate.map(async (fileName) => {
+      const descriptor = request.protoFile.find(desc => desc.name == fileName);
+      if (descriptor === undefined) {
+        throw new Error(`Cannot find descriptor for file-to-generate: ${fileName}`);
+      }
+      const [path, code] = generateFile(ctx, descriptor);
```
stephenh, line 28: Cool, this looks good. Just to make sure `src/plugin.ts` and `integration/codegen.ts` stay in sync (I've had bugs come up from them drifting before), how about:

* Add a new `function protoFilesToGenerate(request): FileDescriptorProto[]` in `src/plugin.ts` that returns `request.protoFile.filter(f => request.filesToGenerate.includes(f.name));`
* Make line 23 in `src/plugin.ts` be `protoFilesToGenerate(request).map(async file =>`
* Make line 34 in `codegen.ts` be `for (let file of protoFilesToGenerate(request)) {`

?

emil-e: That sounds reasonable, will make the change.

PR #290, opened 2021-05-08
feat: Bind service methods to class

codebutler:
This makes it easier to pass around service function references. For example, with this change you can use [react-query](https://react-query.tanstack.com/guides/mutations) directly: `useMutation(fooClient.bar);`


stephenh:
Ah cool! I wouldn't have thought of using it with react-query like that.

Looks like there is a minor lint error, can you fix and then I'll merge?

Thanks!

PR #291, opened 2021-05-08
fix: Fix integration test codegen script

codebutler:


stephenh:
Hey! This seems like a fine change, but out of curiosity what was broken before?

AFAICT this is moving where `protoFilesToGenerate` is located, but otherwise doesn't (shouldn't?) have any semantic changes?

codebutler:
> Hey! This seems like a fine change, but out of curiosity what was broken before?
> 
> AFAICT this is moving where `protoFilesToGenerate` is located, but otherwise doesn't (shouldn't?) have any semantic changes?

Hey sorry my commit description didn't get copied over into the PR. The problem is that plugin.ts tries to do stuff when imported. It calls process.exit and interrupts the codegen. https://github.com/stephenh/ts-proto/blob/main/src/plugin.ts#L59

stephenh:
Oh shoot! I was running this locally and like "it seems fine" but the wrinkle that "codegen exits so fails silently" was tricking me. :-) Thanks!

PR #292, opened 2021-05-08
fix: encode/decode wrap types

vcraescu:
```
syntax = "proto3";

option go_package = "api/blog";

service BlogService {
  rpc GetPost(google.protobuf.StringValue) returns (Post);
}

message Post {
  string ID = 1;
  string title = 2;
  google.protobuf.Timestamp createdAt = 3;
  google.protobuf.Timestamp updatedAt = 4;
}
```

When the service endpoint accepts a wrapped type value or returns a wrapped type value, the generated `grpc-js` client is broken because it tries to call the decode method on the string type. 

e.g. `string.encode(value).finish()` instead of `StringValue.encode(value).finish()`.

I'm not a typescript developer so please bear with me :)


stephenh:
Hey @vcraescu , thanks for the fix!

Can you change the `proto` file in `integration/grpc-js` to include a method that takes a wrapper type, to repro what was broken?

Then include the updated codegen code in the PR as a regression test.

(See the readme about update-bins.sh and codegen.sh, disclaimer codegen.sh was briefly broken until ~this morning, so make sure you have the latest rebased main.)

Thanks!

vcraescu:
@stephenh I'm trying to but I'm not able to `generate integration/grpc-js/simple.ts`. I ran the commands from the README.md but this is what I get:

```
$ ./codegen.sh
meta-typings/import_dir/thing.bin
simple-optionals/import_dir/thing.bin
simple-snake/import_dir/thing.bin
simple-unrecognized-enum/import_dir/thing.bin
simple/import_dir/thing.bin
type-registry/bar/bar.bin







```
Yes, I get a lot of blank lines at the end.

stephenh:
@vcraescu hm, that's weird, so there is a really hack "run these in parallel" approach in that shell script, so it's expected to see blank lines for ~a little bit while those first ~5-6 files run in parallel, but then it should move on to the next batch, i.e. I see:

```
$ ./codegen.sh
angular/simple-message.bin
avoid-import-conflicts/simple2.bin
avoid-import-conflicts/simple.bin
barrel-imports/bar.bin
barrel-imports/foo.bin
batching/batching.bin
batching-with-context/batching.bin
done
done
done
done
done
done
bytes-as-base64/message.bin
bytes-node/point.bin
const-enum/const-enum.bin
global-this/global-this.bin
grpc-js/simple.bin
grpc-web/example.bin
...then keeps going...
```

You can also invoke one file at a time, i.e. `./codegen.sh grpc-js/simple.bin` if for some reason `codegen.sh` just isn't working... 

Maybe it's a platform issue? I'm on linux and am not 100% sure if `codegen.sh` works on mac/WSL versions of bash.

vcraescu:
@stephenh 
```
➜ ./codegen.sh grpc-js/simple.bin
grpc-js/simple.bin/**/*.bin
Error: ENOTDIR: not a directory, open 'grpc-js/simple.bin/**/*.bin'
```

after I commented out: `shopt -s globstar` because it does not work on osx.

vcraescu:
@stephenh Please review this. Thanks!

stephenh:
@vcraescu out of curiosity, all of the deps in `yarn.lock` changed to `npmjs`:

```
"@babel/core@^7.1.0", "@babel/core@^7.7.5":
  version "7.9.6"
  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.9.6.tgz#d9aa1f580abf3b2286ef40b6904d390904c63376"
  resolved "https://registry.npmjs.org/@babel/core/-/core-7.9.6.tgz"
```

I'm good with the change, but just curious do you know why that happened? Like a different version of yarn?

vcraescu:
No idea, maybe because i used npm instead of yarn at the beginning. I will revert the changes and push it again.

stephenh:
@vcraescu this looks great! Fwiw I ran `update-bins` locally, which unfortunately depends on the local version of `protoc`, and that reduced a lot of the churn from the `*.bin` files changes.

I need to get that over to like invoking `protoc` in via docker container so that all contributors have hte same version.

PR #302, opened 2021-05-22
feat: implemented emitImportedFiles flag

### LINKED ISSUE #294: Latest versions don't import google proto files automatically
Trying to generate the messages with `google.protobuf` types used- there are links on them on code, but they are no `.ts` files generated for them anymore. Saw no breaking changes in latest 5 versions

doochik:
Fixes #294.
This commit puts changes from #283 behind the flag.

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -24,8 +24,9 @@ async function main() {
   const utils = makeUtils(options);
   const ctx: Context = { typeMap, options, utils };
 
+  const filesToGenerate = options.emitImportedFiles ? request.protoFile : protoFilesToGenerate(request);
   const files = await Promise.all(
-    protoFilesToGenerate(request).map(async (file) => {
+    filesToGenerate.map(async (file) => {
       const [path, code] = generateFile(ctx, file);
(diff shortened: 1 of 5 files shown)


src/plugin.ts
```diff
@@ -24,8 +24,9 @@ async function main() {
   const utils = makeUtils(options);
   const ctx: Context = { typeMap, options, utils };
 
+  const filesToGenerate = options.emitImportedFiles ? request.protoFile : protoFilesToGenerate(request);
   const files = await Promise.all(
-    protoFilesToGenerate(request).map(async (file) => {
+    filesToGenerate.map(async (file) => {
```
stephenh, line 29: Nice!

I've been bitten by this before, but can you update `integration/codegen.ts` as well? Either with the same logic you have here, or maybe just have it always use `request.protoFile`? I don't think we have an integration test that explicitly tests that `protoFilesToGenerate` codepath, which is fine I think...

stephenh (approved):
Thank you!

src/plugin.ts
doochik: HI!

I've updated `codegen.ts`. I think usage of `request.protoFile` should be enough fo tests.

PR #307, opened 2021-05-28
fix: grpc-js support for nestjs

### LINKED ISSUE #306: grpc-js dependency
So for me generating the types with `--ts_proto_opt=outputServices=grpc-js` apparently changes the output stubs, but still imports e.g. the `Metadata` etc. dependencies from `grpc` but shouldn't it include them from `@grpc/grpc-js` instead?

So I expect

```ts
import { Metadata } from '@grpc/grpc-js';
```
but I got

```ts
import { Metadata } from 'grpc';
```

so I guess that's probably a bug?

jsbrain:
fix #306


stephenh:
@jsbrain cool, looks like a good fix! Looks like there are a few minor failures in CI, can you take a look? Then I'll merge. Thanks!

jsbrain:
Sure, I'll have a look tomorrow :)

jsbrain:
@stephenh Ok I think that should be it. Had to update the snapshots because apparently Google updated their repo references for the timestamp proto so e.g. the [old timestamp link](github.com/golang/protobuf/ptypes/timestamp) now got [updated](google.golang.org/protobuf/types/known/timestamppb) so the old one is broken ...
Hope I didn't mess up something 😀

jsbrain:
@stephenh do we need something else to proceed?

feihan:
Since nestjs v8 is out now, this would be pretty important.
Is there any update? thanks!

PR #309, opened 2021-06-02
fix: close server stream on observer unsubscribe

justinschuldt:
I found that once a server stream is opened, there is no way to close it. Unsubscribing from the observable would not close the connection between the browser and the server.

This change adds a handler on the observer that will call the grpc.invoke's `close`, which will close the connection between the browser and the server.


stephenh:
Thanks @justinschuldt !

PR #316, opened 2021-06-09
feat: framework-agnostic service definitions

aikoven:
This PR adds support for generating rpc-framework-agnostic service definitions.

It allows to build an RPC framework that consumes `ts-proto` generated code, without the need to have framework-specific codegen beforehand.

A framework can still provide type-safe interfaces for the user using TypeScript mapped types.


integration/service-definitions/simple.ts
```diff
@@ -0,0 +1,293 @@
...
+    return message;
+  },
+};
+
+/**
+ * Test
+ *
+ * @deprecated
+ */
+export const TestDefinition = {
```
aikoven, line 84: I've added the `Definition` suffix as per [discussion](https://github.com/stephenh/ts-proto/pull/262#discussion_r606353216).

Though now it seems that the suffix should be `ServiceDefinition`

src/options.ts
```diff
@@ -36,7 +36,7 @@ export type Options = {
   stringEnums: boolean;
   constEnums: boolean;
   outputClientImpl: boolean | 'grpc-web';
-  outputServices: false | 'grpc-js';
+  outputServices: boolean | 'grpc-js';
```
aikoven, line 39: Not sure whether it should be a separate option. Perhaps someone might want to have both `grpc-js` and generic definitions

aikoven, line 39: Need to add documentation on the option to the Readme

integration/update-bins.sh
```diff
@@ -4,7 +4,7 @@
 # proto requests into .bin files that then unit tests can pull in directly as needed.
 list=$(find . -name "*.proto" -type f)
 
-for file in $list; do
+for file in service-definitions/*.proto; do
```
stephenh, line 7: Looks like this snuck in.

src/options.ts
stephenh: How about `outputServices: false | 'grpc-js' | 'generic-definitions` and then maybe someday (doesn't have to be now) we could support like csv of `outputServices=grpc-js,generic-definitions`?

stephenh (approved):

integration/service-definitions/simple.ts
aikoven: Hey @stephenh, thank you for review. What do you think about this one?

integration/update-bins.sh
aikoven: Fixed

src/options.ts
aikoven: Done

aikoven: Done

integration/service-definitions/simple.ts
aikoven: After some thinking, I'm fine with `Definition` suffix as long as we don't overload this term and only use it for services.

aikoven:
Hey @stephenh, any chance you could merge this PR soon? We seem to have settled on the changes. I've already made an integration in `nice-grpc`, and I am ready to publish it as soon as this feature is released.

stephenh:
@aikoven yep, thanks for the ping!

aikoven:
Thanks!

PR #332, opened 2021-06-29
fix: Consistently apply lowerCaseServiceMethods=true

### LINKED ISSUE #310: lowerCaseServiceMethods option does not change client implementation
With the `--ts_proto_opt=lowerCaseServiceMethods=true` option set, the plugin will output the service interfaces with lower case method names, but the client implementations (which implement the interface) are not modified. This means the client implementation does not match the interface and it throws TypeScript errors.

For example:

Given this proto:
```proto
service MathService {
    rpc Add(MathServiceAddRequest) returns (MathServiceAddResponse) {}
    rpc Subtract(MathServiceSubtractRequest) returns (MathServiceSubtractResponse) {}
    rpc Multiply(MathServiceMultiplyRequest) returns (MathServiceMultiplyResponse) {}
    rpc Divide(MathServiceDivideRequest) returns (MathServiceDivideResponse) {}
}
```

`ts-proto` will generate:
```typescript
export interface MathService {
  add(request: MathServiceAddRequest): Promise<MathServiceAddResponse>;
  subtract(
    request: MathServiceSubtractRequest
  ): Promise<MathServiceSubtractResponse>;
  multiply(
    request: MathServiceMultiplyRequest
  ): Promise<MathServiceMultiplyResponse>;
  divide(request: MathServiceDivideRequest): Promise<MathServiceDivideResponse>;
}

export class MathServiceClientImpl implements MathService {
  private readonly rpc: Rpc;
  constructor(rpc: Rpc) {
    this.rpc = rpc;
    this.Add = this.Add.bind(this);
    this.Subtract = this.Subtract.bind(this);
    this.Multiply = this.Multiply.bind(this);
    this.Divide = this.Divide.bind(this);
  }
  Add(request: MathServiceAddRequest): Promise<MathServiceAddResponse> {
    const data = MathServiceAddRequest.encode(request).finish();
    const promise = this.rpc.request(
      "services.math.v1.MathService",
      "Add",
      data
    );
    return promise.then((data) =>
      MathServiceAddResponse.decode(new _m0.Reader(data))
    );
  }
  // Truncated ...
}
```

And you end up with typescript errors like this:
Class 'MathServiceClientImpl' incorrectly implements interface 'MathService'.

protango:
Fixes #310.

The lowerCaseServiceMethods option was inconsistently applied, for example, it was being applied for nestjs but not for `grpc-web`. It was also not applied to auto-generated batching methods, leading to inconsistent output and typescript compilation errors.

This PR aims to address these issues by replacing all `MethodDescriptorProto` objects with a new class `FormattedMethodDescriptor`. This class implements `MethodDescriptorProto` and copies all properties verbatim, so existing usage is not impacted in any way. However, an additional field, `formattedName`, is added which applies any formatting rules in the `Options` to the original method name. I expect that this pattern could be used in other areas where we want to conditionally format strings and other values.

In order to not refactor the whole application to expect a `FormattedMethodDescriptor` in place of a `MethodDescriptorProto`, areas which require access to the `formattedName` property simply `assert` that the methodDescriptor is an instance of `FormattedMethodDescriptor`, which should always succeed because they are all converted in `main`. Code that doesn't use this new property does not need to know about the new class at all, and works as normal.

## Testing

Integration tests are setup to validate that the option produces a lower case implementation, but preserves the original name internally. It also validates the behavior works as expected for batch functions.


src/utils.ts
```diff
@@ -97,3 +99,71 @@ export function maybePrefixPackage(fileDesc: FileDescriptorProto, rest: string):
   const prefix = fileDesc.package === '' ? '' : `${fileDesc.package}.`;
   return `${prefix}${rest}`;
 }
+
+/**
+ * Asserts that an object is an instance of a certain class
+ * @param obj The object to check
+ * @param constructor The constructor of the class to check
+ */
+export function assertInstanceOf<T>(obj: unknown, constructor: { new (...args: any[]): T }): asserts obj is T {
```
stephenh, line 108: This is a pretty neat approach!

stephenh (approved):

Issue #333, opened 2021-07-09
Issues parsing options


andrewparmet:
I'm trying to pass options to the plugin and getting they don't seem to be getting picked up by the generator. I'm running the plugin through Gradle with the protobuf-gradle-plugin, version 1.82.0.

I took a very hacky approach to debugging and replaced the code generation line in `build/plugin.js` with:
```javascript
return { name: path, content: utils_1.prefixDisableLinter(spec) + request.parameter + "\n" + JSON.stringify(options) };
```

And I see in my files:
```
--ts_proto_opt=oneof=unions,--ts_proto_opt=forceLong=long,--ts_proto_opt=esModuleInterop=true
// (below is prettified)
{
  "context": false,
  "snakeToCamel": true,
  "forceLong": "number",
  "useOptionals": false,
  "useDate": "date",
  "oneof": "properties",
  "esModuleInterop": false,
  "lowerCaseServiceMethods": false,
  "outputEncodeMethods": true,
  "outputJsonMethods": true,
  "outputPartialMethods": true,
  "outputTypeRegistry": false,
  "stringEnums": false,
  "constEnums": false,
  "outputClientImpl": true,
  "outputServices": false,
  "returnObservable": false,
  "addGrpcMetadata": false,
  "addNestjsRestParameter": false,
  "nestJs": false,
  "env": "both",
  "unrecognizedEnum": true,
  "exportCommonSymbols": true,
  "outputSchema": false,
  "onlyTypes": false,
  "--ts_proto_opt": "esModuleInterop"
}
```

So I see my options are making it to the plugin, but the parsing seems to choke? What I don't understand is that when I run the plugin manually the options seem to be recognized just fine.

Clearly `forceLong` and `oneof` were not picked up, and `esModuleInterop` shows up unparsed at the end. Am I providing options incorrectly somehow?

andrewparmet:
When I run the hacked plugin manually via protoc, I get `oneof=unions,esModuleInterop=true,forceLong=long` in the first line, which seems to be how the options parser intends to receive them.

andrewparmet:
Ah, I was passing `ts_proto_opt=foo=bar` in the `option()` method on the Gradle plugin. I should have been passing just `foo=bar`.

stephenh:
[...]

PR #334, opened 2021-07-11
docs: Add Gradle invocation instructions

andrewparmet:
See #333.

Provides basic instructions for integration ts-proto with Gradle and the protobuf-gradle-plugin. The provided snippet does not address installation of ts-proto or usage of the generated code. My own use case involves automated publishing through Gradle, but one could presumably consume the code locally by pointing `tsconfig.json` to the output directory.

Feel free to modify the location within the readme and the wording.

Thanks for a great tool!


stephenh:
Great, thanks @andrewparmet !

Issue #329, opened 2021-06-27
Observable<?>.fromPartial appearing in source code, no known type


doctorpangloss:
I don't know why it's being emitted regardless of how I configure ts-proto.

Source example:

```
service A {
  rpc B(stream google.proto.Empty) returns (stream google.proto.Empty) {}
}
```

will generate

```
B(
      request: DeepPartial<Observable<Empty>>,
      metadata?: grpc.Metadata,
    ): Observable<HostReply> {
      return this.rpc.invoke(
        RemoteBDesc,
        Observable<Empty>.fromPartial(request),
        metadata,
      );
    }
```

version: 1.81.3
options:

```
        option "outputClientImpl=grpc-web"
        option "esModuleInterop=true"
        option "returnObservable=false"
        option "addGrpcMetadata=true"
        option "forceLong=number"
        option "oneof=unions"
        option "outputJsonMethods=false"
        option "lowerCaseServiceMethods=true"
```

stephenh:
Yeah, hm, it looks like this line:

https://github.com/stephenh/ts-proto/blob/main/src/generate-grpc-web.ts#L65

Was written before/without realizing that this line:

https://github.com/stephenh/ts-proto/blob/main/src/types.ts#L540

Might return `Observable`, because streaming support was added later, and for a specific use case instead across the board for all combinations of options (as you're finding).

If you could reproduce in one of the `integration/...` directories that covers the grpc-web output, and submit a PR/fix, that'd be great. Thanks!

samelie:
Bump :]

stephenh:
Updating my prior diagnosis, this doesn't come from "we shouldn't use `Observable` here, it's that methods that are `clientStreaming` basically have to use `Observable` (to present the client is sending 0 or N messages, and so we need `Observable` to tell us when the last `N`-th message has been written by the client) and ts-proto doesn't have `Observable` / grpc-web wiring built yet.

snyh:
The PR #617 is seems trying to support client stream for grpc-web

luhuaei:
Try to support (client/server) streaming not `Observable` type. #633

PR #335, opened 2021-07-14
fixing grpc-web issue#239 where client streaming was producing incorr…

samelie:
fixing grpc-web issue#239 where client streaming was producing incorect requestType

Based on this issue https://github.com/stephenh/ts-proto/issues/329

Please note there something wrong with `integration/meta-typings/meta-typings-test.ts` & I'm really not sure why.
All other tests pass, including the new `grpc-web-client-streaming` used to demonstrate the issue with grpc-web + clientStreaming

Hopefully this fix isn't overly naive

Thanks for all the hardwork putting this together; it's really cool


src/types.ts
```diff
@@ -539,7 +539,8 @@ export function detectMapType(
 
 export function requestType(ctx: Context, methodDesc: MethodDescriptorProto): Code {
   let typeName = messageToTypeName(ctx, methodDesc.inputType);
-  if (methodDesc.clientStreaming) {
+  const { options } = ctx;
+  if (options.nestJs && methodDesc.clientStreaming) {
```
stephenh, line 543: I think this fixes the compile error, but in a pretty indirect way, right? Like the `*.proto` file is saying "the client can submit a stream of messages" and this is basically saying "we won't do that"...which, compiles, but does it actually work?

Like looking at the grpc-web examples, if a method is `stream Cred` we can't just write a `Cred` and call it good, we need to tell the server "we _could_ have sent more `Cred`s but we're not", i.e. by calling `finishSend`:

https://github.com/improbable-eng/grpc-web/blob/master/client/grpc-web/src/client.ts#L304

So this is less of "just fix the compile error" and more that I think we need to actually support client-to-server / request streaming. Until then, ts-proto just won't work with `*.proto` files that use `stream Cred`.

Maybe we could like skip putting those methods in the output until it's supported, so that if you have a `*.proto` file that has those, but you don't need to call them, you'd at least get output that would let you use the other methods.

stephenh:
Fwiw @samelie I pushed a "re-codegen" commit to this PR that undoes a lot of `*.bin` file changes you'd had in here that come from having different versions of `protoc`. Ideally you shouldn't need to include those changes in the PR.

src/types.ts
samelie: Yes, you are right, this does just fix the compile error. I don't have rpc services with a backend setup right now test; though; from what [improbable says](https://github.com/improbable-eng/grpc-web#client-side-streaming) , it's not supported. 

Omitting entirely the `stream Cred` methods sounds like a good plan; stay tuned

samelie:
Hey @stephenh  hopefully this approach is better.  the goal of not generating  `rpc ChangeUserSettingsStream (stream Cred) returns (stream DashUserSettingsState) {}` has been accomplished for grpc-web

boukeversteegh:
Hi @samelie Is this PR still under development?

samelie:
@stephenh  @boukeversteegh 

I'm really not familiar with client streaming in grpc-web. I saw a post on their issues about it saying you needed the websocket transport. Makes sense.
One day I'll do an experiment (we like to say) , but I've been holding out for streamsAPI https://github.com/improbable-eng/grpc-web#client-side-streaming
All that to say I don't have an easy way to setup a websockets and write the rxjs push code we'd need.

For me, this wouldn't be an issue (am fine with it not working), however I'm in a work situation where we need to run `tsc` on the generated output that produces errors like these for clientStreaming

```
 '(' expected.

         Observable<WriteVerifyFileRequest>.fromPartial(request),
```

`tsc` is really annoying in the way there no way to silence errors. For me since this is all in CI, this is a problem as I can't swallow them either (bash noob?) 

The easiest solution in the vein of _here be dragons_ is just to treat clientSide **in the case of grpc-web && !nestJs** like a normal Promise rpc

I'd be happy to propose my forked fix for this that changes `requestType` in `types/types.ts` to 

```
export function requestType(ctx: Context, methodDesc: MethodDescriptorProto): Code {
  let typeName = rawRequestType(ctx, methodDesc);
  if (methodDesc.clientStreaming && ctx.options.nestJs) {
    return code`${imp('Observable@rxjs')}<${typeName}>`;
  }
  return typeName;
}
```

Less magical solutions would be better. Perhaps some ctx option to ignore all clientStreaming?

PR #339, opened 2021-08-03
fix: resolve import collisions for enums

doochik:


stephenh:
Nice, thanks @doochik !

PR #344, opened 2021-08-05
fix: Field starting with '_' generates an interface property starting with 'undefined'

### LINKED ISSUE #343: Field starting with '_' generates an interface property starting with 'undefined'
Firstly, thank you for this great library!

Version: `1.82.4`

Protobuffer:
```
message Test {
  optional string _uuid = 1;
  required string last_name = 2;
}
```
Expected result with default command line options:
```
export interface Test {
  _uuid: string;
  lastName: string;
}
```

Result with default command line options:
```
export interface Test {
  undefinedUuid: string;
  lastName: string;
}
```

Result with `--ts_proto_opt=snakeToCamel=false`:
```
export interface Test {
  _uuid: string;
  last_name: string;
}
```

doochik:
Fix #343

### DIFF
--- a/src/case.ts
+++ b/src/case.ts
@@ -6,7 +6,8 @@ export function maybeSnakeToCamel(s: string, options: Pick<Options, 'snakeToCame
       .split('_')
       .map((word, i) => {
         if (i === 0) {
-          return word[0] + word.substring(1).toLowerCase();
+          // if first symbol is "_" then skip it
+          return word ? word[0] + word.substring(1).toLowerCase() : '';
         } else {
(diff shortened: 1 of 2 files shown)


doochik:
@stephenh I can't find any formal rules for snakeToCamel (RFC? ISO?). So I just replicate `protoc` behaviour

stephenh:
@doochik oh yeah...I'd just made that up :-) not sure if there are some officials rules somewhere. Good call on just replicating the protoc behavior behavior!

stephenh (approved):

PR #345, opened 2021-08-12
fix: generated code for browser gives error "Cannot find name global"

marcos10soares:
When using the *.ts generated code for browser I get the error:
```js
error  in .../dashboard/node_modules/my_package/my_proto/my_proto.ts

ERROR in .../dashboard/node_modules/my_package/my_proto/my_proto.ts(1128,14):
1128:14 Cannot find name 'global'.
    1126 |   if (typeof self !== "undefined") return self;
    1127 |   if (typeof window !== "undefined") return window;
  > 1128 |   if (typeof global !== 'undefined') return global;
         |              ^
    1129 |   throw "Unable to locate global object";
    1130 | })();
    1131 |
```

adding this line solves the issue:
```
declare var global: any | undefined;
```

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -40,6 +40,7 @@ export const Message = {
 
 declare var self: any | undefined;
 declare var window: any | undefined;
+declare var global: any | undefined;
 var globalThis: any = (() => {
   if (typeof globalThis !== 'undefined') return globalThis;
   if (typeof self !== 'undefined') return self;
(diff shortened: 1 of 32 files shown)


stephenh (approved):
Huh; I kinda feel like there should be a better way to do this, like with a `types` hint in downstream project's `tsconfig`...but I'm not really sure, and we've already got lines here for `self` and `window` ... so ... :shrug: sounds good to me. :-) Thanks!

PR #347, opened 2021-08-13
feat: Make esModuleInterop generate fully valid ESM code

### LINKED ISSUE #338: ESM fails with protos which require other protos
I'm using this library with [lnd](https://githu.com/lightningnetwork/lnd)'s rpc.proto, stateservice.proto and walletunlocker.proto and the `esModuleInterop` flag. 

Because walletunlocker.proto uses `ChanBackupSnapshot` from rpc.proto, it adds the following line
```
import { ChanBackupSnapshot } from "./rpc";
```
For a generated file to be true ESM and work however, it needs to be

```
import { ChanBackupSnapshot } from "./rpc.js";
```

Because TypeScript can still resolve the import with .js, but node needs the full path with extension in the output.

The same applies for

Current:

```
import _m0 from "protobufjs/minimal";
```

Working:

```
import _m0 from "protobufjs/minimal.js";
```

AaronDewes:
Fixes #338.

This is still WIP.

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import { util, configure, Writer, Reader } from 'protobufjs/minimal';
+import { util, configure, Writer, Reader } from 'protobufjs/minimal.js';
 import * as Long from 'long';
 
 export const protobufPackage = 'angular';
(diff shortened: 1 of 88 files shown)


stephenh:
@AaronDewes hey! Just curious, did you run into some specific challenges with the PR? Just wondering for the next person that tackles this problem. Thanks!

AaronDewes:
> @AaronDewes hey! Just curious, did you run into some specific challenges with the PR? Just wondering for the next person that tackles this problem. Thanks!

I'll publish my recent local changes, for some reason, when I tried to only change the imp function (overwrite it with a custom one which appends .js to paths), files ended up importing themselves. I think I just found the issue though and will have a look again as soon as possible.

wtachau:
👋 @AaronDewes thanks for starting this! If you are able to share your notes about why this was happening, I'd be happy to take a look as well

paralin:
Did you ever figure this out? I'm having the issue still that the import has to be 'protobufjs/minimal.js'

Issue #349, opened 2021-08-16
Allow disabling of service definition generation


emil-e:
We're generating service definitions using our own custom generator and would like to disable the generation that ts-proto does. I could implement this myself and provide a PR but would like to know what is the preferred way of implementing the option itself. Would setting `--ts_proto_opt=outputServices=false` be appropriate?

stephenh:
@emil-e sure, `outputServices=false` sounds like a good idea; a PR would be great. Thanks!

PR #357, opened 2021-09-06
feat: Service generation option

Demiguise:
Fixes https://github.com/stephenh/ts-proto/issues/349

Added some minor refactoring to extract out the service option to its own Enum.


stephenh:
Cool, thanks for the PR! Fwiw any other work / clean up you want to do around the service code would be great.

stephenh (approved):

Issue #372, opened 2021-10-25
Streaming support


Jille:
Streaming support currently seems broken. It generates a file with parse errors for me at least. It looks like there is partial support, going in the direction of using rxjs Observables.

Would it be valuable to agree on the interface of where we(/you) want to go?

Is there a plan for cancellation of streaming RPCs?

```diff
@@ -2,6 +2,7 @@
 import Long from 'long';
 import _m0 from 'protobufjs/minimal';
 import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
 import { Timestamp } from './google/protobuf/timestamp';
 
 export const protobufPackage = '';
@@ @@
   
     MyServerStreaming(
       request: MyServerStreamingRequest
-    ): Promise<MyServerStreamingResponse> {
+    ): Observable<MyServerStreamingResponse> {
       const data = MyServerStreamingRequest.encode(request).finish();
-      const promise = this.rpc.request(
+      const stream = this.rpc.serverStreaming(
         
         "MyService",
         "MyServerStreaming",
         data
       );
-      return promise.then(data => MyServerStreamingResponse.decode(new _m0.Reader(data)));
+      return stream.pipe(map(packet => MyServerStreamingResponse.decode(new _m0.Reader(packet))));
     }
   
     MyBiDi(
       request: Observable<MyBiDiRequest>
-    ): Promise<MyBiDiResponse> {
-      const data = Observable<MyBiDiRequest>.encode(request).finish();
-      const promise = this.rpc.request(
+    ): Observable<MyBiDiResponse> {
+      const data = request.pipe(map(packet => MyBiDiRequest.encode(packet).finish()));
+      const stream = this.rpc.bidirectionalStreaming(
         
         "MyxService",
         "MyBiDi",
         data
       );
-      return promise.then(data => MyBiDiResponse.decode(new _m0.Reader(data)));
+      return stream.pipe(map(packet => MyBiDiResponse.decode(new _m0.Reader(packet))));
     }
   
     GetProduct(
@@  @@
 interface Rpc {
       request(
         
         service: string,
         method: string,
         data: Uint8Array
       ): Promise<Uint8Array>;
+
+      serverStreaming(
+        
+        service: string,
+        method: string,
+        data: Uint8Array
+      ): Observable<Uint8Array>;
+
+      clientStreaming(
+        
+        service: string,
+        method: string,
+        data: Observable<Uint8Array>
+      ): Uint8Array;
+
[...]

PR #373, opened 2021-10-25
feat: Streaming support

Jille:
I just went ahead and implemented my proposal from #372.

This should make it possible to do streaming RPCs, given that the user's Rpc implementation implements a few extra methods.

Not sure why the Reader/Writer order was changed in the generated files. I'm assuming it's because I changed some calling order, but I didn't investigate.

### DIFF
--- a/integration/batching-with-context/batching.ts
+++ b/integration/batching-with-context/batching.ts
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import { util, configure, Reader, Writer } from 'protobufjs/minimal';
+import { util, configure, Writer, Reader } from 'protobufjs/minimal';
 import * as Long from 'long';
 import * as DataLoader from 'dataloader';
 import * as hash from 'object-hash';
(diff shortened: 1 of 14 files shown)


stephenh (approved):
Hey @Jille ! Thanks for the PR! I haven't personally done a lot of streaming, so contributions to flush this out is great. Looks like maybe its just prettier failing the CI checks, but otherwise lgtm!

Jille:
Thanks @stephenh!

There seems to be one other weird thing from the CI: https://github.com/stephenh/ts-proto/runs/4037980567?check_suite_focus=true - It rewrites integration/grpc-web-go-server/example.ts and integration/no-proto-package/no-proto-package.ts and eats all the indentation. I don't see how that'd be caused by my PR, is it a known issue?

stephenh:
@Jille ah yeah, those errors can be hard to debug; there is a likely-silly syntax error, which means that the last "run prettier to get nice indentation" step can't run, and so you're seeing the raw ts-proto output. Maybe try running `./codegen.sh grpc-web-go-server` locally and then opening the file in vscode/etc. to see if it will, ugly formatting aside, point you in the right direction of the syntax error.

Jille:
Thanks, that was the hint I needed :)

stephenh:
Looks good @Jille ! Going to hit merge and should be auto-released.

PR #374, opened 2021-10-25
chore: Harden ts config and do some housekeeping 🧹

webmaster128:


webmaster128:
I think this would be better and easier achieved by setting up eslint. Then we can keep everything in one project (or at least seperate in a second step) and globally [allowing `_` prefixed unused variables](https://github.com/cosmos/cosmjs/blob/v0.26.3/.eslintrc.base.js#L71). Would you be fine with addting eslint?

stephenh:
@webmaster128 sure!

webmaster128:
Closing in favour of lint solution

PR #376, opened 2021-10-26
feat: Reduce code size by using nullish coalescing operator in fromPartial

### LINKED ISSUE #371: Use nullish coalescing operator (??) for assigning binaries
We generate a lot of code like this:

```ts
  fromPartial(object: DeepPartial<ExistenceProof>): ExistenceProof {
    const message = { ...baseExistenceProof } as ExistenceProof;
    message.path = [];
    if (object.key !== undefined && object.key !== null) {
      message.key = object.key;
    } else {
      message.key = new Uint8Array();
    }
    if (object.value !== undefined && object.value !== null) {
      message.value = object.value;
    } else {
      message.value = new Uint8Array();
    }
    if (object.leaf !== undefined && object.leaf !== null) {
      message.leaf = LeafOp.fromPartial(object.leaf);
    } else {
      message.leaf = undefined;
    }
    if (object.path !== undefined && object.path !== null) {
      for (const e of object.path) {
        message.path.push(InnerOp.fromPartial(e));
      }
    }
    return message;
  },
```

in which the binary fields can be replaced with the [nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) as follows:

```diff
@@ -429,16 +429,8 @@ export const ExistenceProof = {
   fromPartial(object: DeepPartial<ExistenceProof>): ExistenceProof {
     const message = { ...baseExistenceProof } as ExistenceProof;
     message.path = [];
-    if (object.key !== undefined && object.key !== null) {
-      message.key = object.key;
-    } else {
-      message.key = new Uint8Array();
-    }
-    if (object.value !== undefined && object.value !== null) {
-      message.value = object.value;
-    } else {
-      message.value = new Uint8Array();
-    }
+    message.key = object.key ?? new Uint8Array();
+    message.value = object.value ?? new Uint8Array();
     if (object.leaf !== undefined && object.leaf !== null) {
       message.leaf = LeafOp.fromPartial(object.leaf);
     } else {
```

This output is easier to read and saves a good amount of code size that is otherwise shipped to the user. 

[Nullish Coalescing is available since TypeScript 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) and will be transpiled to a compact replacement. [For targets >= es2020 it will be preserved](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#es2020-for-target-and-module) as valid JavaScript syntax.  See also [JavaScript operator: Nullish coalescing operator  on caniuse.com](https://caniuse.com/mdn-javascript_operators_nullish_coalescing).

In order to make such a change, a bit of refactoring is needed. Instead of have [a separate block for the default initializations](https://github.com/stephenh/ts-proto/blob/v1.83.1/src/main.ts#L1225-L1229), each field should be fully initialized in one step, either with data or with default. But such a refactoring would be nice anyways as it makes the flow easier and future optimizations possible, like this one:

```diff
@@ -439,11 +439,9 @@ export const ExistenceProof = {
     } else {
       message.value = new Uint8Array();
     }
-    if (object.leaf !== undefined && object.leaf !== null) {
-      message.leaf = LeafOp.fromPartial(object.leaf);
-    } else {
-      message.leaf = undefined;
-    }
+    message.leaf = (object.leaf !== undefined && object.leaf !== null)
+      ? LeafOp.fromPartial(object.leaf)
+      : undefined;
     if (object.path !== undefined && object.path !== null) {
       for (const e of object.path) {
         message.path.push(InnerOp.fromPartial(e));
```

I'm happy to work on this if there is general interest on the approach.

webmaster128:
Closes #371

~Builds on #375 (merge that first, then rebase)~

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -54,9 +54,7 @@ export const SimpleMessage = {
 
   fromPartial(object: DeepPartial<SimpleMessage>): SimpleMessage {
     const message = { ...baseSimpleMessage } as SimpleMessage;
-    {
-      message.numberField = object.numberField ?? 0;
-    }
+    message.numberField = object.numberField ?? 0;
     return message;
   },
 };
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -54,10 +54,8 @@ export const SimpleMessage = {
 
   fromPartial(object: DeepPartial<SimpleMessage>): SimpleMessage {
     const message = { ...baseSimpleMessage } as SimpleMessage;
-    if (object.numberField !== undefined && object.numberField !== null) {
-      message.numberField = object.numberField;
-    } else {
-      message.numberField = 0;
+    {
+      message.numberField = object.numberField ?? 0;
     }
     return message;
   },
--- a/integration/avoid-import-conflicts/simple.ts
+++ b/integration/avoid-import-conflicts/simple.ts
@@ -117,9 +117,7 @@ export const Simple = {
 
   fromPartial(object: DeepPartial<Simple>): Simple {
     const message = { ...baseSimple } as Simple;
-    {
-      message.name = object.name ?? '';
-    }
+    message.name = object.name ?? '';
     if (object.otherSimple !== undefined && object.otherSimple !== null) {
       message.otherSimple = Simple2.fromPartial(object.otherSimple);
     } else {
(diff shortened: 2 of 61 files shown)


webmaster128:
Alright, I'm happy with this now. In https://github.com/confio/cosmjs-types/pull/11/files you see what this change means for the code generated larger projects.

Rebased onto current main branch.

How do you want this to be released? Is this a feature or a fix?

stephenh (approved):
@webmaster128 looks great! Much cleaner! Yeah, releasing as a feature sgtm.

webmaster128:
Thanks :)

PR #385, opened 2021-11-11
docs: Add note about transport

enyo:
I've just spent about an hour trying to figure out why it wasn't working because I didn't realise that this just applies to a node environment. (Now that I know, it's quite obvious).

Since this is the reference for `grpc-web` I assume that many people will just copy this and a comment like this might be helpful.


stephenh:
Cool, thanks for the clarification!

PR #387, opened 2021-11-14
feat: Initialize lists with map

webmaster128:
This uses `map` to initialize lists, such that all cases can be handled within one assignment. `if (object.ids !== undefined && object.ids !== null) {` was replaced with `object.ids ?? []` to save nesting depth and code size.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -983,25 +983,24 @@ function generateFromJson(ctx: Context, fullName: string, messageDesc: Descripto
 
     // and then use the snippet to handle repeated fields if necessary
     if (isRepeated(field)) {
-      const value = isMapType(ctx, messageDesc, field) ? '{}' : '[]';
-      const name = maybeSnakeToCamel(field.name, options);
-      chunks.push(code`message.${name} = ${value};`);
-      chunks.push(code`if (object.${fieldName} !== undefined && object.${fieldName} !== null) {`);
       if (isMapType(ctx, messageDesc, field)) {
+        chunks.push(code`message.${fieldName} = {};`);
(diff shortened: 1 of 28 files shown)


webmaster128:
Any idea why CI is failing here? The problematic "Example 6" was added here in #370. The other integration/*/google/protobuf/timestamp.ts alle have the older version of the comment. Why did this not fail in the PR #370 or on main but now?

stephenh (approved):

stephenh:
@webmaster128 yeah, I'm not really sure how that `timestamp.ts` change got through that other PR... like it was only a month or so ago, and I haven't changed any of the lint/CI setup in quite awhile.

I merged #388 so am personally pretty willing to procrastinate on figuring out true root cause + better yet moving to a better `protoc` setup. :-) But if that's something you'd like to improve, that'd be great, we definitely need something more deterministic than the current setup.

This PR also looks great! Thanks!

src/main.ts
```diff
@@ -1249,7 +1225,7 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
       chunks.push(code`message.${fieldName} = ${readSnippet(`object.${fieldName}`)};`);
       chunks.push(code`} else {`);
       const fallback = isWithinOneOf(field) ? 'undefined' : defaultValue(ctx, field);
-      chunks.push(code`message.${fieldName} = ${fallback}`);
+      chunks.push(code`message.${fieldName} = ${fallback};`);
```
webmaster128, line 1228: This change is unrelated and simply here to increase consistency between `generateFromPartial` and `generateFromJson`.

webmaster128:
Rebased. When it passes CI, I'm happy.

PR #394, opened 2021-11-16
feat: Use ternary operator for conditional assignments

webmaster128:
That's the second step initially proposed in https://github.com/stephenh/ts-proto/issues/371. In cases where the input value needs to be transformed, we [cannot use nullish coalescing](https://stackoverflow.com/questions/69696665/optional-chaining-with-free-functins) or anything new and fancy.

This improves the code output in two ways:
- Less code (that ultimatly reduces the app bundle size for the end user)
- Exactly one assignment per field which I find easier to reason about

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1212,11 +1212,7 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
       const fallback = isWithinOneOf(field) ? 'undefined' : defaultValue(ctx, field);
       chunks.push(code`message.${fieldName} = ${fallback}`);
       chunks.push(code`}`);
-    } else if (
-      isPrimitive(field) ||
-      (isTimestamp(field) && (options.useDate === DateOption.DATE || options.useDate === DateOption.STRING)) ||
-      isValueType(ctx, field)
-    ) {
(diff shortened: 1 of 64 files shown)


src/main.ts
```diff
@@ -1212,21 +1212,17 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
       const fallback = isWithinOneOf(field) ? 'undefined' : defaultValue(ctx, field);
       chunks.push(code`message.${fieldName} = ${fallback}`);
       chunks.push(code`}`);
-    } else if (
-      isPrimitive(field) ||
-      (isTimestamp(field) && (options.useDate === DateOption.DATE || options.useDate === DateOption.STRING)) ||
-      isValueType(ctx, field)
-    ) {
+    } else if (readSnippet(`x`).toCodeString() == 'x') {
```
webmaster128, line 1215: This somehwat unrelated change is worth having a 👀.

stephenh: Ah yeah, that's a neat way of doing this.

stephenh (approved):
Looks great!

webmaster128:
Thanks!

PR #395, opened 2021-11-19
fix: code generation for int64 map values in fromPartial and fromJson

SchroederChris:
Hi @stephenh,

I noticed that with `--ts_proto_opt=forceLong=long`, if you have a map with an int64 value (e.g. `map<string, int64> data`), the generated typescript code in `fromJSON` and `fromPartial` wasn't correct (it fell back to `message.data[key] = Long(value)`, which didn't compile). I changed that to use `Long.fromString` in `fromJSON` (int64 should always be string in JSON, right?) and `Long.fromValue` in `fromPartial`.

As we don't use binary protobuf, I only checked these two methods.

Also, I unfortunately didn't figure out how to add a test for this, so I'll leave this to you. 

Kind regards!


stephenh:
@SchroederChris thanks for the PR; I've added some tests and a few more fixes. Going to merge, thanks!

SchroederChris:
Thanks for the quick merge!

While implementing the changes in our codebase, I realized that it might be a good idea to also use `Long.fromValue()` instead of `Long.fromString()` in `fromJSON`. This would make the code more robust if called with an already parsed object. What do you think?

stephenh:
@SchroederChris ah sure! That sgtm. Can you submit a PR with that change? Thanks!

SchroederChris:
👍 here it is: https://github.com/stephenh/ts-proto/pull/405

PR #401, opened 2021-11-21
refactor: Switch to @grpc/grpc-js altogether. Bump to nestjs v8.

sgielen:
The npm package grpc [is deprecated](https://www.npmjs.com/package/grpc) and the authors recommend to use @grpc/grpc-js instead. While ts-proto uses the new version almost everywhere, it still needed grpc for the nestjs integration tests, as nestjs v7 used the old grpc.

However, nestjs v8 uses @grpc/grpc-js as well, so switch to that instead.

I needed this, because grpc [does not seem to compile on M1 Macs anymore](https://github.com/grpc/grpc-node/issues/1880) (on this issue,[ one of the authors recommended](https://github.com/grpc/grpc-node/issues/1880#issuecomment-897147496) moving to @grpc/grpc-js as a solution).

Additional changes:
- Nestjs v8 requires requires rxjs v7+, so bump to it, too.
- `listenAsync` is deprecated and should be replaced by `listen`.
- Change port number for the nestjs-simple-restparameters test to 8081, so it doesn't conflict with the nestjs-simple test which also used 8080. I think this is the cause that the test occasionally failed for me when the two were run together, but never when they were run separately.


stephenh (approved):
Nice, thanks!

sgielen:
@stephenh I had to add a commit to fix the tests on node 12, could you re-approve? :)

stephenh:
@sgielen yep, lgtm, thanks!

PR #402, opened 2021-11-21
feat: Add useOptionals=all to enable non-field members to be optional.

sgielen:
This PR also adds a test to verify writing empty objects and reading them back.


src/main.ts
```diff
@@ -904,14 +914,18 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
             `
           : writeSnippet(`{ ${maybeTypeField} key: key as any, value }`);
         chunks.push(code`
-          Object.entries(message.${fieldName}).forEach(([key, value]) => {
-            ${entryWriteSnippet}
-          });
+          if (message.${fieldName} !== undefined && Object.keys(message.${fieldName}).length !== 0) {
+            Object.entries(message.${fieldName}).forEach(([key, value]) => {
```
stephenh, line 918: I would if we could make this `Object.entries(message.foo || []).forEach`. I know it's slightly less efficient, in terms of making a throw away `[]` and `lambda` if the field is not set, but I'm thinking that @webmaster128 just did several PRs that a lot of explicit `message.field !== undefined && ...` checks to reduce the code size ouptut.

So, I suppose it's a trade-off, do you prefer less code w/a few extra allocations, or more code w/a few less allocations?

I suppose in an ideal world, ts-proto would let the user pick, but we already have a ton of flags, so dunno, I think I'd prefer to just pick one.

@sgielen wdyt about `|| []`?

src/main.ts
```diff
@@ -922,21 +936,25 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
         const tag = ((field.number << 3) | 2) >>> 0;
         const toNumber = getEnumMethod(typeMap, field.typeName, 'ToNumber');
         chunks.push(code`
-          writer.uint32(${tag}).fork();
-          for (const v of message.${fieldName}) {
-            writer.${toReaderCall(field)}(${toNumber}(v));
+          if (message.${fieldName} !== undefined && message.${fieldName}.length !== 0) {
```
stephenh, line 939: Unlike the other spot where I suggested `|| []`, it does seem nice here to keep the tag completely off the wire if there are no elements in the collection.

src/main.ts
```diff
@@ -586,8 +587,17 @@ function makeTimestampMethods(options: Options, longs: ReturnType<typeof makeLon
...
-  return (options.useOptionals && isMessage(field) && !isRepeated(field)) || field.proto3Optional;
+// When forceOptionals=true, all fields are translated into optional properties, except map Entry key/values.
+function isOptionalProperty(
+  field: FieldDescriptorProto,
+  messageOptions: MessageOptions | undefined,
+  options: Options
+): boolean {
+  return (
+    (options.useOptionals && isMessage(field) && !isRepeated(field)) ||
+    (options.forceOptionals && !messageOptions?.mapEntry) ||
```
stephenh, line 598: Per discussion, I'm good with removing `options.forceOptionals` and just remove the `isMessage` / `!isRepeated` from the existing check.

Also I assume `!...mapEntry` is in here just b/c there are some nuances to making it optional that you didn't want to tackle in this PR? If so, maybe just add an in-source comment/todo that it's for later.

src/types.ts
```diff
@@ -251,28 +251,28 @@ export function notDefaultCheck(ctx: Context, field: FieldDescriptorProto, place
       const zerothValue = enumProto.value.find((v) => v.number === 0) || enumProto.value[0];
       if (options.stringEnums) {
         const enumType = messageToTypeName(ctx, field.typeName);
-        return code`${place} !== ${enumType}.${zerothValue.name}`;
+        return code`${place} !== undefined && ${place} !== ${enumType}.${zerothValue.name}`;
```
stephenh, line 254: Hm, maybe we could do:

```
const maybeNotUndefined = options.useOptionals ? `${place} !== undefined && ` : "";
```

At the start of this method, and then in these places do:

```
return code`${maybeNotUndefined} ${place} !== ${enumType}.${zerothValue.name}`;
```

So that users that don't have `useOptional`s set won't have the unnecessary `!== undefined` in their output.

Granted, it doesn't hurt, but I think I'd like to keep it out if possible.

stephenh (commented):
Looks great! A few questions but otherwise lgtm direction wise.

src/main.ts
webmaster128: I'd not worry about performance too much. JIT engines run this stuff amazingly fast these days. The better the optimizer understands the intend of the code, the better.

I'd use `Object.entries(message.foo ?? {})` to
1. Only use the fallback in case of `null` and `undefined` and not bugs like `""`, `0` or `false`,
2. Always have an onject as input to `Object.entries(`

webmaster128: But this extra undefined check is only needed in case of `forceOptionals = true`? So maybe the code output should only be generated in that case. Same in the diff nelow.

src/main.ts
```diff
@@ -1309,7 +1327,7 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
         chunks.push(code`
           Object.entries(object.${fieldName}).forEach(([key, value]) => {
             if (value !== undefined) {
-              message.${fieldName}[${i}] = ${readSnippet('value')};
+              message.${fieldName}![${i}] = ${readSnippet('value')};
```
webmaster128, line 1330: This exclamation mark can be avoided if the code is rewritten using `reduce`. Then there is only a single assignment to `message.${fieldName}`. I did the same with `map` already for lists.

Clearly out of scope for this PR but I'll have a look.

webmaster128: https://github.com/stephenh/ts-proto/pull/410

webmaster128: Okay, if you rebase on latest main, you don't need this diff here anymore because there is a single assignment of `T` to `message.fiedlName: T?` now.

src/main.ts
sgielen: I can switch to `|| []` and `|| {}` no problem.

I'll respond to the "extra undefined check" comment in another thread on this PR. :)

src/main.ts
sgielen: Yes, otherwise I had to assert that the Map entry key is set in various places, which seems silly as the Map entry key is mandatory.

src/main.ts
sgielen: Although - if we switch to `|| []`, we do put empty lists on the wire, while if we keep `if (.. !== undefined)`, we don't put the list on the wire. I have a slight preference not to put empty lists on the wire. What do you both think?

src/main.ts
sgielen: Perfect, will do! Thanks!

src/types.ts
sgielen: I actually had problems with this before implementing `force/useOptionals`. Suppose that you have a message FooRequest with a `repeated int64 list = 0;` (remember you can't mark repeateds as optional), and an RPC `SendFoo(FooRequest)`. This code, intended to send an empty list (as with Go semantics), would trigger an error:

```
SendFoo({})
```

The error triggered is quite deep inside ts-proto and reads something like "list is not iterable", regardless of the value of force/useOptionals. IMHO, it is a small price to pay to always do the undefined check to allow this code to work normally.

If you strongly prefer to only enable this behaviour with force/useOptionals enabled then I can apply your suggestion, but personally I would strongly prefer not to.

src/main.ts
sgielen: Derp, nevermind -- an empty map doesn't actually cause any writer calls, so there's no on-wire difference between an empty map and not writing the map at all. So, this is good to go.

stephenh:
> Rerun codegen + add the diff to this PR before merging.

Cool, yeah, I think now is a good time to do that.

> I've kept the flag forceOptionals since it looks like that's where consensus is heading.

I'd like to keep just the existing `useOptionals`, b/c we already have a large proliferation of flags. Instead of just `true`/`false`, about how doing `useOptionals=...`:

* `false` = current/default behavior
* `messages` = the current `useOptionals=true` behavior 
* `all` = the `forceOptional` behavior you're implementing here
* `true` = outputs a warning that `true` is deprecating, but keeps going using the `messages` behavior

I think this should let us achieve no breaking changes for current `useOptionals=true` users, but also set us up for `useOptionals=messages/all` going forward as I think nice/valid long-term options for users to choose which they want.

Wdyt/sgty?

src/types.ts
sgielen: OK, I'll revert to only add the undefined-check if `useOptionals` is set and make a separate issue for the `[]` being undefined in a partial object, so we can discuss it further there. :)

sgielen:
> > Rerun codegen + add the diff to this PR before merging.
> 
> Cool, yeah, I think now is a good time to do that.
> 
> > I've kept the flag forceOptionals since it looks like that's where consensus is heading.
> 
> I'd like to keep just the existing `useOptionals`, b/c we already have a large proliferation of flags. Instead of just `true`/`false`, about how doing `useOptionals=...`:
> 
> * `false` = current/default behavior
> * `messages` = the current `useOptionals=true` behavior
> * `all` = the `forceOptional` behavior you're implementing here
> * `true` = outputs a warning that `true` is deprecating, but keeps going using the `messages` behavior
> 
> I think this should let us achieve no breaking changes for current `useOptionals=true` users, but also set us up for `useOptionals=messages/all` going forward as I think nice/valid long-term options for users to choose which they want.
> 
> Wdyt/sgty?

Yes, sounds good to me! I'll make a tiny change and make `useOptionals: boolean | 'none' | 'messages' | 'all'` and make `none` the default so that in a future version, we can remove the boolean option altogether. If that's OK with you?

sgielen:
Both changes now implemented. This also means that rerunning codegen is not necessary: the codegen output is now equal if `useOptionals` is not set to `messages` or `all`.

I think this is ready for final review. :)

src/options.ts
```diff
@@ -32,7 +32,7 @@ export type Options = {
   context: boolean;
   snakeToCamel: boolean;
   forceLong: LongOption;
-  useOptionals: boolean;
+  useOptionals: boolean | 'none' | 'messages' | 'all'; // boolean is deprecated
```
stephenh, line 35: One option would be to make this type:

```
useOptionals: ['messages' | 'primitives'];
```

And then around in here:

https://github.com/stephenh/ts-proto/blob/main/src/options.ts#L122

Do a translation of `true` -> `[messages, primitives]`, `false` / `none` -> `[]`.

Hm, I guess `parseParameter` is not fancy enough to like detect `,` and do a `.split(`,`)`. It probably could though, and seems like a reusable thing.

If you want to add this as part of this PR, that'd be great, but also fine to merge as-is.

stephenh (approved):
@sgielen looks great! Thanks for iterations on this.

If you want to also update the readme, that'd be great, but that's also fine to do as a follow up/later.

stephenh:
@sgielen I went ahead and rebased this due to a minor import conflict and am going to merge. Thanks for the PR!

oyvindwe:
A little late, but great to see this PR, as we had a local patch to do the same thing (we parse JSON responses from our gRPC servers directly into types generated by `ts-proto` to avoid including the `fromPartial` methods for hundreds of messages in our webapp).

However, [README.markdown](https://github.com/stephenh/ts-proto/blob/main/README.markdown) is not updated with `useOptionals=messages` and `useOptionals=all`, so it was a bit hard to discover this feature.

stephenh:
@oyvindwe thanks for the call out; I've updated the `useOptionals` docs in the readme, feel free to suggest further improvements. Thanks!

PR #405, opened 2021-11-22
fix: use Long.fromValue instead of Long.fromString

SchroederChris:
for improved robustness regarding already parsed objects

as discussed [here](https://github.com/stephenh/ts-proto/pull/395#issuecomment-975525988)

### DIFF
--- a/integration/simple-long/simple.ts
+++ b/integration/simple-long/simple.ts
@@ -227,7 +227,7 @@ export const SimpleWithMap = {
     );
     message.longLookup = Object.entries(object.longLookup ?? {}).reduce<{ [key: string]: Long }>(
       (acc, [key, value]) => {
-        acc[key] = Long.fromString(value as string);
+        acc[key] = Long.fromValue(value as Long | string);
         return acc;
       },
       {}
--- a/src/main.ts
+++ b/src/main.ts
@@ -990,7 +990,7 @@ function generateFromJson(ctx: Context, fullName: string, messageDesc: Descripto
                 return code`${utils.bytesFromBase64}(${from} as string)`;
               }
             } else if (isLong(valueType) && options.forceLong === LongOption.LONG) {
-              return code`Long.fromString(${from} as string)`;
+              return code`Long.fromValue(${from} as Long | string)`;
             } else if (isEnum(valueType)) {
               return code`${from} as number`;
             } else {


SchroederChris:
Sorry for the failing build step - codegen somehow doesn't fail on my machine 🤷‍♂️

integration/simple-long/google/protobuf/wrappers.ts
```diff
@@ -218,7 +218,7 @@ export const Int64Value = {
 
   fromJSON(object: any): Int64Value {
     const message = { ...baseInt64Value } as Int64Value;
-    message.value = object.value !== undefined && object.value !== null ? Long.fromString(object.value) : Long.ZERO;
+    message.value = object.value !== undefined && object.value !== null ? Long.fromValue(object.value) : Long.ZERO;
```
stephenh, line 221: Hm, I'm kind of surprised this line changed, b/c the only line you changed in `main.ts` was:

```
code`Long.fromValue(${from} as string)`;
```

And there is no `as string` here. Is there another change to `main.ts` that is not in the PR?

integration/simple-long/simple.ts
```diff
@@ -226,7 +226,7 @@ export const SimpleWithMap = {
     message.longLookup = {};
     if (object.longLookup !== undefined && object.longLookup !== null) {
       Object.entries(object.longLookup).forEach(([key, value]) => {
-        message.longLookup[key] = Long.fromString(value as string);
+        message.longLookup[key] = Long.fromValue(value as string);
```
stephenh, line 229: Given we're calling `fromValue` now we can/should remove the `as string`?

stephenh: Hm, maybe just `as Long | string` since those are the cases we intend to support? I think if I saw just `as string` at some point in the future, I'd forgot about this PR and think "well this should be just `Long.fromString`. :-) But the `as string | Long` I think would be a good reminder of our intent.

SchroederChris: Yes, good idea! I know the problem :-D I changed it and also added a link to this comment in the commit message.

integration/simple-long/google/protobuf/wrappers.ts
SchroederChris: After looking at it again, the corresponding place in `main.ts` should be line 1013. It might be a good idea to also change that to `fromValue`. But I wasn't sure if that's actually the right place. Maybe you could have a look, too.

stephenh (approved):

stephenh:
Rebased on `main` after some fixes to the CI pipeline, so going to merge. Thanks @SchroederChris !

SchroederChris:
Thanks @stephenh!

PR #408, opened 2021-11-23
feat:  Add support for 'json_name' annotation

ajaykarthikr:
Inspired by #210, this PR adds support for `json_name` as per @stephenh's feedback on that PR.
> Hm, my interpretation of json_name attribute would be that this particular name would still be name and then other_name would only show up on the JSON on the wire (i.e. in toJson methods) and then looked for when data is parsed in from the wire (i.e. in the fromJson methods).

I made changes to `generateFromJson` and `generateToJson` functions and created a new function `determineFieldJsonName` for json field names. If the `json_name` is not present in the proto file it uses the name attribute for json field names.

There's also a simple unit test for testing this feature.


src/utils.ts
```diff
@@ -167,3 +173,13 @@ export class FormattedMethodDescriptor implements MethodDescriptorProto {
     return result;
   }
 }
+
+export function determineFieldJsonName(field: FieldDescriptorProto, options: Options): string {
+  // By default jsonName is camelCased by the protocol compilier unless the user has
+  // set a "json_name" option on this field. So, if field name is already snake_case
+  // then this will keep jsonName also snake_case.
+  if (field.jsonName.length > 0 && options.snakeToCamel) {
```
stephenh, line 181: Hm, we should probably just check `field.jsonName.length` here and skip checking `options.snakeToCamel`? I.e. if there is a `jsonName` we should use it regardless of what the `snakeToCamel` flag is, I think...

stephenh (approved):

stephenh:
@ajaykarthikr thanks, looks great! One small question about the `snakeToCamel` check, but otherwise we can merge and release. Thanks!

src/utils.ts
ajaykarthikr: Hi, I was actually confused with how to handle that case. I also agree with your suggestion regarding `snakeToCamel` flag.

ajaykarthikr:
@stephenh I don't think there's a way to know if the user had actually set the `json_name` field because that field is never empty according to the docs. 
> JSON name of this field. The value is set by protocol compiler. If the user has set a "json_name" option on this field, that option's value will be used. Otherwise, it's deduced from the field's name by converting it to camelCase.

So, I just removed the `snakeToCamel` flag check according to your previous feedback. This will `lowerCameCase` json field names if `json_name` option is not present.

Anyway, if you read the [language guide](https://developers.google.com/protocol-buffers/docs/proto3#json) for proto3, `lowerCamelCase` is the default behavior. So we should `lowerCamelCase` it by default for compatibility reasons too. 

If there's a usecase, we can probably give an option to the user to dynamically change casing to `lowerCamelCase` or `snake_case`. Similarly to this [betterproto](https://github.com/danielgtaylor/python-betterproto#json). I can make a seperate PR for this. What do you say?

stephenh:
@ajaykarthikr ah, gotcha, I didn't realize that `protoc` always fills in the json name for us...

Okay, yeah, I think this PR makes sense / is great.

It is technically a breaking change for users that were using `snakeToCamel=false`, but I think that's fine for this PR.

If users report breaking changes and want to keep the old behavior, I think maybe turning `snakeToCamel` into multi-values like `true` (default) | `false` (current behavior in this PR) | `keys-and-json` (new behavior that would keep snake case in JSON).

Going to hit merge; thanks!

PR #410, opened 2021-11-23
feat: Improve map reading (fromJSON/fromPartial)

webmaster128:
This follows the same idea as previous PRs for fromJSON/fromPartial: have exactly one assignment per field.

The explicit typing in reduce is necessary to avoid that TS infers `{}` as the output/accumulator type. `${fieldType}` could be replaced with `typeof message.outputField` but then in https://github.com/stephenh/ts-proto/pull/402 the output field might become optional but `${fieldType}` should not change.


stephenh (approved):
Nice!

PR #419, opened 2021-11-28
docs: improve readme section on generic service definitions

aikoven:
I added a small description of what the generic definitions are and added a link to [nice-grpc](https://github.com/deeplay-io/nice-grpc) as an example of a library that utilizes this feature.


stephenh:
Nice!

PR #420, opened 2021-11-29
fix: Respect stringEnums option in wrap function

boukeversteegh:
Fixes type error in generated `Value.wrap` when `stringEnums=true`


integration/pbjs.sh
```diff
@@ -60,3 +60,7 @@ yarn pbts --no-comments -o integration/struct/pbjs.d.ts integration/struct/pbjs.
 # value/
 yarn pbjs --force-message --force-number -t static-module -o integration/value/pbjs.js integration/value/value.proto
 yarn pbts --no-comments -o integration/value/pbjs.d.ts integration/value/pbjs.js
+
+# value/
+yarn pbjs --force-message --force-number -t static-module -o integration/nullvalue/pbjs.js integration/nullvalue/nullvalue.proto
+yarn pbts --no-comments -o integration/nullvalue/pbjs.d.ts integration/nullvalue/pbjs.js
```
stephenh, line 66: Hm, we can probably skip this? We only need the `pbjs` output if we have unit tests that are explicitly testing encode/decode operations against the `pbjs` output as a compliance check.

stephenh (approved):
Ah nice! Makes sense.

stephenh:
Hm, maybe add a struct to the existing string-enum example? Just to keep our `integration/*` sprawl done a bit:

https://github.com/stephenh/ts-proto/tree/main/integration/simple-string-enums

boukeversteegh:
Will do! Thanks for checking

stephenh:
Looks great!

PR #429, opened 2021-12-05
fix: support mutliple options in snakeToCamel flag 

### LINKED ISSUE #423: version 1.90.0 broke snakeToCamel=false functionality
Hey, as you already discussed in #408, that PR introduced breaking changes when using `snakeToCamel=false`.

In our case, the json we're receiving contains snake_case properties, so all `fromJSON` methods accessing those properties are broken with the new version.

I'd very much appreciate getting back the old behavior as you suggested with the `keys-and-json` option.

ajaykarthikr:
Restores `snakeToCamel=false` functionality due the breaking change introduced by #210. Implements changes as per the discussion in #423. 

Now `snakeToCamel: boolean` is `snakeToCamel: boolean | string` and accepts string input too `keys, json`.

This is how the option will be mapped:
- snakeToCamel=true (the default) --> ["keys", "json"]
- snakeToCamel=false --> []
- snakeToCamel=keys --> ["keys"], use camel cased message keys, but still snake case in to/from JSON
- snakeToCamel=json --> ["json"], use proto3-spec camel casing in to/fromJSON, but still snake case in message keys
- snakeToCamel=keys,json --> ["keys","json"]

Resolves #423

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -30,7 +30,7 @@ export enum ServiceOption {
 
 export type Options = {
   context: boolean;
-  snakeToCamel: boolean;
+  snakeToCamel: boolean | string;
   forceLong: LongOption;
   useOptionals: boolean;
   useDate: DateOption;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -9,7 +9,7 @@ import {
 import ReadStream = NodeJS.ReadStream;
 import { SourceDescription } from './sourceInfo';
 import { Options, ServiceOption } from './options';
-import { camelCase, maybeSnakeToCamel } from './case';
+import { camelCase, stringToSnakeCase } from './case';
 
 export function protoFilesToGenerate(request: CodeGeneratorRequest): FileDescriptorProto[] {
   return request.protoFile.filter((f) => request.fileToGenerate.includes(f.name));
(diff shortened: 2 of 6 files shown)


src/options.ts
```diff
@@ -30,7 +30,7 @@ export enum ServiceOption {
 
 export type Options = {
   context: boolean;
-  snakeToCamel: boolean;
+  snakeToCamel: boolean | string;
```
stephenh, line 33: Ah yeah; this makes sense, although what might be nifty is to make this:

```typescript
  snakeToCamel: Array<'json' | 'keys'>;
```

So that the call sites like `maybeSnakeToCamel` and do just `options.snakeToCamel.includes('keys')` and then inside of `optionsFromParamter` we can handle "boolean --> string[]" kinda like:

```typescript
  if ((options.snakeToCamel as any) === false) {
    options.snakeToCamel = [];
  } else if ((options.snakeToCamel as any) === true) {
    options.snakeToCamel = ['keys', 'json'];
  }
```

Which has a few `as any`s which is a little weird, but we've done it for a few other options like this have that "might be boolean/might be strings".

Maybe try that? I think it will simplify the call sites quite a bit.

src/utils.ts
```diff
@@ -177,8 +177,10 @@ export class FormattedMethodDescriptor implements MethodDescriptorProto {
...
   // By default jsonName is camelCased by the protocol compilier unless the user has
   // set a "json_name" option on this field.
-  if (field.jsonName.length > 0) {
+  const fromStringSnakeToCamel = typeof options.snakeToCamel === 'string' && options.snakeToCamel.includes('json');
+  const fromBooleanSnakeToCamel = options.snakeToCamel === true;
+  if (fromBooleanSnakeToCamel || fromStringSnakeToCamel) {
     return field.jsonName;
   }
-  return maybeSnakeToCamel(field.name, options);
+  return stringToSnakeCase(field.name);
```
stephenh, line 185: Ah yeah...I see what you're thinking here, but maybe just leave this out?

I'm thinking that if the user has set `snakeToCamel=false` that's not _exactly_ the same thing as "camelToSnake=true" i.e. requesting that we actively snake-ize fields in the JSON. So I'm tempted to just leave the field name as it is.

src/options.ts
stephenh: @ajaykarthikr I'm going to go ahead and merge this and do ^ as a follow up. Thanks for the PR!

PR #430, opened 2021-12-06
docs(README): add watch mode instructions

juanlatorre:
Add instructions for watch mode using `chokidar-cli`


stephenh (approved):
Great, thanks!

Issue #383, opened 2021-11-10
Proposal: Optional Generated File Suffix


sallustfire:
ts-proto is great and we've been using it for a while in production. However, having the generated typescript files matching the name of the protobuf file with just the `.ts` extension is increasingly becoming a maintenance problem. For instance, there is no convenient way to gitignore these generated file with a glob if they are adjacent to non-generated sources.

Other widely used protoc plugins seem introduce suffixes as a matter of convention. For a `helloworld.proto`

1. grpc-go => `helloworld.pb.go`
2. protoc js_out=import_style=commonjs => `helloworld_pb.js`
3. protoc python_out => `helloworld_pb2.py`
etc

Unless there is already a mechanism for appending proto suffixes to the generated TS files which are also correctly used in the imports, we'd like to see a mechanism like `--ts_proto_opt=suffix=_pb` which would produce `helloworld_pb.ts`. Since it can default to empty, it should be backwards compatible. 

Hopefully with a little direction, we'd be willing to work on a PR to introduce this functionality.

sallustfire:
@stephenh Great! I'm working on a commit when I have free time. Might have a few questions before I open a PR.

PR #431, opened 2021-12-06
feat: Allow optional suffix for generated files

sallustfire:
Implements the functionality described in #383 which allows for custom suffixes on generated files, which can be specified via the new plugin option `fileSuffix`.


sallustfire:
@stephenh Ok, opened the pull request. Please provide feedback. 

Due to how the code generation handles imports, it was a little more invasive that I was initially expecting. The `ts-poet` `imp` clauses are scattered across the codebase. It might be worth consolidating that logic, but it could be beyond the scope of this issue.

src/main.ts
```diff
@@ -721,7 +722,7 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
         }
       } else if (isEnum(field)) {
         if (options.stringEnums) {
-          const fromJson = getEnumMethod(typeMap, field.typeName, 'FromJSON');
+          const fromJson = getEnumMethod(ctx, field.typeName, 'FromJSON');
```
stephenh, line 725: Ah yeah, good call passing in the whole `ctx` instead of both `typeMap` + `options`.

src/encode.ts
```diff
@@ -4,28 +4,29 @@ import { messageToTypeName, wrapperTypeName } from './types';
...
 export function generateEncoder(ctx: Context, typeName: string): Code {
+  const { options } = ctx;
   const name = wrapperTypeName(typeName);
   if (!name) {
     return code`${messageToTypeName(ctx, typeName, { keepValueType: true })}.encode(value).finish()`;
   }
 
   if (name == 'Timestamp') {
-    const TimestampValue = imp(`${name}@./google/protobuf/timestamp`);
+    const TimestampValue = imp(`${name}@./google/protobuf/timestamp${options.fileSuffix}`);
```
stephenh, line 14: Wdyt about making an `imp` helper method in `src/utils` that looks something like:

```
imp(ctx, name, `./google/protobuf/timestamp`);
```

And then does the stitching together there? Might be nice to centralize the `fileSuffix` handling.

stephenh (approved):
@sallustfire nice, looks great! Let me know if you want to make the `imp` helper method update; if not, np, I'm good to merge this as-is.

src/encode.ts
sallustfire: @stephenh Ok, I created a helper `impProto` to consolidate the logic as described, the suffix avoids naming collisions since there are imports of public JS libraries too.  All of the imports had template literals with relative paths, so it's really a style call on whether you consider the `'./'` part of the module name or not.

There is also a circular dependency caused by makeUtils if the context is one of the function parameters. Just using the options avoids this.

Anyway, happy to make changes here if you have different preferences.

stephenh:
@sallustfire cool, looks great!

sallustfire:
@stephenh Thanks, that was very fast!

PR #435, opened 2021-12-08
round to integer if they are of an integer type

yoavmil:
Do not merge untill you check it!
I cannot compile and check it on my computer because I dont have Docker and cannot install it too.


stephenh:
Picked up and merged in #444.

PR #436, opened 2021-12-08
fix: noImplicitReturns error in Value.unwrap

### LINKED ISSUE #432: struct unwrap fails TypeScript's `noImplicitReturns` check
Since 1.88.0 was released, the code to unwrap a struct has failed TypeScript's `noImplicitReturns` check, which is normally a sane check to have on.

In 1.88.0, the function [`unwrapAnyValue` was added](https://github.com/stephenh/ts-proto/blob/v1.88.0/integration/struct/google/protobuf/struct.ts#L406), without an explicit return (even though you may know one of the conditionals is guaranteed to be true), resulting in:

```
src/gen/google/protobuf/struct.ts:415:4 - error TS7030: Not all code paths return a value.

415 ): string | number | boolean | Object | null | Array<any> | undefined {
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is still the case, though in a [slightly different form in the latest 1.91](https://github.com/stephenh/ts-proto/blob/v1.92.1/integration/struct/google/protobuf/struct.ts#L355), still resulting in the lint error:

```
src/gen/google/protobuf/struct.ts:397:6 - error TS7030: Not all code paths return a value.

397   ): string | number | boolean | Object | null | Array<any> | undefined {
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

boukeversteegh:
Fixes #432

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.sh text eol=lf
 parameters.txt text eol=lf
+protoc-gen-dump text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.sh text eol=lf
+parameters.txt text eol=lf
--- a/integration/grpc-js/google/protobuf/struct.ts
+++ b/integration/grpc-js/google/protobuf/struct.ts
@@ -353,19 +353,20 @@ export const Value = {
   },
 
   unwrap(message: Value): string | number | boolean | Object | null | Array<any> | undefined {
-    if (message.stringValue !== undefined) {
+    if (message?.stringValue !== undefined) {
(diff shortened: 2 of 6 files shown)


boukeversteegh:
Thanks to the dockerization, I was able to easily run the scripts on a holiday laptop. However, automatic line-conversion made the shell script use `CRLF` instead of `LF`, which is OK when running bash on windows, but within docker, Linux tripped over those, so I added a `.gitattributes` file that makes sure that all `.sh` scripts are always checked out as `LF`. This is cross-platform compatible.

stephenh:
> Thanks to the dockerization, I was able to easily run the scripts on a holiday laptop

:100: 

> so I added a .gitattributes file

Ah nice, good idea!

PR #438, opened 2021-12-08
fix: dockerized protoc aliases

boukeversteegh:
Restores the dockerized version of `protoc`, for standalone use, as documented under [Readme.md#development](https://github.com/stephenh/ts-proto#development)

`Dockerfile` and `yarn` commands not touched. Added a default _docker-compose_ mount of `$PWD:/host`, this should not affect other uses of the container as the container does not automatically enter this directory, and the project directory is mounted in case $PWD is not set (e.g. when running without TTY)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 *.sh text eol=lf
 parameters.txt text eol=lf
 protoc-gen-dump text eol=lf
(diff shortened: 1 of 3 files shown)


stephenh (approved):
Thanks!

PR #448, opened 2021-12-20
docs: Clarify fromJSON/toJSON format in README

lorensr:
On first glance, I thought "Protobuf JS format" meant the `protobufjs` format, which is not spec-compliant. I edited to clarity. I also changed `i.e.` to `e.g.` because the former implies that's the only thing that makes the method spec-compliant. I imagine it correctly encodes all the other types as well, not just `Timestamp`?

image

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -159,7 +159,7 @@ creating a class and calling the right getters/setters.
 
   (Configurable with the `useDate` parameter.)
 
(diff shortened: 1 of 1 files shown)


stephenh:
@lorensr thanks for the update; fwiw yes the intention is to support the proto3 json spec, the only disclaimer is that it's been implemented kind of type-by-type / as-needed, so I'm not 100% sure things like duration and any have been implemented yet. But basically yes. :-) Thanks!

PR #450, opened 2021-12-21
feat: `enumsAsLiterals` option

### LINKED ISSUE #379: Option to convert Enum to Union Type
Enum Type is a JavaScript that is created as IIFE when generate JS. It is difficult to take Tree Shaking. How about adding options to convert to Union Type?

### as-is

```
export enum MOBILE_OS {
  IOS = 'iOS',
  ANDROID = 'Android'
}
```

when generate js =>
```
export var MOBILE_OS;
(function (MOBILE_OS) {
    MOBILE_OS[MOBILE_OS["IOS"] = 0] = "IOS";
    MOBILE_OS[MOBILE_OS["ANDROID"] = 1] = "ANDROID";
})(MOBILE_OS || (MOBILE_OS = {}));
```

### to-be

```
export const MOBILE_OS = {
  IOS: 'iOS',
  Android: 'Android'
} as const;
type MOBILE_OS = typeof MOBILE_OS[keyof typeof MOBILE_OS]; // 'iOS' | 'Android'
```

when generate js =>
```
export const MOBILE_OS = {
    IOS: 'iOS',
    Android: 'Android'
};
```

otofu-square:
Close #379 

This PR implements ~~`enumAsConst`~~ `enumsAsLiterals` option.


stephenh (approved):
Hey @otofu-square thanks for the PR!

This looks great. Two suggestions:

1. Can you update the readme to include the new option?
2. What do you think about renaming `enumAsConst` to `enumsAsLiterals`? My thinking is that `enumAsConst` makes me think of "const enums" which are a separate thing, and so using the term "literal" is still describing the change in output, but without using the "const" term.

Otherwise looks great, thanks!

otofu-square:
@stephenh 
Thank you for your review and suggestions!
I think they are both good suggestions, so let me adopt them both. 🚀

stephenh:
Looks great, thanks @otofu-square !

PR #452, opened 2021-12-21
fix: add service property to client constructor generator

DavyJohnes:
According to [this](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/make-client.ts#L173) client constructor has `service` property which contains `ServiceDefinition`. It can be useful to access it in some cases.


stephenh:
Hey @DavyJohnes , thanks for the PR! I made exactly your change in #455 just b/c it was easier to re-run codegen and then add a small test. Thanks for the fix!

PR #456, opened 2021-12-27
feat: add an option to disable Exact types

### LINKED ISSUE #454: Problems with Exact types
After the Exact types were introduced in #412, the [nice-grpc](https://github.com/deeplay-io/nice-grpc) integration with `ts-proto` stopped working.

The reason is that `nice-grpc` automatically wraps client requests with `fromPartial`, e.g.:

```protobuf
service TestService {
  rpc TestMethod(TestMessage) returns (TestMessage);
}

message TestMessage {
  string test_field = 1;
  string another_field = 2;
}
```

Before Exact types, the generated client would have a type roughly equal to:

```ts
type TestServiceClient = {
  testMethod(request: DeepPartial<TestMessage>): Promise<TestMessage>;
}
```

So instead of

```ts
client.testMethod(TestMessage.fromPartial({testField: ''}))
```

we could just write

```ts
client.testMethod({testField: ''})
```

The client type is produced using mapped types, and the request parameter is calculated as `Parameters<RequestMessageType['fromPartial']>[0]`. But with Exact types, this type resolves to

```ts
{ ... } & { ... } & Record<string | number | symbol, never>
```

which only matches empty object.

Ideally, the generated client type should do the Exact trick in its methods as well, i.e.:

```ts
type TestServiceClient = {
  testMethod<I extends Exact<DeepPartial<TestMessage>, I>>(request: I): Promise<TestMessage>;
}
```

Unfortunately, this [isn't currently possible](https://stackoverflow.com/questions/69855855/can-i-preserve-generics-in-conditional-types) to do with mapped types, since there's no support for higher-kinded types in TypeScript.
 
The one thing I came up with is a helper that can un-exact a type recursively:

```ts
type Unexact<T> = {
  [K in keyof T as string extends K ? never : number extends K ? never : K]: Unexact<T[K]>;
};
```

which takes us back to `DeepPartial<...>` so we can use `Unexact<Parameters<RequestMessageType['fromPartial']>[0]>` as the request parameter. It kinda works, but IDE hints and suggestions are now hardly readable.

If there's no other solution, should we maybe introduce an option to disable Exact types?

aikoven:
Closes #454


stephenh (approved):
I'm still a little bummed about not finding a way for your mapped types to keep these, but am not coming up with anything ... i.e. a way to pass long / "look up" the "appropriate" / "contextual" `DeepPartial` from a given message.

On another project, I've made fake properties, i.e.:

```
class Foo { 
  __types: { encoder: FooEncoder, decoder: FooDecoder }
}
```

And then always just set `__types: null!` because it's not actually used at runtime, but is just a way for mapped types to look into `__types` and grab what it needs.

For that project, the `__types` approach was more appropriate, but it seems more icky here, sitting on protobuf message types... :shrug: 

But yeah, this is a cheap option to add, so why not/makes sense. Maybe add it to the readme, just in case others run into issues with `Exact`?

aikoven:
The hardest problem is preserving the trick with generic parameter required for `Exact` in mapped type.

Things might get better if/when TypeScript supports exact types natively :)

> But yeah, this is a cheap option to add, so why not/makes sense. Maybe add it to the readme, just in case others run into issues with `Exact`?

I've updated the readme, could you review it please?

stephenh (approved):
Looks great, thanks @aikoven !

PR #457, opened 2021-12-27
perf: optimize object creation in `decode`, `fromJSON` and `fromPartial`

aikoven:
Creating an object via object spread can be 100 times slower than creating via object literal: https://jsbench.me/0nkxobf9gt/1

### DIFF
--- a/integration/oneof-unions/oneof.ts
+++ b/integration/oneof-unions/oneof.ts
@@ -69,7 +69,7 @@ export interface SimpleButOptional {
   age?: number | undefined;
 }
 
-const basePleaseChoose: object = { name: '', age: 0 };
+const createBasePleaseChoose = (): PleaseChoose => ({ name: '', age: 0 });
 
 export const PleaseChoose = {
   encode(message: PleaseChoose, writer: Writer = Writer.create()): Writer {
@@ -115,7 +115,7 @@ export const PleaseChoose = {
   decode(input: Reader | Uint8Array, length?: number): PleaseChoose {
     const reader = input instanceof Reader ? input : new Reader(input);
     let end = length === undefined ? reader.len : reader.pos + length;
-    const message = { ...basePleaseChoose } as PleaseChoose;
+    const message = createBasePleaseChoose();
     message.signature = new Uint8Array();
     while (reader.pos < end) {
       const tag = reader.uint32();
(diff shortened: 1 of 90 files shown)


stephenh (approved):
Nice!

aikoven:
@stephenh could you please take another look?

Now all properties are initialized in the base instance, including `repeated`, `optional`, `map` and `bytes`.

stephenh (approved):
Nice! Great idea to just initialize all the fields while we're at it.

webmaster128:
Nice. We might consider removing the base object creation (`create*Message` functions) alltogether and simply ensure every field is set in every creator function. WDYT?

aikoven:
> Nice. We might consider removing the base object creation (`create*Message` functions) alltogether and simply ensure every field is set in every creator function. WDYT?

Yep, at first I thought that a separate function is needed to prevent duplication (that can result in e.g. increased bundle size). But actually an object literal with default values is only needed in the `decode` method, and `fromJson` and `fromPartial` can simply create an object with actual values right away.

integration/oneof-unions/oneof.ts
```diff
@@ -115,8 +117,7 @@ export const PleaseChoose = {
   decode(input: Reader | Uint8Array, length?: number): PleaseChoose {
     const reader = input instanceof Reader ? input : new Reader(input);
     let end = length === undefined ? reader.len : reader.pos + length;
-    const message = { ...basePleaseChoose } as PleaseChoose;
-    message.signature = new Uint8Array();
+    const message = createBasePleaseChoose();
```
webmaster128, line 120: Here the creator function is actually nice to have. If it creates all fields, its pretty simple.

Does it make sense to just rename create**Base**PleaseChoose to create**Empty**PleaseChoose and leave it for the decode implementation?

integration/simple-deprecated-fields/simple.ts
```diff
@@ -32,7 +32,9 @@ export interface Child {
   name: string;
 }
 
-const baseSimple: object = { name: '', age: 0, testField: '', testNotDeprecated: '' };
+function createBaseSimple(): Simple {
+  return { name: '', age: 0, child: undefined, testField: '', testNotDeprecated: '' };
```
webmaster128, line 36: Nice to see this creating the correct type with all fields and avoid the invalid cast that we had before!

webmaster128:
> actually an object literal with default values is only needed in the decode method, and fromJson and fromPartial can simply create an object with actual values right away

Exactly. This changed a lot over the last months when I migrated many of the assignments to the ?? operator or `map` such that now in fromPartial/fromJSON each field should be set in exactly one assignment now, either with a value or its default. The base object was valuable before those changes when many fields were set in an `if` condition only if they where set.

aikoven:
By the way, consider that `createBaseMsg()` is actually the same as `Msg.fromPartial({})`. Which means that we could extract the logic of `fromPartial` instead, and use it in `encode`, thus reducing the generated code size.

A downside of it is that support for `outputPartialMethods=false` would be more complicated.

Issue #233, opened 2021-03-07
[Feature Request] Decoding delimited messages


ChristianIvicevic:
I'd like to start a discussion since I am not sure whether my feature request is actually a reasonable scope for this library.

Since clients can receive fragmented data over a socket I had to implement a custom logic to handle framed messages that the server encodes using length-prefixes with the `google::protobuf::util::SerializeDelimitedToZeroCopyStream` method. Basically all packets sent over the wire receive a varint as a prefix that denotes the actual message length. My current prototype for handling this looks as follows:
```typescript
let client: Socket;
let buffer = Buffer.alloc(0);
let messageLength = Infinity;

client.on('data', data => {
  buffer = Buffer.concat([buffer, data]);
  if (messageLength === Infinity && buffer.length > 0) {
    try {
      messageLength = varint.decode(buffer);
      buffer = buffer.slice(varint.decode.bytes);
    } catch (e: unknown) {
      // RangeError from varint.decode indicated wrong encoding or incomplete data.
    }
  }
  if (buffer.length >= messageLength) {
    const encodedMessage = buffer.slice(0, messageLength);
    const message = WrappedPacket.decode(encodedMessage);
    // ...do something with message...
    buffer = buffer.slice(messageLength);
    messageLength = Infinity;
  }
});
```
It seems like the `protobufjs` code generator exposes methods for reading such delimited messages, but I cannot use the library since the generated Typescript typings include unsupported namespaces. Therefore I sticked to your library and was curious whether handling length-prefixed messages may be a good feature for it or should be an implementation detail that consumers define themselves.

stephenh:
Hey, sure, good question.

As far as I understand, what you're describing is typically handled by higher-level protocols like grpc?

Assuming that's right, then, yeah, dunno like ts-proto does technically generate framework-specific code for say Twirp or GRPC or NestJS. So it's not out of the realm of possibility.

But given you're doing very low-level / custom / socket level stuff that I generally assume most users aren't/won't do, and instead use some existing higher-level protocol/library, yeah, seems unlikely to be a great fit for ts-proto.

Kinda depends on what you need though.

And, if anything, ts-proto I assert is pretty hackable, if you want to just maintain your own fork.

Just out of curiosity, is there a reason you're doing direct-socket-level handling? Disclaimer I'm not an expert on all of the approaches/solutions in the same other than the obvious ones like grpc (which I know have HTTP/2 couplings/complexities).

ChristianIvicevic:
[...]

PR #460, opened 2021-12-31
feat: `delimitedMethods` option

AlaaZorkane:
## Related issue(s)
https://github.com/stephenh/ts-proto/issues/233

## This PR adds the following:

- [x] `encodeDelimited` generator
- [x] `decodeDelimited` generator
- [x] `delimitedMethods` option to generate above methods
- [x] `delimiatedMethods` usage documentation 
- [ ] Tests for above methods

## Notes

I tried to find an API where I don't have to copy paste redundant generation logic but I failed to make it backward-compatible, per example, one of the approaches is to have an option instead of having specific functions like this:
```js
decode(input: _m0.Reader | Uint8Array, length?: number, isDelimited = false) {}
encode(message: JarAccountState, writer: _m0.Writer = _m0.Writer.create(), isDelimited = false) {}
```

But this would appose these problems:
- The length will be ignored but needs to be passed on (or `null) explicitly.
- You have to pass a `writer` or `null` explicitly

I also couldn't find a way to not have redundant code for adding cases for each incoming field due to my unfamiliarity with the codebase. 

## Useful links
Protobuf streaming: https://developers.google.com/protocol-buffers/docs/techniques#streaming
Protobuf-js implementations: [encodeDelimited](https://github.com/protobufjs/protobuf.js/blob/95b56817ef6fb9bdcb14d956c159da49d0889bff/src/type.js#L494) and [decodeDelimited](https://github.com/protobufjs/protobuf.js/blob/95b56817ef6fb9bdcb14d956c159da49d0889bff/src/type.js#L517)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -159,6 +159,10 @@ export function generateFile(ctx: Context, fileDesc: FileDescriptorProto): [stri
         if (options.outputEncodeMethods) {
           staticMembers.push(generateEncode(ctx, fullName, message));
           staticMembers.push(generateDecode(ctx, fullName, message));
+          if (options.delimitedMethods) {
+            staticMembers.push(generateEncodeDelimited(ctx, fullName, message));
+            staticMembers.push(generateDecodeDelimited(ctx, fullName, message));
+          }
         }
         if (options.outputJsonMethods) {
           staticMembers.push(generateFromJson(ctx, fullName, message));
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -811,6 +815,127 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
+  const { options, utils, typeMap } = ctx;
+  const chunks: Code[] = [];
+
+  // create the basic function declaration
+  chunks.push(code`
+    decodeDelimited(
+      input: ${Reader} | Uint8Array,
+    ): ${fullName} {
+      const reader = input instanceof ${Reader} ? input : new ${Reader}(input);
+      let end = reader.uint32();
```
stephenh, line 828: Instead of repeating the entire contents of `decode`, could this be something like:

```typescript
const reader = ...
let end = reader.uint32();
return decode(reader, end);
```

I.e. `decode` already accepts an optional `length` argument, which I think? is the only thing that has changed here?

src/main.ts
```diff
@@ -978,6 +1103,172 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
+function generateEncodeDelimited(ctx: Context, fullName: string, messageDesc: DescriptorProto): Code {
+  const { options, utils, typeMap } = ctx;
+  const chunks: Code[] = [];
+
+  // create the basic function declaration
+  chunks.push(code`
+    encodeDelimited(
+      ${messageDesc.field.length > 0 ? 'message' : '_'}: ${fullName},
+      writer: ${Writer} = ${Writer}.create(),
+    ): ${Writer} {
```
stephenh, line 1116: Same musing, can `encodeDelimited` just call `encode` after writing the prefix? Also, is the prefix being written? I'm scanning it but don't see it right now.

stephenh (commented):
@AlaaZorkane thanks for the PR!

I had a few questions, and then also if you could make a new `integration-tests/delimited-methods` directory with a minimal proto and a test that round trips like:

```
const foo1 = FooMessage.fromPartial({ ... });
const foo2 = FooMessage.decodeDelimited(FooMessage.encodeDelimited(foo1));
expect(foo2).toEqual(foo1);
```

That'd be great. Thanks!

stephenh:
Also @AlaaZorkane , I'm just curious, I've not personally used delimited messages before, but I get the concept and was reading the docs:

https://developers.google.com/protocol-buffers/docs/techniques#streaming

One thing I'm curious about is, if there are multiple messages in a stream coming in, how do we know which message is message? Like if a stream if `[FooMessage, BarMessage]` then with delimited lengths it might be `[200, FooMessage, 300, BarMessage]`, i.e. we know to read 200 bytes for `foo` and 300 bytes for `bar`, but how do we know, after reading `FooMesasge`, that `BarMessage` is what comes next?

Or is the assumption that the caller of `decodeDelimited` knows that there will be a specific / fixed order of messages?

src/main.ts
AlaaZorkane: You're right! Since `delimitedMethods` depends on `outputEncodeMethods` we can definitely reuse that logic.

AlaaZorkane:
Hey @stephenh - I added the integration tests and the reusing refactor! Would appreciate if you can review them please 🙏

Also, about streaming multiple messages, I think implementations can vary from application to another? And it's mostly a higher level topic that won't be included and/or abstracted in decoding/encoding primitives (like `decodeDelimited`).

There isn't really a set of best-practices on how to deal with the case of a stream of **different** types of messages AFAIK, and the assumption I made when I read the `streaming` docs is for use cases like [50, FooMessage, 100, FooMessage, ...]

src/main.ts
boukeversteegh: @AlaaZorkane could you clarify how the length delimiter is being written before calling encode  here?

I am not familiar with the encoding logic, but I understand it as follows:

- `.ldelim()` writes a length-prefix, and then writes the message content _after_ it? So basically `message.ldelim()` outputs `[length, message]`
- encode itself already calls `ldelim()`, so it writes its own length delimiter
- you're calling `encode().ldelim()`, so that means an _extra_ delimiter is inserted.

Questions:
- Is this description correct?
- Do we then not write 2 delimiters each time we encode?
- If encoding already outputs its own length prefix _before_, why do we need to delimit messages at all?

AlaaZorkane: @boukeversteegh Of course! AFAIK:

- Yes an extra delimiter for the **whole** message as in `[length, message]` which in details can be per example: `[length, tag, length, data, tag, length, data...]`
- We do, but one is for the last type/value in a message (which includes a lot of types/values) not for the **whole** message.
- Protocol buffers are not length delimited by default, they length delimits encoded values inside of them, think of use cases where you have the value `42` encoded as `int64`, pb won't encode all `8 bytes` (I think there is an exception for `fixed` scalars but I never used them before) - to know how many bytes are encoded we must prefix the specific value `42` with a length

I made this little drawing that's missing a lot of stuff but kinda explains what I'm trying to say:

image


This is a good article to read about delimiting **whole** messages for streaming purposes: https://sureshjoshi.com/development/streaming-protocol-buffers

integration/delimited-methods/delimited-methods.ts
```diff
@@ -0,0 +1,117 @@
...
+        default:
+          reader.skipType(tag & 7);
+          break;
+      }
+    }
+    return message;
+  },
+
+  encodeDelimited(message: Simple, writer: Writer = Writer.create()): Writer {
+    return this.encode(message, writer).ldelim();
```
stephenh, line 56: I'm looking at some of the code we create today and it does this:

```
      Nested_InnerMessage.encode(message.message, writer.uint32(18).fork()).ldelim();
```

And the docs for `ldelim` say "Resets to the last state and appends the fork state's current write length as a varint followed by its operations."

This makes me think we need to ensure a fork exists, and change this to be:

```
return this.encode(message, writer.fork()).ldelim();
```

Does that seem right?

AlaaZorkane: You're right! It seems like it's the case in the `protobuf-js` implementation [here](https://github.com/protobufjs/protobuf.js/blob/95b56817ef6fb9bdcb14d956c159da49d0889bff/src/type.js#L495).

boukeversteegh: I have a feeling that this should have been caught by a test. Would it be possible to change the tests to write and read 2 messages instead of 1? Then we will really test the main use-case of this feature @AlaaZorkane

integration/delimited-methods/delimited-methods-test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    expect(length).toEqual(delimiter);
+  });
+
+  it('decodes length-delimited messages', () => {
+    const encoded = Simple.encodeDelimited(message).finish();
+    const decoded = Simple.decodeDelimited(encoded);
+
+    expect(decoded).toEqual(message);
+  });
```
boukeversteegh, line 25: How about changing this test (or adding one) to encode and decode 2 messages?

It will give us more confidence that the main use case is working, and also demonstrate how to use the feature in practice.

boukeversteegh (changes requested):

src/main.ts
boukeversteegh: Thank for your explanation @AlaaZorkane . It makes it more clear now!

integration/delimited-methods/delimited-methods-test.ts
AlaaZorkane: Definitely! Just added streaming tests via 0ec5e53, please let me know what do you think. Thanks!

src/main.ts
```diff
@@ -978,6 +1001,24 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
 
+/** Creates a function to encode a message by loop overing the tags then delimits the encoding with a length prefix */
+function generateEncodeDelimited(ctx: Context, fullName: string, messageDesc: DescriptorProto): Code {
+  const { options, utils, typeMap } = ctx;
+  const chunks: Code[] = [];
+
+  // create the basic function declaration
+  chunks.push(code`
+    encodeDelimited(
+      ${messageDesc.field.length > 0 ? 'message' : '_'}: ${fullName},
```
stephenh, line 1012: Really minor, but I think this `message || _` variable name was for `encode` whose `for` loop would actually never use `message` if there are no fields. For `encodeDelimited`, the function body always uses `message`, so this can be just `message: ${fullName}`

src/main.ts
```diff
@@ -811,6 +815,25 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
...
+  // create the basic function declaration
+  chunks.push(code`
+    decodeDelimited(
+      input: ${Reader} | Uint8Array,
+    ): ${fullName} {
+      const reader = input instanceof ${Reader} ? input : new ${Reader}(input);
+      const length = reader.uint32();
+  `);
+  chunks.push(code`const message = this.decode(reader, length);`);
+  chunks.push(code`return message;`);
```
stephenh, line 831: Maybe just `return this.decode(...)`?

stephenh (approved):
One small thing about `message` vs. `_` but otherwise looks great!

src/main.ts
AlaaZorkane: Yep, definitely!

src/main.ts
AlaaZorkane: Removed in b0f5e38, thanks!

boukeversteegh (approved):
Nice tests added! Looks good to me 👍

stephenh:
@AlaaZorkane looks great! Looks like just the build is failing, maybe b/c some output needs updated after getting rebased:

https://github.com/stephenh/ts-proto/runs/4766206207?check_suite_focus=true

AlaaZorkane:
> @AlaaZorkane looks great! Looks like just the build is failing, maybe b/c some output needs updated after getting rebased:
> 
> https://github.com/stephenh/ts-proto/runs/4766206207?check_suite_focus=true

Hey @stephenh, sorry can you walk me through it? I already ran the codegen commands and updated the `.bin/.ts` outputs to conform with last changes.

stephenh:
@AlaaZorkane sure, np; I pulled your PR down locally, and AFAICT just rebasing on main and then re-running codegen for your new `delimited-methods` test (i.e. `yarn bin2ts delimited-methods`) did update the `delimited-methods.ts` file to match what the CI build expected.

I'd push my change up to this PR just to be sure, but I can't push to your `main` branch; so probably try that on your side and then I think the build will be clean?

Thanks!

PR #463, opened 2022-01-05
perf: fromJSON returns object literal to allow v8 optimizations

boukeversteegh:
Felt like doing something useful and understood from https://github.com/stephenh/ts-proto/pull/462/files/2649b9aeb6c35779dc8294575fa70b7a35c5f79d#r778884590 that the plan was to return objects directly from `fromJSON`, rather than constructing a base object and assigning the fields.

But I may have been too quick, as then I found out that @aikoven is already working on this as well, so please close this PR if needed.

before:

```ts
 fromJSON(object: any): Simple {
    const message = createBaseSimple();
    message.name = isSet(object.name) ? String(object.name) : '';
    message.age = isSet(object.age) ? Number(object.age) : 0;
    return message;
  };
```

after:
```ts
fromJSON(object: any): Simple {
    return {
      name: isSet(object.name) ? String(object.name) : '',
      age: isSet(object.age) ? Number(object.age) : 0,
    };
 },
```

### DIFF
--- a/integration/batching-with-context/batching.ts
+++ b/integration/batching-with-context/batching.ts
@@ -77,9 +77,9 @@ export const BatchQueryRequest = {
   },
 
   fromJSON(object: any): BatchQueryRequest {
-    const message = createBaseBatchQueryRequest();
-    message.ids = Array.isArray(object?.ids) ? object.ids.map((e: any) => String(e)) : [];
-    return message;
+    return {
+      ids: Array.isArray(object?.ids) ? object.ids.map((e: any) => String(e)) : [],
+    };
   },
 
   toJSON(message: BatchQueryRequest): unknown {
@@ -130,9 +130,9 @@ export const BatchQueryResponse = {
   },
 
   fromJSON(object: any): BatchQueryResponse {
-    const message = createBaseBatchQueryResponse();
-    message.entities = Array.isArray(object?.entities) ? object.entities.map((e: any) => Entity.fromJSON(e)) : [];
-    return message;
+    return {
+      entities: Array.isArray(object?.entities) ? object.entities.map((e: any) => Entity.fromJSON(e)) : [],
+    };
   },
 
   toJSON(message: BatchQueryResponse): unknown {
@@ -183,9 +183,9 @@ export const BatchMapQueryRequest = {
   },
 
   fromJSON(object: any): BatchMapQueryRequest {
-    const message = createBaseBatchMapQueryRequest();
-    message.ids = Array.isArray(object?.ids) ? object.ids.map((e: any) => String(e)) : [];
-    return message;
+    return {
+      ids: Array.isArray(object?.ids) ? object.ids.map((e: any) => String(e)) : [],
+    };
   },
 
   toJSON(message: BatchMapQueryRequest): unknown {
@@ -239,12 +239,14 @@ export const BatchMapQueryResponse = {
   },
 
   fromJSON(object: any): BatchMapQueryResponse {
-    const message = createBaseBatchMapQueryResponse();
-    message.entities = Object.entries(object.entities ?? {}).reduce<{ [key: string]: Entity }>((acc, [key, value]) => {
-      acc[key] = Entity.fromJSON(value);
-      return acc;
-    }, {});
-    return message;
+    return {
+      entities: isObject(object.entities)
+        ? Object.entries(object.entities).reduce<{ [key: string]: Entity }>((acc, [key, value]) => {
+            acc[key] = Entity.fromJSON(value);
+            return acc;
+          }, {})
+        : {},
+    };
   },
 
   toJSON(message: BatchMapQueryResponse): unknown {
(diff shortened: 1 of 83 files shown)


integration/batching-with-context/batching.ts
```diff
@@ -239,12 +239,14 @@ export const BatchMapQueryResponse = {
...
-      return acc;
-    }, {});
-    return message;
+    return {
+      entities: isObject(object.entities)
+        ? Object.entries(object.entities).reduce<{ [key: string]: Entity }>((acc, [key, value]) => {
+            acc[key] = Entity.fromJSON(value);
+            return acc;
+          }, {})
+        : {},
```
webmaster128, line 248: Any reason for the `object.entities ?? {}` => `isObject() ? X : Y` change other than taste? it's probably a bit easier to review but is more code that has to be send to web users.

integration/batching-with-context/batching.ts
```diff
@@ -503,8 +505,7 @@ export const WriteMethodResponse = {
   },
 
   fromJSON(_: any): WriteMethodResponse {
-    const message = createBaseWriteMethodResponse();
-    return message;
+    return {};
```
webmaster128, line 508: 😍

integration/oneof-unions/oneof.ts
```diff
@@ -174,39 +174,32 @@ export const PleaseChoose = {
...
+        ? { $case: 'aString', aString: String(object.aString) }
+        : isSet(object.aMessage)
+        ? { $case: 'aMessage', aMessage: PleaseChoose_Submessage.fromJSON(object.aMessage) }
+        : isSet(object.aBool)
+        ? { $case: 'aBool', aBool: Boolean(object.aBool) }
+        : isSet(object.bunchaBytes)
+        ? { $case: 'bunchaBytes', bunchaBytes: bytesFromBase64(object.bunchaBytes) }
+        : isSet(object.anEnum)
+        ? { $case: 'anEnum', anEnum: pleaseChoose_StateEnumFromJSON(object.anEnum) }
+        : undefined,
```
webmaster128, line 191: Love this single assignment implementation

webmaster128 (commented):
Great stuff

integration/batching-with-context/batching.ts
boukeversteegh: Not sure, but I think I was trying to keep the pattern of "if there's a value, read the value, otherwise use the default", to make it more easily readable.

integration/oneof-unions/oneof.ts
boukeversteegh: Haha, that's a relief 😅 It's a bit unconventional, but I didn't have any better alternatives

integration/batching-with-context/batching.ts
webmaster128: Yeah, sure. The _iterate over an empty object_ approach is not the most obvious.

src/main.ts
```diff
@@ -1111,57 +1116,71 @@ function generateFromJson(ctx: Context, fullName: string, messageDesc: Descripto
...
+        .filter(isWithinOneOf)
+        .filter((otherField) => otherField.oneofIndex === field.oneofIndex);
+
       const oneofName = maybeSnakeToCamel(messageDesc.oneofDecl[field.oneofIndex].name, options);
-      chunks.push(code`
-        message.${oneofName} = { $case: '${fieldName}', ${fieldName}: ${readSnippet(`object.${jsonName}`)} }
-      `);
-      chunks.push(code`}`);
+
+      // First case, output field name
```
stephenh, line 1144: Ah nice, I see...

Seems like this is relying on all of the oneofFields being grouped together, sequentially in the incoming `messageDesc.field` list.

I wouldn't be surprised if this is only true for our relatively clean `*.proto` test files, and that real-world proto files this may not be the case? Maybe? Not sure.

Actually I guess if the order is based on `*.proto` order than we should be good, so maybe this is just fine. I'm good keeping it as it is, b/c it's simpler this way, and we can see if anyone ends up reporting not-actually-in-order oneof fields as users try it out.

stephenh (approved):
This is great! Thanks @boukeversteegh !

Fwiw I'm pretty sure @aikoven was more specifically looking at optimizing `fromPartial` and not necessarily `fromJSON`, so I'm pretty good merging this as-is. Doesn't hurt to wait for him to chime in, but I'll defer either way to you @boukeversteegh .

aikoven:
Looks great!

I didn't yet start working on object literals for `fromPartial`, so no conflict here.

src/main.ts
boukeversteegh: Yeah, this one got me thinking for some time as well...

As far as I can reason through this, it shouldn't matter in what order the fields are delivered, and even whether there are random unrelated fields in between.

We are generating the fields in order of `messageDesc.field`, and the index checks (first / last) are based on a subset of the same sequence. So no matter how the order might be messed up, the first will always be the first etc. If that makes sense.

boukeversteegh: I realize that the above merely guarantees correct detection of the _first_ and _last_ item, but if indeed _somehow_ oneof field members were interspersed with other random fields, we would mess up the output.

However, I do wish to rely on the protobuf sdk to keep the fields in the declared order, and in that case, it should all work out.

I've changed this part of the code a bit to be more readable, hope that helps.

PR #465, opened 2022-01-06
feat: yarn watch updates (specified) tests when source files change

boukeversteegh:
Attempting to further improve developer workflow. Hoping - but not sure - that I didn't go too far, I would like feedback on these additions separately.

- Let `yarn watch` monitor source directories as well
  - Tell the user what to do when the sources change
  - Press Enter to recompile ALL proto files
  - Run `yarn watch TEST1 TEST2` to automatically recompile TEST1 and TEST2 when sources change

```
$ yarn watch -h

Watches the integration directory for changes and regenerates .bin and .ts files.
Watches the src directory and regenerates given TEST files.

Usage:
    $ yarn watch [options] [TEST, TEST2, ...]
    $ ts-node watch.ts [options] [TEST, TEST2, ...]

Options:
    -h, --help    Show this help message.
    --polling     Use polling instead of native watchers.
    TEST          Regenerate the specified TEST(s) when implementation files change.
                  Equivalent to running 'yarn bin2ts TEST' manually each time.

Examples:
    $ yarn watch
    $ yarn watch --polling
    $ yarn watch simple struct
```

# Example run

```
$ yarn watch
yarn run v1.22.10
$ ts-node integration/watch.ts

Watching for changes:
- integration/  Regenerating integration/*/*.proto to .bin and .ts

Press enter to regenerate code for all .proto files.
Press ctrl+c or q to exit

src/main.ts [watch] Plugin modified! Press [enter] to regenerate all integration tests or run 'yarn watch [TEST, ...]'. See 'yarn watch --help'.
```

### DIFF
--- a/integration/watch.ts
+++ b/integration/watch.ts
@@ -24,21 +24,10 @@ function main() {
   if (process.argv.includes('-h') || process.argv.includes('--help')) showHelp();
   showSettings(tests);
 
-  chokidar
-    .watch('integration/*/*.proto', watchOptions)
-    .on('all', integrationHandler(yarn, 'proto2bin'));
-
-  chokidar
-    .watch('integration/*/*.proto', watchOptions)
-    .on('all', integrationHandler(yarn, 'proto2pbjs'));
-
-  chokidar
-    .watch('integration/*/*.bin', watchOptions)
-    .on('all', integrationHandler(yarn, 'bin2ts'));
-
-  chokidar
-    .watch('src/**/*.ts', watchOptions) //
-    .on('all', sourceHandler(yarn, 'bin2ts', tests));
+  chokidar.watch('integration/*/*.proto', watchOptions).on('all', integrationHandler(yarn, 'proto2bin'));
+  chokidar.watch('integration/*/*.proto', watchOptions).on('all', integrationHandler(yarn, 'proto2pbjs'));
+  chokidar.watch('integration/*/*.bin', watchOptions).on('all', integrationHandler(yarn, 'bin2ts'));
+  chokidar.watch('src/**/*.ts', watchOptions).on('change', srcHandler(yarn, 'bin2ts', tests));
 
   setupKeys({
     ['']: () => yarnRun(yarn, 'bin2ts', 'enter'),
@@ -71,34 +60,31 @@ Usage:
 Options:
     -h, --help    Show this help message.
     --polling     Use polling instead of native watchers.
-    TEST          Recompile the specified TEST(s) when implementation files change.
+    TEST          Regenerate the specified TEST(s) when implementation files change.
                   Equivalent to running 'yarn bin2ts TEST' manually each time.
     
 Examples:
     $ yarn watch
     $ yarn watch --polling
-    $ yarn watch simple struct
-        `);
+    $ yarn watch simple struct`);
   process.exit(0);
 }
 
 function showSettings(tests: string[]) {
   console.log(`${colors.yellow}`);
   console.log(`Watching for changes:`);
-  console.log('- integration/\tRecompiling .proto to .bin and .ts');
+  console.log('- integration/\tRegenerating integration/*/*.proto to .bin and .ts');
   if (tests.length) {
-    console.log(`- src/\t\tRecompiling .proto to .ts\n\t\t(${tests.join(', ')})`);
+    console.log(`- src/\t\tRegenerating integration/{${tests.join(', ')}}/*.proto to .ts`);
   }
-  console.log('\nPress enter to recompile all .proto files.');
+  console.log('\nPress enter to regenerate all integration tests.');
   console.log('Press ctrl+c or q to exit');
   console.log(colors.reset);
 }
 
 function integrationHandler(yarn: string, task: string) {
   return (event: FsEvent, triggerPath: string) => {
-    if (event !== 'add' && event !== 'change') {
-      return;
-    }
+    if (event !== 'add' && event !== 'change') return;
 
     triggerPath = triggerPath.replace(/\\/g, '/'); // windows
     const relativePath = triggerPath.replace(/^integration\//, '');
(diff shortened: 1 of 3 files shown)


integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
+  process.exit(0);
+}
 
-    chokidar
-      .watch("*/*.proto", { ignoreInitial: true, usePolling  })
-      .on('all', yarnRunHandler(yarn, 'proto2bin'));
+function showSettings(tests: string[]) {
+  console.log(`${colors.yellow}`);
+  console.log(`Watching for changes:`);
+  console.log('- integration/\tRecompiling .proto to .bin and .ts');
```
stephenh, line 88: Maybe use the term `Regenerating`

integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
+  if (tests.length) {
+    console.log(`- src/\t\tRecompiling .proto to .ts\n\t\t(${tests.join(', ')})`);
```
stephenh, line 90: Maybe include a prefix of `integration/${tests.join(...)}` to highlight its subdirectories of `integration`.

integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
+  }
+  console.log('\nPress enter to recompile all .proto files.');
```
stephenh, line 92: Maybe "to regenerate all integration tests"? Or "to codegen all integration tests", granted I like the term codegen and regenerate is probably better.

integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
+      console.log(formatLog(colors.green, header, task, 'Done'));
     }
+  });
 }
 
-function yarnRun(yarn: string, task: string, path: string) {
-    const yarnArgs = [task, path];
-
-    console.log(formatLog(colors.green, task, path, `${yarn} ${yarnArgs.join(' ')}`));
+function sourceHandler(yarn: string, task: string, tests: string[]) {
```
stephenh, line 128: Maybe `srcHandler` to highlight its handling the `src/` dir?

integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
-    yarnProcess.stderr.on('data', (data: Buffer) => console.error(formatLog(colors.red, task, path, data.toString())));
-    yarnProcess.on('error', (err: Error) => console.error(formatLog(colors.red, task, path, err.message)));
-    yarnProcess.on('close', (code: number) => {
-        if (code !== 0) {
-            console.error(formatLog(colors.red, task, path, `Exited with code ${code}`));
-        }
-    });
+    triggerPath = triggerPath.replace(/\\/g, '/');
+    if (tests.length === 0) {
+      let notice = `Source changed! Press [enter] to recompile all .proto files or rerun this command as 'yarn watch [TEST, ...]'. See 'yarn watch --help'.`;
```
stephenh, line 134: Maybe `src/ changed! ... to recodegen / to regenerate`...

integration/watch.ts
```diff
@@ -1,86 +1,150 @@
...
+  const usePolling = process.argv.includes('--polling');
+  const tests = process.argv.slice(2).filter((x) => !x.startsWith('-'));
+  const watchOptions = { ignoreInitial: true, usePolling };
+
+  if (process.argv.includes('-h') || process.argv.includes('--help')) showHelp();
+  showSettings(tests);
+
+  chokidar
+    .watch('integration/*/*.proto', watchOptions)
+    .on('all', integrationHandler(yarn, 'proto2bin'));
```
stephenh, line 29: Pretty minor/mostly just curious, but can we pass `'add', 'change'` here to `on` instead of passing `'all'` and later dropping any non-add/non-change?

stephenh (approved):
Even more amazing! Thank you!

integration/watch.ts
boukeversteegh: I tend to think full names are more clear than abbreviations, but sure :-)

integration/watch.ts
boukeversteegh: The message is prefixed with the affected path: `src/main.ts [watch] Source changed! ...`.

I used "source" to indicate that it's not the tests that have changed, but the actual plugin. Not so much to show which directory it is. I wonder if another name would be better to express this? "Plugin changed", "Code generation logic changed", "Source code changed!"?

boukeversteegh: Changed to

`src/main.ts [watch] Plugin modified! Press [enter] to regenerate all integration tests or run 'yarn watch [TEST, ...]'. See 'yarn watch --help'.`

for now

integration/watch.ts
boukeversteegh: Unfortunately not... the api only permits literal strings that are one of ` 'all' | 'add' | 'addDir' | 'change' | 'unlink' | 'unlinkDir'`. Passing a comma delimited string breaks the listener.

But I did change the `srcHandler` to only listen to `change` instead of `all` and then filtering out the rest. There I needed only 'change'.

integration/watch.ts
stephenh: Ah gotcha, I hadn't actually tried it, so didn't realize `src/` was already part of the path. But that's great, thanks!

stephenh (approved):
I'm pretty excited to use this; hacking on ts-proto is going to be so easy now :-)

boukeversteegh:
> I'm pretty excited to use this; hacking on ts-proto is going to be so easy now :-)

Cool 👍 Hope it will prove useful!

The developer experience of ts-proto is quite nice once you have a PR open. I like how everything gets released automatically, so you can immediately enjoy the improvements without waiting for manual actions. The only thing that was hard was getting started and figuring out the commands. Hopefully this is now a lot better. It would be great if more devs will start playing with the source now and create pull requests. Cheers!

Issue #471, opened 2022-01-10
outputServices=grpc-js always generates client stub


alexeyten:
I've tried to use options `--ts_proto_opt=esModuleInterop=true,outputServices=grpc-js,outputClientImpl=false`, but `interface TestClient extends Client { ... }` and `const TestClient = makeGenericClientConstructor(...)` are always generated.

I guess it's because there is no options check in https://github.com/stephenh/ts-proto/blob/main/src/generate-grpc-js.ts#L41-L42

So far, I've found that `outputClientImpl` option has effect only for `outputServices=default`.

stephenh:
Closing as fixed in #472 ; thanks!

PR #472, opened 2022-01-10
fix: respect generateClientImpl=false in grpc-js (#471)

alexeyten:
Do not generate client stub and constructor when `outputServices=grpc-js` and `outputClientImpl=false`


stephenh:
Great, thanks @alexeyten !

PR #473, opened 2022-01-12
feat: add support for unknown fields

Vilsol:
Implements unknown fields as discussed in #437 behind the `unknownFields` feature flag.

I had to resort to using undocumented API for writing the proto, as it's apparently unintended to be able to just write out arbitrary bytes to the buffer:
```js
(writer as any)['_push'](
  (val: Uint8Array, buf: Buffer, pos: number) => buf.set(val, pos),
  value.length,
  value
);
```

The test makes sure that decoding -> encoding -> decoding works.


stephenh (approved):
Wow, this is great; the changes to `main.ts` were much smaller/cleaner than I anticipated. Nice!

PR #477, opened 2022-01-17
feat: `removeEnumPrefix` option

### LINKED ISSUE #228: Drop shared enum prefixes
It's very common to write your protobuf enums like this:
```proto
enum Foo {
   FOO_BAR = 0;
   FOO_BAZ = 1;
}
```

It would be nice if there was an option (and later the default) to drop the shared prefix (if there is one), so that the TS enums can be generated like this:
```ts
enum Foo {
   BAR = 0;
   BAZ = 1;
}
```

Toofifty:
This change adds a `removeEnumPrefix` option that allows for removing the enum name from enum members and values in generated code.

Fixes #228

Using enum members with prefixes is a common practice and is part of the [Buf style guide](https://docs.buf.build/best-practices/style-guide#enums), but fully prefixed enums aren't super useful as they wouldn't match other APIs/data sources.

This option simply removes the (upper snake) enum name from the members and values of the enum.
So
```proto
enum FooBar {
  FOO_BAR_BAZ = 1;
  FOO_BAR_QUX = 2;
}
```

would become

```ts
export enum FooBar {
  BAZ = "BAZ",
  QUX = "QUX",
}
```

instead of

```ts
export enum FooBar {
  FOO_BAR_BAZ = "FOO_BAR_BAZ",
  FOO_BAR_QUX = "FOO_BAR_QUX",
}
```

(with stringEnums enabled)

---

Happy to discuss the option name since, as pointed out in #228, there may be other/more configurable options that may be needed. This is the only situation I've encountered however, so it's all I've implemented in this PR.


src/enums.ts
```diff
@@ -30,9 +30,12 @@ export function generateEnum(
 
   enumDesc.value.forEach((valueDesc, index) => {
     const info = sourceInfo.lookup(Fields.enum.value, index);
-    maybeAddComment(info, chunks, valueDesc.options?.deprecated, `${valueDesc.name} - `);
+    const valueName = options.removeEnumPrefix
+      ? valueDesc.name.replace(`${camelToSnake(fullName)}_`, '')
+      : valueDesc.name;
```
stephenh, line 35: I think we do this logic in a few places, maybe add a helper method like `getValueName(ctx, valueDesc)` that internally does the `options.removeEnumPrefix` check?

stephenh (approved):
@Toofifty this looks great! One small ask for a helper method, but otherwise will be great to merge.

Toofifty:
@stephenh I thought about this a bit more and decided to split it into two option values: `all` and `members`:

  `all`: `FooBar.FOO_BAR_BAZ = "FOO_BAR_BAZ"` will generate `FooBar.BAZ = "BAZ"`

  `members`: `FooBar.FOO_BAR_BAZ = "FOO_BAR_BAZ"` will generate `FooBar.BAZ = "FOO_BAR_BAZ"`

I imagine there might be some circumstances where a user would want to have nicer member names in code, but can't transform the string beforehand (though I'm not sure, I'm still pretty new to gRPC). Both will have the same effect for number enums.

Let me know what you think :)

stephenh:
@Toofifty shoot! Sorry for not responding to your comment sooner. Per your question:

> all: FooBar.FOO_BAR_BAZ = "FOO_BAR_BAZ" will generate FooBar.BAZ = "BAZ"
> members: FooBar.FOO_BAR_BAZ = "FOO_BAR_BAZ" will generate FooBar.BAZ = "FOO_BAR_BAZ"

I'm really hesitant to support the `all` option, b/c it seems to much like changing a value that would go on the wire/something.

...that said, I think that's not actually the case, b/c the `.proto` file / and (non-JSON wire) format will always use the indexes (or the full name for JSON)...

So I don't think it's not strictly wrong...but, dunno, still tempted to stick with the simpler `removeEnumPrefix=true` and wait for an explicit request / use case for `stringEnums`-that-also-removes-prefixes-in-values.

Does that sound okay?

boukeversteegh:
> I'm really hesitant to support the `all` option, b/c it seems to much like changing a value that would go on the wire/something.

Nice feature @Toofifty !

I would agree to keep the string values intact. At datastack for example, protobuf is used as a contract specifying JSON-messages, and enums are passed as strings (to make the API human readable). Changing the string values would break the server-side parsing of enums.

As I understand, the proposal is to make that behavior optional, but if it's really only a hypothetical use-case, I would also not include it, just to keep the code-base as small as possible.

Other than that, I would vote for this PR.

superclaw:
Why it hasn't been merged yet? It is really useful feature!

maxgr0:
@stephenh is there any progress on this? Any help needed?🙂

stephenh:
Ah yeah, @maxgr0 it's been awhile but scanning the PR I think I was hoping to see the feature trimmed down a little bit, and be just `removeEnumPrefix=true` that only changes the enum symbols, like:

```
enum Foo {
   FOO_BAR = 0;
   FOO_BAZ = 1;
}
// becomes
enum Foo {
   BAR = 0;
   BAZ = 1;
}
```

And not touch the values, like if the string enums it's still be:

```
enum Foo {
   BAR = "FOO_BAR";
   BAZ = "FOO_BAZ";
}
```

If that makes sense and you'd like to take over the PR, that'd be great. (Of course also happy if @Toofifty would like to come back to this.) Thanks!

superclaw:
@stephenh some time ago I had forked your repository and did special parameter, which allows to remove prefixes without changing values, but I didn't make PR. Can I make new PR, or continue this?

stephenh:
@superclaw ah sure, a new PR would be great!

superclaw:
@stephenh Based on this PR with `removeEnumPrefix=members` option:
https://github.com/stephenh/ts-proto/pull/779

stephenh:
This landed in #779 , thanks @Toofifty / @superclaw !

Issue #218, opened 2021-02-15
Object properties missing since 1.61.0


webmaster128:
I just tried to upgrade ts-proto [from 1.60.0 to 1.61.0](https://github.com/stephenh/ts-proto/compare/v1.60.0...v1.61.0) and notices that many of my tests start failing. This is because properties of objects are missing, such that Jasmine's toEqual does not find them anymore. The tests use externally generated proto serialization with default fields omitted.

It is always about default values. I wonder if there is a different way to implement https://github.com/stephenh/ts-proto/commit/136cd28e5d2e0e708d0cc150d64d890c7dfcbd37. With `hasOwnProperty` being false, the properties become pretty much invisible (e.g. for `Object.keys(a)`/`Object.entries(a)`).

------

Test fails:

```
@cosmjs/stargate: **************************************************
@cosmjs/stargate: *                    Failures                    *
@cosmjs/stargate: **************************************************
@cosmjs/stargate: 1) AuthExtension account works for unused account
@cosmjs/stargate:   - Expected object to have properties
@cosmjs/stargate:       sequence: 0
@cosmjs/stargate:   /home/circleci/project/packages/stargate/src/queries/auth.spec.ts:25:23
@cosmjs/stargate:         assert(account);
@cosmjs/stargate:         expect(account).toEqual({
@cosmjs/stargate:                         ~
@cosmjs/stargate:           address: unused.address,
@cosmjs/stargate:           // pubKey not set
@cosmjs/stargate:   internal/process/next_tick.js:68:7
@cosmjs/stargate:   jasmine-spec-reporter: unable to open 'internal/process/next_tick.js'
@cosmjs/stargate:   Error: ENOENT: no such file or directory, open 'internal/process/next_tick.js'
@cosmjs/stargate: 2) AuthExtension account works for account with pubkey and non-zero sequence
@cosmjs/stargate:   - Expected object to have properties
@cosmjs/stargate:       accountNumber: 0
@cosmjs/stargate:   /home/circleci/project/packages/stargate/src/queries/auth.spec.ts:42:23
@cosmjs/stargate:         const pubkey = encodePubkey(validator.pubkey);
@cosmjs/stargate:         expect(account).toEqual({
@cosmjs/stargate:                         ~
@cosmjs/stargate:           address: validator.delegatorAddress,
@cosmjs/stargate:           pubKey: Any.fromPartial(pubkey),
@cosmjs/stargate:   internal/process/next_tick.js:68:7
[...]

Issue #139, opened 2020-09-21
Support proto2 optional fields


philikon:
Now that we have support for proto3 optional (#73, yay!), should we also add support for proto2's version, `LABEL_OPTIONAL`? Right now something like `optional uint32 index = 42` translates into `index: number` but IMHO it should be `index?: number`.

I'd be happy to write the code for this.

stephenh:
@philikon sure, that sounds like a good idea.

ssilve1989:
I am using a patch of `ts-proto` in Bazel + NestJS to achieve this and opened  PR #162 with some additional changes on top of what my patch had, to modify `fromPartial` and `fromJson` and I think it should be ok

stephenh:
I looked a little bit at this, and the proto2 docs on `optional`:

https://developers.google.com/protocol-buffers/docs/proto#optional

> When a message is parsed, if it does not contain an optional element, the corresponding field
> in the parsed object is set to the default value for that field.

So, if we have an `optional foo` proto, should it become a `foo: number` b/c when decoded off the wire, it will (?) always be set to `0` if not present, or should it be `foo?: number` and left as undefined?

Fwiw my guess is that users mostly expect `foo?: number` but the docs insinuate otherwise.

Another option would be to have it be `foo: number` but get the default value from a prototype, so that you could still check for presence via `message.hasOwnProperty('foo')`. That is probably where I would lean.

ssilve1989:
Optionals are not deserialized into default values. The reader should check if there are any bytes at the field position and if not it gets turned into the language representation of undefined. For ex in JS that's `undefined`, in Scala it's an Optional of `None`, etc

stephenh:
> Optionals are not deserialized into default values.

I mean, I agree intuitively, that's just not what the proto docs I linked to above assert...

I.e. you say "turned into the language representation of `undefined`" but they say "turned into the field's default value".

Maybe the nuance is that in Scala/class-based messages, the internal field can be `undefined` but the generated getter function could still return the default value?

Dunno. I mean, I generally agree with you, and maybe we just do that anyway, even if its not technically what the docs assert (or I'm misunderstanding them).

ssilve1989:
> I mean, I agree intuitively, that's just not what the proto docs I linked to above assert...

Yes it is. The docs say 

>  well-formed message may or may not contain an optional element. When a message is parsed, if it does not contain an optional element, the corresponding field in the parsed object is set to the default value for that field. 

[...]

PR #479, opened 2022-01-17
feat: enable unknown fields for descriptor protos

Vilsol:
Builds upon #473 to add support for options in #437.

I had to also enable `--downlevelIteration` tsc flag as that is now a requirement due to looping over `Uint8Array`.

The tests will most likely need to be updated after the package is published.

### DIFF
--- a/protos/build.sh
+++ b/protos/build.sh
@@ -6,7 +6,7 @@
 protoc \
   --plugin=./node_modules/ts-proto/protoc-gen-ts_proto \
   --ts_proto_out=. \
-  --ts_proto_opt=exportCommonSymbols=false \
+  --ts_proto_opt=exportCommonSymbols=false,unknownFields=true \
   ./google/protobuf/descriptor.proto \
   ./google/protobuf/compiler/plugin.proto
 
@@ -14,7 +14,7 @@ protoc \
   ./index.ts \
   ./google/protobuf/descriptor.ts \
   ./google/protobuf/compiler/plugin.ts \
-  --outDir dist --declaration
+  --outDir dist --declaration --downlevelIteration
 
 
 
--- a/protos/google/protobuf/compiler/plugin.ts
+++ b/protos/google/protobuf/compiler/plugin.ts
@@ -1,6 +1,6 @@
 /* eslint-disable */
-import * as Long from 'long';
 import { util, configure, Writer, Reader } from 'protobufjs/minimal';
+import * as Long from 'long';
 import { FileDescriptorProto, GeneratedCodeInfo } from '../../../google/protobuf/descriptor';
 
 /** The version number of protocol compiler. */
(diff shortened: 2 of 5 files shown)


stephenh:
@Vilsol looks great! The `ts-proto-descriptors` package is still manually published, so I just did that locally, and this should be in npm as version `1.4.0`.

Fwiw I _might_ have pinned ts-proto-descriptors to ts-proto that had optional-field behavior that worked with protobuf v2's `oneofIndex`...

Like at one point (which is probably still true today), how ts-proto handled optional fields meant that (with the ts-proto-descriptors generated by that version of ts-proto) we couldn't tell the difference between `oneofIndex=0`-because-its-unset and `oneofIndex=0`-because-its-actually-zero. 

I think it was this commit:

https://github.com/stephenh/ts-proto/commit/3e78405e42e6507c76f6513339c8c43d5ef3fe1c

...more context in this issue https://github.com/stephenh/ts-proto/issues/218

...and then it goes back to this one https://github.com/stephenh/ts-proto/issues/139 which is basically highlighting that ts-proto does not have the `hazzer` methods that most protobuf generated messages do (b/c we're "just data"), so we can't apply the behavior of "the data is 0 ... but also hasData is false".

So...you can try bumping ts-proto-descriptors to 1.4.0 and see what happens, but after re-remembering things, I'm 90% sure you're going to hit this limitation. :-/ 

Wdyt? Any good ideas?

Vilsol:
@stephenh It looks like indeed that causes a lot of issues. One alternative we could have is provide yet another metadata field behind some option flag like `definedFields`.

All it would do is add the field name to a list (or set) when it was decoded. So this:
```js
case 9:
    message.oneofIndex = reader.int32();
    break;
```
would become
```js
case 9:
    message._definedFields[tag] = __spread((message._definedFields[tag] || []), ['oneofIndex']);
    message.oneofIndex = reader.int32();
    break;
```

I see how this is an ugly solution, but I can't think of any other that wouldn't also impact other users of the generator.

stephenh:
@Vilsol yeah... I guess if we have to go the flag route (which I think we do...) we could add a flag that opts into the `Object.create` behavior that is in the currently-pinned version of ts-proto that ts-proto-descriptors uses.

I.e. it put the defaults into a prototype, so the  `hazzer` for "is `oneOfIndex` explicit or defaulted" was "if it's in the object's own keys, it was explicitly set".

Personally I really like that behavior, it just sucks that `Object.keys` does not pull in keys from the prototype chain (I get having an "own" version of it like `Object.getOwnkeys` or what not, but would be nice if the default `Object.keys` included the prototype...like that's what the prototype is for imo...), and so it was a large breaking change to ts-proto, and had to be rolled back.

But, complaining about `Object.keys` aside, I think some sort of `usePrototypeForDefaults=true`, which is what that reverted PR did, would let us enable it for ts-proto-descriptors and then correctly handle oneofs...

The upshot would also be that then anyone else who really needs `hazzer` methods could use the same flag...

Wdyt? Do you want to try and resurrect that old PR? I can as well, disclaimer not sure when I'd get to it...

Vilsol:
@stephenh Seems like that is the only way out of this. When you say PR, I assume you meant the issue #139 ?

Either way, I can try implementing such a flag, seems like it would just be the 2 changes to `main.ts` in this commit, no? https://github.com/stephenh/ts-proto/commit/3e78405e42e6507c76f6513339c8c43d5ef3fe1c#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3d

PR #484, opened 2022-01-20
feat: add usePrototypeForDefaults option

Vilsol:
This adds a feature flag `usePrototypeForDefaults` as discussed in #479

I have tested this by building `ts-proto` with this change, rebuilding the `ts-proto-descriptors` lib using the rebuilt `ts-proto`, then linking to the new `ts-proto-descriptors` from `ts-proto` and re-building all integrations and testing them.


integration/simple-prototype-defaults/simple-test.ts
```diff
@@ -0,0 +1,22 @@
+import { Child_Type, Simple, StateEnum } from './simple';
+
+describe('simple', () => {
+  it('generates types correctly', () => {
```
stephenh, line 4: Maybe add a test that does a `Simple.decode` that shows a from-default value like `age: 0` has `Object.keys().includes("age") === false` or what not?

src/main.ts
```diff
@@ -758,6 +758,11 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
   const { options, utils, typeMap } = ctx;
   const chunks: Code[] = [];
 
+  let createBase = code`createBase${fullName}()`;
+  if (options.usePrototypeForDefaults) {
+    createBase = code`Object.create(${createBase}) as ${fullName}`;
+  }
```
stephenh, line 764: Ah wow, nice how small this change is.

stephenh (approved):
Looks great!

Vilsol:
@stephenh I added some more appropriate tests. I hope that is what you meant.

integration/simple-prototype-defaults/simple-prototype-defaults-test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    expect(Object.keys(decodedUndefined).includes('age')).toBeFalsy();
+    expect(decodedUndefined.hasOwnProperty('age')).toBeFalsy();
+
+    const decodedWithZero = Simple.decode(Simple.encode(Simple.fromPartial({
+      age: 0
+    })).finish());
+
+    expect(Object.keys(decodedWithZero).includes('age')).toBeFalsy();
+    expect(decodedWithZero.hasOwnProperty('age')).toBeFalsy();
```
stephenh, line 15: Ah right right, I'd expected this to be truthy, but it's the whole proto3 thing that optionals are not encoded, so presence is broken. But this will work for proto2, which is what the descriptors comein as.

stephenh:
> I added some more appropriate tests. I hope that is what you meant.

Yep, that's what I meant. Thanks!

stephenh (approved):

PR #489, opened 2022-01-21
feat: Bump ts-proto-descriptors.

stephenh:


Vilsol:
It would probably be best if we stripped it from the output, as the generated `fromPartial` call would strip it anyway. It would also mean that this change would have a persistent output and nothing would change from the previous version.

stephenh:
@Vilsol I've not looked super deeply into this yet, but initially I agree that just dropping them from the meta-typings seems like the right thing to do.

Vilsol (approved):

Vilsol:
I think you can merge this now, I don't think we can avoid having JS reordering the fields when it serializes them for meta-typings.

PR #491, opened 2022-01-22
docs: minor improvements on nestjs documentation

micalevisk:
you can view this version here: https://github.com/micalevisk/ts-proto/blob/patch-1/NESTJS.markdown


stephenh (approved):
Awesome! Thank you!

PR #499, opened 2022-02-04
fix: usePrototypeForDefaults with repeated fix

### LINKED ISSUE #498: usePrototypeForDefaults cause issues with repeated fields
For this proto
```protobuf
message Test {
  repeated string tags = 1;
}
```
the generated code is 
```js
function createBaseTest() {
  return { tags: [] };
}

const message = Object.create(createBaseTest());
// ...
case 1:
  message.tags.push(reader.string());
```

and this will push tags into prototype's `tags` field. And, e.g. console.log(message) will show an “empty” object and `Object.keys().includes("tags")` will be false.

alexeyten:
Just an idea how to fix #498.

Declare repeated fields as “possibly undefined” and create them on demand. So “hazzle check” would still work, but callers will have to create array manually.


integration/simple-prototype-defaults/simple.ts
```diff
@@ -1004,42 +1011,49 @@ export const SimpleWithMap = {
       const tag = reader.uint32();
       switch (tag >>> 3) {
         case 1:
+          message.entitiesById ||= [];
```
pbhuynh, line 1014: Shouldn't all the properties under this `SimpleWithMap` be instantiated with an empty object, i.e. `{}` instead of array?

src/main.ts
```diff
@@ -736,7 +736,7 @@ function generateBaseInstanceFactory(
       : isMapType(ctx, messageDesc, field)
       ? '{}'
```
pbhuynh, line 737: Do we need to account the map case as well?

pbhuynh (commented):
Also running into this issue with my project. What are the next steps something like this fix in @stephenh ? Thanks!

stephenh:
Hi @pbhuynh , just taking a look--sorry @alexeyten , you had some great work here, and I must not have had time to review it/think about it when you first submitted the PR.

In terms of approach, I think if I was personally using `usePrototypeForDefaults` with repeated fields all day, every day, I'd probably _try_ to implement @alexeyten 's "auto-set on read" ideas, which seem pretty neat/ergonomic.

That said, I think what the current PR is doing, just making the repeated fields undefined, is probably more straight-forward/an easier path to getting merged.

Just in general @pbhuynh , I think next steps would be 1) getting the PR rebased and 2) have a clear "failed before, passing now" test/reproduction in the integration tests.

PR #503, opened 2022-02-10
fix: make struct types play well with type registry

aikoven:
With type registry enabled, the code generated from `struct.proto` would not compile due to missing `$type` field.

In this PR all objects are initialized using `createBase<Type>` function.

I also replaced `Struct.fromPartial({})` with `createBaseStruct()`, because `fromPartial` can be disabled.


stephenh (approved):
Looks great! Thanks!

PR #510, opened 2022-02-13
fix: generate canonical JSON encoding for FieldMasks

### LINKED ISSUE #497: Implement FieldMask Well-Known type
https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#fieldmask

This should serialize to JSON as a comma-separated string.
And probably in JS this should be represented as an array of string. E.g. this
```proto
syntax = "proto3";

package dummy;

import "google/protobuf/field_mask.proto";

message GetFields {
  uint32 id = 1;
  google.protobuf.FieldMask fields = 2;
}
```
should be represented as an object `{ id: 1, fields: ['foo', 'bar', 'baz.quux'] }`.
Now it is represented as `{ id: 1, fields: { paths: ['foo', 'bar', 'baz.quux'] }`.

buyology:
Fixes/updates #497.

The [canonical JSON encoding of FieldMasks](https://developers.google.com/protocol-buffers/docs/proto3#json) is a comma-separated string of the `FieldMask.paths`-array. Generate that JSON mapping for FieldMask-fields.

In #497 @alexeyten suggested that

> […] probably in JS this should be represented as an array of string.

This change does not touch the representation of the FieldMask. I checked other generator libs, e.g. the Go generator, and it did not touch the representation of the FieldMask. It is probably also simpler not do so in order to get the correct wire-encoding "for free".


stephenh (approved):
> This change does not touch the representation of the FieldMask.
> I checked other generator libs, e.g. the Go generator, and it did
> not touch the representation of the FieldMask. It is probably also
> simpler not do so in order to get the correct wire-encoding "for free".

Cool, that makes sense. This looks great!

Fwiw I think ts-proto's "idiomatic TypeScript" mission statement would generally lean towards using the `string[]` representation directly in the message interfaces themselves as well, but it's also definitely np to defer that to a separate PR.

integration/fieldmask/fieldmask-test.ts
```diff
@@ -0,0 +1,35 @@
+import { Reader } from 'protobufjs';
+import { FieldMaskMessage } from './fieldmask';
+
+import { FieldMaskMessage as PbFieldMaskMessage } from './pbjs';
```
stephenh, line 4: Fwiw we use the `pbjs` types primarily for sanity checking ts-proto core `encode` / `decode` logic. I feel comfortable enough with that code coverage that I'd be fine skipping pbjs types for these `integration/fieldmask` tests, and just use the ts-proto provided encode/decode.

(Although I guess if we were changing the object representation, e.g. `FieldMessageMesage.fieldMask` as a `string[]`, then yeah we'd probably want to sanity check our encode/decode against a 3rd party.)

(It looks like the CI build failed on a pbjs error as well, so the easiest way of fixing that is to probably just skip it all together. If you want to keep pbjs, I think look at the `integration/pbjs.sh` file; for now we hand-maintain that given that most integration tests don't use pbjs types.)

buyology: I see! I minimized the test to only cover the JSON encoding/decoding.

boukeversteegh:
Awesome! I was just going to use field-masks but refrained after noticing this missing feature. Thanks 👍 

For the typescript representation part, we could use the `wrap/unwrap` approach as we do for Wrapper types.
So it won't be necessary to re-implement the encoding/decoding.

Basically, the FieldMask representation would be kept as-is, but it will be responsible for the JSON coding.

The container messages will:
- generate json for a fieldmask field like this: `FieldMask.ToJson(FieldMask.wrap(message.someFieldMask))`
- and parse json like this: `message.someFieldMask = FieldMask.unwrap(FieldMask.FromJson(obj.someFieldMask))`.

That way we can keep a 'clean' FieldMask concept, which is a true representation of the Protobuf object, which gives the wire-format out of the box, while allowing for TS-friendly types within the container messages.

Additionally, the logic for the canonical json is isolated in separate methods.

Let me know if your thoughts and if you're interested in implementing this. Otherwise I might look into it if I have some time.

boukeversteegh:
hi @buyology , I went ahead to add the above suggestions. Please have a look if this implementation works for you.

See #525

PR #512, opened 2022-02-16
feat: Allow simultaneous services and generic service definitions

fizx:
Hi, we have this as a use case at my company because we use the basic service interfaces yet need runtime reflection, and have been running for a while.  

This patch is backwards compatible and covered by an integration test.

I'm hoping to upstream.  Please let me know how I can make it easier to accept the patch!


src/main.ts
```diff
@@ -735,10 +738,10 @@ function generateBaseInstanceFactory(
     const val = isWithinOneOf(field)
       ? 'undefined'
       : isMapType(ctx, messageDesc, field)
-      ? '{}'
-      : isRepeated(field)
-      ? '[]'
-      : defaultValue(ctx, field);
+        ? '{}'
```
fizx, line 741: prettier did this, not sure if you want to pick it up cause it does make sense.

src/options.ts
```diff
@@ -25,6 +25,7 @@ export enum ServiceOption {
   GRPC = 'grpc-js',
   GENERIC = 'generic-definitions',
   DEFAULT = 'default',
+  GENERIC_AND_SERVICES = 'generic-definitions-and-services',
```
stephenh, line 28: Hm, kinda think, let's maybe not add a new definition and instead change `options.outServices` to `ServiceOption[]`?

I was thinking like `outputServices=generic,default`, but `,` is already taken as a separator.

We could list options twice, like `outputServices=generic-definitions,outputServices=default`... I just made a fix to `snakeToCamel` that provides a precedent for how to support this "nudged to an array" pattern for options:

https://github.com/stephenh/ts-proto/pull/513

Does that seem okay?

stephenh (commented):
Nice, using about using an array of options, but otherwise looks great, and thanks for the PR

fizx:
Still working on the PR right now ...

src/options.ts
fizx: Great!

fizx:
I implemented your suggestion @stephenh!  Tests pass locally, but I don't have permission to run on github.

Note that this PR includes #516 so that I can test locally.  We should probably squash and merge them independently.

fizx:
Question about anticipated behavior when someone does a conflicting option with serviceOption=false,serviceOption=default.  I'm leaving it as undefined, but open to alternatives.

src/options.ts
fizx, line 28: Unneeded, drop this line.

stephenh (approved):
Looks great @fizx , thanks!

PR #521, opened 2022-02-21
fix: support json_name with non-alphanumeric characters for all field types

boukeversteegh:
# Use `array[access] syntax` for _all field types_ with a json_name containing non-alphanumeric characters.

Sorry to bother you once more @stephenh 😅.

I missed a few instances where `${jsonName}` was still used instead of the safe property accessor.
Double checked it now using IDE variable usage, so it should work everywhere now.


stephenh (approved):
Np!

PR #529, opened 2022-03-07
fix: Type Error when map contains string enums #382

### LINKED ISSUE #382: Maps with enum values do not generate fromJSON/fromPartial correctly when stringEnums=true
When `stringEnums` flag is set to true, the generated types for a `map<string, T>` (where T is an enum) do not generate correctly. It seems that the generated code is not respecting that `stringEnums=true` and is assuming the `value` is a number.

```
TS2322: Type 'number' is not assignable to type 'SomeEnumType'.
```

#### Input
```proto
enum SomeEnumType {
  UNDEFINED_SOME_ENUM_TYPE = 0;
  DEFINED_SOME_ENUM_TYPE = 1;
}

message Entity {
  map<string, SomeEnumType> parameters = 1;
}
```

#### Output
```ts

export enum SomeEnumType {
  UNDEFINED_SOME_ENUM_TYPE = 'UNDEFINED_SOME_ENUM_TYPE',
  DEFINED_SOME_ENUM_TYPE = 'DEFINED_SOME_ENUM_TYPE',

}

export interface Entity {
  parameters: { [key: string]: SomeEnumType };
}

// Entity
fromJSON(object: any): Entity {
    const message = { ...baseEntity } as Entity;

    if (object.parameters !== undefined && object.parameters !== null) {
      Object.entries(object.parameters).forEach(([key, value]) => {
        message.parameters[key] = value as number; // <-- error is here
      });
    }
```

boukeversteegh:
Fixes #382


src/main.ts
```diff
@@ -1489,7 +1491,9 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
             if (isBytes(valueType)) {
               return code`${from}`;
             } else if (isEnum(valueType)) {
-              return code`${from} as number`;
+              const mapType = detectMapType(ctx, messageDesc, field);
```
stephenh: Ah nice, @boukeversteegh this looks like a great fix; we were just hard-coded `as number` here before...makes sense.

Re-using `detectMapType` here made a lot of sense, but it made me curious why the code didn't already do that somewhere (given this all "handle maps" related code); I found this line:

```
const valueType = (typeMap.get(field.typeName)![2] as DescriptorProto).field[1];```

Which a) was doing basically what `detectMapType` itself does (but getting the `FieldDescriptorProto` instead of `detectMapType`'s output type), and b) was calling it `valueType` instead of what it probably should be called, `valueField`.

So I pushed up a change to this PR that does that, i.e. keeps your fix but removes the old `const valueType = ...ugly code...` with a call to `detectMapType` which now returns both `valueType` (as before) as well as `valueField`, which is what that line really wanted.

stephenh (approved):
@boukeversteegh looks great!

src/main.ts
boukeversteegh: Thanks a lot for that! I did notice that weird line, but didn't have the time to figure out if it was the same or not, so this is better for sure. Cheers!

PR #533, opened 2022-03-11
feat: import proto as type import if onlyTypes is set

### LINKED ISSUE #367: use `import type` for type only imports
TS has the option `"importsNotUsedAsValues": "error"` which results in an error, when an import is only importing types and no actual values (which is the case in svelte TS projects because of esbuild/vitejs):

```
This import is never used as a value and must use 'import type' because 'importsNotUsedAsValues' is set to 'error'.ts(1371)
```

the fix would be to change 

```
import { Oauth } from "foo";
```

to

```
import type { Oauth } from "foo";
```

given that `Oauth` is an interface / type and no actual services are imported

RSWilli:
closes #367

I added a switch that prepends `t:` if onlyTypes is set

Currently the tests fail though, I don't know how I could fix them


stephenh:
Ah shoot, @RSWilli thanks for the PR; I forgot that ts-proto is still on ts-poet 4.5.0 and ts-poet didn't learn the `t:` --> type import trick until ~4.6.x; can you bump ts-poet to 4.10.0 and that should help things along? Thanks!

stephenh:
@RSWilli everything looked just, I just checked this out locally to run prettier, and then merged to main from there. Thanks!

PR #534, opened 2022-03-15
fix: regression in being able to return a Date as a GRPC return value

fizx:
```
service Clock {
  rpc Now(google.protobuf.Empty) returns (google.protobuf.Timestamp);
}
```

The above will not respect ts_opt useDate=true.  I know it's more idiomatic to wrap the response in a message type, but bear with me--it's legal, used to work fine, and is a pretty terse, relevant interface.

This PR fixes this, adds a regression test.  I do feel like the fix is clunky, and likely doesn't address every instance of similar problems.  Some advice there would be nice.

Also, I don't really know why my M1 laptop insists on making the following 1 chr diff to random *.bins, but it doesn't seem to be affecting anything.  I'll investigate that separately:

```
diff --git a/integration/grpc-js/google/protobuf/wrappers.bin b/integration/grpc-js/google/protobuf/wrappers.bin
index 5b4b186..7f34d6c 100644
--- a/integration/grpc-js/google/protobuf/wrappers.bin
+++ b/integration/grpc-js/google/protobuf/wrappers.bin
@@ -1,5 +1,5 @@
 
-^^google/protobuf/wrappers.proto^Z^H^H^C^P^N^X^@"^@z<AF>!
+^^google/protobuf/wrappers.proto^Z^H^H^C^P^S^X^A"^@z<AF>!
 ^^google/protobuf/wrappers.proto^R^Ogoogle.protobuf"#
 ^KDoubleValue^R^T
 ^Evalue^X^A ^A(^AR^Evalue""
```


src/generate-services.ts
```diff
@@ -106,18 +106,22 @@ function generateRegularRpcMethod(
...
   const outputType = responseType(ctx, methodDesc);
+  const rawOutputType = responseType(ctx, methodDesc, { keepValueType: true });
 
   const params = [...(options.context ? [code`ctx: Context`] : []), code`request: ${inputType}`];
   const maybeCtx = options.context ? 'ctx,' : '';
 
   let encode = code`${rawInputType}.encode(request).finish()`;
   let decode = code`data => ${outputType}.decode(new ${Reader}(data))`;
-
+  console.error();
```
stephenh, line 121: This snuck in :-)

stephenh (approved):
@fizx sorry for the delayed review, but I think this lgtm.

Yeah, figuring up the `.bin` files would be great; fwiw let me know if that's just a pita and I can also locally checkout/copy this branch, fix up those bins, and then merge on my side of things, if that's more pragmatic than figuring out root case on that for now.

Thanks!

fizx:
Now that we're trued up in #546, this looks more reasonable.  Going to land after the status checks pass.

stephenh:
Great, thanks for pushing this threw @fizx !

PR #535, opened 2022-03-15
fix(Grpc-Web): Fix compilation failure when a service definition contains a client streaming call.

zakhenry:
Note this does not implement client streaming, instead the function will throw due to the feature not being implemented. This is a stop-gap until true client streaming support is added.

This PR relates to #329 (the original issue), and partially overlaps with #335 (a WIP PR that actually implements client streaming, but has stalled)


src/generate-services.ts
```diff
@@ -51,12 +51,10 @@ export function generateService(
...
     }
 
-    let inputType = requestType(ctx, methodDesc);
     // the grpc-web clients auto-`fromPartial` the input before handing off to grpc-web's
     // serde runtime, so it's okay to accept partial results from the client
-    if (options.outputClientImpl === 'grpc-web') {
-      inputType = code`${utils.DeepPartial}<${inputType}>`;
-    }
+    const partialInput = options.outputClientImpl === 'grpc-web'
+    const inputType = requestType(ctx, methodDesc, partialInput);
```
stephenh, line 57: Cool, nice clean up!

src/generate-grpc-web.ts
```diff
@@ -35,33 +35,46 @@ export function generateGrpcClientImpl(
...
 }
 
 /** Creates the RPC methods that client code actually calls. */
 function generateRpcMethod(ctx: Context, serviceDesc: ServiceDescriptorProto, methodDesc: MethodDescriptorProto) {
   assertInstanceOf(methodDesc, FormattedMethodDescriptor);
-  const { options, utils } = ctx;
-  const inputType = requestType(ctx, methodDesc);
-  const partialInputType = code`${utils.DeepPartial}<${inputType}>`;
+  const requestMessage = rawRequestType(ctx, methodDesc)
+  const inputType = requestType(ctx, methodDesc, true);
```
stephenh, line 53: Instead of invoking both `rawRequestType` + `requestType`, I wonder if `requestType` would return a tuple or something...

Just thinking out loud, b/c we also have an open issue where someone's input type is not actually a message, so we need to make the `fromPartial` call optional, and maybe that'd be nice to handle inside of `requestType` itself... :shrug:

integration/grpc-web/example.ts
```diff
@@ -613,6 +619,13 @@ export class DashStateClientImpl implements DashState {
   ActiveUserSettingsStream(request: DeepPartial<Empty>, metadata?: grpc.Metadata): Observable<DashUserSettingsState> {
     return this.rpc.invoke(DashStateActiveUserSettingsStreamDesc, Empty.fromPartial(request), metadata);
   }
+
+  ChangeUserSettingsStream(
+    request: Observable<DeepPartial<DashUserSettingsState>>,
+    metadata?: grpc.Metadata
+  ): Observable<DashUserSettingsState> {
+    throw new Error('ts-node does not yet support client streaming!');
```
stephenh, line 627: Nit `ts-proto` :-D

stephenh (approved):
@zakhenry thanks for the PR! I agree this is a good baby step / mitigation until we have full support. I just had one comment about ts-node/ts-proto typo, but otherwise looks great! Ping when that's updated and I'll hit merge.

Thanks!

integration/grpc-web/example.ts
zakhenry: oh whoops! fixed.

src/generate-grpc-web.ts
zakhenry: Yea I think there is a bit of a code smell that the requestType function manages the wrapping of the type into partial, but it was a minimal change. I suspect it might need a bit of a rethink for that other issue

stephenh:
Thanks @zakhenry !

PR #544, opened 2022-04-04
fix: Use Uint8Array.forEach in base64FromBytes

douglasc-nflx:
Otherwise, TypeScript 4.6.2 complains:

TS2569: Type 'Uint8Array' is not an array type or a string type. Use compiler option '--downlevelIteration' to allow iterating of iterators.
    184 | function base64FromBytes(arr: Uint8Array): string {
    185 |   const bin: string[] = [];
  > 186 |   for (const byte of arr) {
        |                      ^^^

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -59,9 +59,9 @@ const btoa: (bin: string) => string =
   globalThis.btoa || ((bin) => globalThis.Buffer.from(bin, 'binary').toString('base64'));
 function base64FromBytes(arr: Uint8Array): string {
   const bin: string[] = [];
-  for (const byte of arr) {
+  arr.forEach((byte) => {
     bin.push(String.fromCharCode(byte));
-  }
+  });
   return btoa(bin.join(''));
 }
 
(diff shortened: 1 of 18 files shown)


stephenh:
@douglasc-nflx nice, thanks for the PR! Can you run `yarn bin2ts` to have the `integration-test` files updated, which serve as our test suite?

douglasc-nflx:
Ran `yarn bin2ts` and updated integration test files.

stephenh:
Looks great, thanks!

coyotte508:
Hello,

This PR decreases the performance of `base64FromBytes` on Node.JS.

Here's a benchmark: https://gist.github.com/coyotte508/a66c1c4f26e137d1f9baab78b8185bed

```
loop: 16.063s
forEach: 25.794s
buffer: 2.545s
```

The use of `forEach` takes 56% more time.

Using directly `Buffer.from` on the `Uint8Array`, on Node.JS, increases performance 10x from the current implem. Maybe we could avoid the whole `forEach` / `for` path when on Node.JS?

stephenh:
@coyotte508 ah wow, yeah that's a huge difference. We have a `env=node|browser|both` flag, if you'd like to submit a PR to use `Buffer.from` that'd be great!

PR #546, opened 2022-04-06
fix: My local output from running yarn build:test

fizx:
I just ran:

```
:$ yarn
:$ yarn build:local
```

on the main branch.  Intel mac.  I'm not sure why there's a diff, but there shouldn't be!


fizx:
Seems like this is the correct output.  CI doesn't hate it either.  Now I'm curious how previous CI runs were passing.

stephenh:
@fizx hrm, yeah, I think you're right; not sure what happened, but we do have a known issue where running `proto2bin` with different versions of `protoc` will produce different `*.bin` files.

The non-`:local` all run within a docker image that's supposed to mitigate this, but either there is still some drift in that, or more likely the last person that touched these files (could have been me, no idea) used `proto2bin:local` instead of `proto2bin`.

Not going to sink any time to figuring out why/what happened, unless it becomes a common occurrence (it was definitely much more common b/c the docker-based commands), but appreciate the PR to true them up.

stephenh (approved):

Issue #548, opened 2022-04-13
useOptionals for oneofs fields


Larkooo:
It is quite annoying and tedious that a oneofs fields do not default to optionals because it triggers typescripts type safety. Thus, it would be great if optionals were used by default for a oneofs fields without needing to enable the useOptionals option, as it also allows other fields to be optional. 

We could perhaps create another option for allowing a oneofs fields to be optional or just check if the property is within a oneof and directly make it optional
https://github.com/stephenh/ts-proto/blob/main/src/main.ts
https://github.com/stephenh/ts-proto/blob/b90821d05539848cdda7ab2b59daa1313a69a123/src/types.ts#L340

stephenh:
Hm, yeah, @Larkooo I'd generally recommend using the `oneof=unions` option, which should probably be the default, but in the case of `oneof=unions` not being in use, having optional keys for oneof properties seems reasonable. If you want to work on a PR, that'd be great.

PR #549, opened 2022-04-15
feat(plugin): optional fields for oneof properties

Larkooo:
addresses issue #548


integration/meta-typings/google/protobuf/wrappers.ts
```diff
@@ -655,134 +655,134 @@ export const protoMetadata: ProtoMetadata = {
           path: [4, 0],
           span: [50, 0, 53, 1],
           leadingComments:
-            ' Wrapper message for `double`.\n\n The JSON representation for `DoubleValue` is JSON number.\n',
+            ' Wrapper message for `double`.\r\n\r\n The JSON representation for `DoubleValue` is JSON number.\r\n',
```
stephenh, line 658: @Larkooo FWIW these changes (and the `*.bin` file changes) might be from running `proto2bin:local` instead of the docker-based `proto2bin`? 

We used to have a lot of churn on these files from contributors having different versions of `protoc` installed on their machines, but in theory the docker-based setup that @boukeversteegh put in place _should_ have removed this churn.

Can you try the docker-based commands and see if it resolves these differences? Or were you already doing that, and we still have some non-determinism issues?

stephenh (approved):
Great, thanks for the PR @Larkooo !

yarn-error.log
```diff
@@ -0,0 +1,6831 @@
+Arguments:
```
boukeversteegh, line 1: Seems that this file got included in the commit accidentally

boukeversteegh:
I think it might be worthwhile to add a simple test that only instantiates an object of a type with one-of fields, and omits a field.

If we ever break this feature, the TS file won't compile.

for example, in [oneof-properties-tests](https://github.com/stephenh/ts-proto/blob/8769ef2ad01d53edc2d373201afeab8ca7403885/integration/oneof-properties/oneof-properties-test.ts):

```ts

  it('allows omitting fields', () => {
      let msg: PleaseChoose = {
         name: 'foo',
      }
  });

```

SteveVaknin (approved):

stephenh:
@scott-lc yeah, would be great if you wanted to pick this up; just scanning the PR, I think the issues would be to:

a) See the PR comment about running `yarn proto2bin` to hopefully remove all of the `*.bin` files as changed files in the PR

b) Remove the `yarn-error.log` from the PR

c) Rebase the PR onto the latest `main`

d) Update one of the `integration-tests/*` projects to have a schema that reproduces the issue and has a test case that shows the codegen output continues to work as you expect; i.e. @boukeversteegh had suggested this has a starting place:

https://github.com/stephenh/ts-proto/blob/8769ef2ad01d53edc2d373201afeab8ca7403885/integration/oneof-properties/oneof-properties-test.ts

Thanks!

PR #551, opened 2022-04-21
chore: OS X requires coreutils to be installed

YeeJone:
Some shell scripts use the realpath command, which will report an error...


stephenh (approved):
Thanks!

PR #552, opened 2022-04-21
feat: include service and definition types with implementations

niedzielski:
When exporting a service or service definition, also export it as a
type. Previously, you would have to import the value then `typeof` off
of it which is cumbersome. Automate this as it has no runtime overhead
and seems more intuitive.

Before:

```ts
export const TestDefinition = {
  name: 'Test',
  fullName: 'simple.Test',
  methods: {
    …
  },
} as const;
```

After:

```ts
export type TestDefinition = typeof TestDefinition;
export const TestDefinition = {
  name: 'Test',
  fullName: 'simple.Test',
  methods: {
    …
  },
} as const;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,32 @@
+## [next](https://github.com/stephenh/ts-proto/compare/v1.110.4...main) (????-??-??)
+
+### Features
+
+* When outputing service and service definition implementations, include types. Eg, before:
+
+  ```ts
+  export const TestDefinition = {
+    name: 'Test',
+    fullName: 'simple.Test',
+    methods: {
+      …
+    },
+  } as const;
+  ```
+
+  Now:
+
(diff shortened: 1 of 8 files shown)


fizx:
Hi @stephenh, 
Code looks good here, and it's a useful feature, but I was holding off on merging it because it's from someone on my work team, and we haven't chatted yet about whether you feel that represents a conflict of interest on my review.

Thanks!

fizx (commented):
:shipit:

stephenh (approved):
Nice! Thanks @niedzielski !

Sorry for the really late review here; have just been busy. Fwiw @fizx yeah I'm perfectly happy with you + other maintainers (cc @boukeversteegh who's been actively lately as well) reviewing and merging PRs, to keep me from being the bottleneck on the project.

PR #555, opened 2022-04-24
feat: add support for generating `nice-grpc` server and client stubs

### LINKED ISSUE #545: Feature request: First class support for nice-grpc
Ts-proto currently has the "generic" service output target that creates a data representation of a service that can be used both at runtime, and to infer Typescript type signatures of services and clients.

With the magic of Typescript, this works with respect to type checking. An implementation of a service in nice-grpc that implements `ServiceImplementation<typeof SomeServiceDefinition>` will be appropriately typed and any mismatch in signature will be flagged.

It is not however that IDE friently. Specifically due to the inferred type signatures, there's no clean way to simply use "implement interface" or similar refactoring tools in IntelliJ or other IDEs. It is possible, but the signature it generates is far from readable.

To support nice-grpc, it would be extremely helpful if a regular Typescript interface could be generated, matching the signature of ServiceImplementation<typeof SomeServiceDefinition> and extending from it, so that an implementing class could simply provide the methods of this regular interface and typing errors would be much more readable.

aikoven:
This PR adds support for generating client and server interfaces for [nice-grpc](https://github.com/deeplay-io/nice-grpc). 

Currently `nice-grpc` generates these interfaces at compile time from generic definitions using TypeScript mapped types. This approach has drawbacks, mainly in developer experience, making type errors and editor hints less readable.

Closes #545


README.markdown
```diff
@@ -362,6 +362,8 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=outputServices=generic-definitions`, ts-proto will output generic (framework-agnostic) service definitions. These definitions contain descriptors for each method with links to request and response types, which allows to generate server and client stubs at runtime, and also generate strong types for them at compile time. An example of a library that uses this approach is [nice-grpc](https://github.com/deeplay-io/nice-grpc).
 
+- With `--ts_proto_opt=outputServices=nice-grpc`, ts-proto will output server and client stubs for [nice-grpc](https://github.com/deeplay-io/nice-grpc). This should be used together with generic definitions, i.e. you should specify two options: `outputServices=nice-grpc,outputServices=generic-definitions`.
```
aikoven, line 365: Not sure about the latter, maybe it would be better to generate generic definitions implicitly on `outputServices=nice-grpc`.

src/main.ts
```diff
@@ -313,6 +317,7 @@ export function makeUtils(options: Options): Utils {
     ...makeTimestampMethods(options, longs),
     ...longs,
     ...makeComparisonUtils(),
+    ...makeNiceGrpcServerStreamingMethodResult(),
```
aikoven, line 320: I needed to introduce a type that would be conditionally rendered once per file only when used. Could not find a better way than adding it to the global `Utils` object.

stephenh (approved):
Nice! Looks great; FWIW @aikoven anything that you want to ship in ts-proto to make it "more out of the box" for nice-grpc and/or just in general is great imo.

Also feel free to ping another maintainer for a +1 & merge without me, if I take too long to get to a PR.

aikoven:
It seems to have conflicted with #552, the build is failing after merge. Will make a fix tomorrow.

PR #556, opened 2022-04-28
docs: In README change copy-paste mistake of primitives to messages

jonaskello:
I think it was the intention to say messages rather than primtives here.


stephenh (approved):
Thanks!

PR #562, opened 2022-05-05
fix: use Long.fromValue instead of Long.fromString

xalanq:
#405


boukeversteegh:
Looks like this was about applying the same intended changes to a previously overlooked spot. I'll merge it as tests pass. Thank you!

PR #565, opened 2022-05-15
fix: enum type encodes to JSON in number when stringEnums=false

xalanq:
when add an option `stringEnums=false`, the enum type decoding and encoding behave like

```
xxxFromJson: number -> enum
xxxToJson: enum -> string
```

which should be

```
xxxFromJson: number -> enum
xxxToJson: enum -> number
```

### DIFF
--- a/integration/angular/package-lock.json
+++ b/integration/angular/package-lock.json
@@ -7488,9 +7488,9 @@
       }
     },
     "minimist": {
-      "version": "1.2.5",
-      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.5.tgz",
-      "integrity": "sha512-FM9nNUYrRBAELZQT3xeZQ7fmMOBg6nWNmJKTcgsJeaLstP/UODVpGsr5OhXhhXg6f+qtJ8uiZ+PUxkDWcgIXLw==",
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.6.tgz",
+      "integrity": "sha512-Jsjnk4bw3YJqYzbdyBiNsPWHPfO++UGG749Cxs6peCu5Xg4nrena6OVxOYxrQTqww0Jmwt+Ref8rggumkTLz9Q==",
       "dev": true
     },
     "minipass": {


xalanq:
> @xalanq thanks for the PR! Looking at the build, it looks like maybe prettier is failing and then also if you could rerun the codegen setup (see the Development section https://github.com/stephenh/ts-proto#development) and add those changed files to the PR, that'd be great!

@stephenh done

xalanq:
> @xalanq thanks for the PR! Looking at the build, it looks like maybe prettier is failing and then also if you could rerun the codegen setup (see the Development section https://github.com/stephenh/ts-proto#development) and add those changed files to the PR, that'd be great!

@stephenh done! BTW, I just realized that there were a lot of test cases need to be updated

stephenh:
Oh right! Seeing how many files this updated reminded me of why we'd done this...

I think my/our previous reading of the proto3 json spec:

https://developers.google.com/protocol-buffers/docs/proto3#json

Was that enums should always be output as strings, and then yeah conditionally allow either strings or numbers as input.

Granted, I agree "parsers should accept either ints or strings" insinuates that outputting ints is fine, but that was our historical rationale for why we still output strings in the JSON...

Given that recollection, I'm worried that this is a large breaking change, so we might need to make our `stringEnum` option more flexible/fine-grained like `stringEnums=false` could be what it is today, but some sort of new `stringEnums=false-and-json` would trigger the "numbers in JSON" output behavior you're looking for...

So, dunno, sorry to only realize this now, but I think this is safest for ts-proto's existing users, i.e.:

* Update `options.ts` to allow a new `stringEnums=false-and-json` value (very open to suggestions on a better name, or a better way of doing this)
* Add a new/minimalist `integration/string-enums-and-json/` example with a `parameters.txt` that uses the new option

Does that sound okay/wdyt?

(Also fwiw the `.bin` files should generally not change, can you use the docker flow to generate them? That should eliminate any cross-system / cross-native-`protoc` differences in them. Let us know if it doesn't though, but that's the theory...)

Thanks!

xalanq:
Agree with u!
According to the recommended standard https://developers.google.com/protocol-buffers/docs/proto3#json, it should not convert enum to number.
It was my mistake, I didn't know that before.
Thx for your correction!

PR #566, opened 2022-05-16
fix: enum type returns 'UNRECOGNIZED' or '-1' in xxxToJSON/xxxToNumber

xalanq:
Is returning 'unknown' or '0' in xxxToJSON/xxxToNumber as expected? It may conflict with the default enum value '0'

I'm not sure, so I correct it to 'UNRECOGNIZED' and '-1'

wdyt?

### DIFF
--- a/integration/avoid-import-conflicts/simple.ts
+++ b/integration/avoid-import-conflicts/simple.ts
@@ -43,8 +43,9 @@ export function simpleEnumToJSON(object: SimpleEnum): string {
       return 'LOCAL_FOO';
     case SimpleEnum.LOCAL_BAR:
       return 'LOCAL_BAR';
+    case SimpleEnum.UNRECOGNIZED:
     default:
-      return 'UNKNOWN';
+      return 'UNRECOGNIZED';
   }
 }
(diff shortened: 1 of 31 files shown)


stephenh (approved):
Yeah, @xalanq I think this is a good change!

My vague recollection is that the `return UNKNOWN` / `return 0` code is from a long time ago and just never got updated to know about the unrecognizedEnum setting.

Reading the proto3 docs, I think technically there are use cases where ideally ts-proto / protobuf libraries would leave the unrecognized enum in place, like if a service accepts a message with `Color.BLUE`, and the service itself is on an older version of the schema w/o `BLUE`, then it's internal code would see `UNRECOGNIZED` ... _but_ if it didn't mutate the message, and just re-serialized to some other client (like a database), then `BLUE` would still be in the serialized JSON.

This is a somewhat niche use-case though, and there would be a lot more work, not just with enums, but all fields iirc, to really support it, so just mentioning it out of thoroughness, but otherwise I think this approach is good and definitely better than what we had been doing.

Thanks!

Issue #69, opened 2020-05-22
Incompatible Wrapper Types with Nest.js (protobufjs)


ssilve1989:
When using [Google's Well-Known Types](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf) with Nest (or just plain protobufjs) the types generated from ts-proto fail to be parsed.

Given this protobuf
```proto
syntax = "proto2"
import "google/protobuf/wrappers.proto";

message MonitorHelloRequest {
    optional google.protobuf.StringValue name_prefix = 2;
    optional google.protobuf.Int64Value limit = 3;
}
```

we get a typescript interface of 
```typescript

export interface MonitorHelloRequest {
  namePrefix: string | undefined;
  limit: number | undefined;
}
``` 
which seems fine. But when attempting to use that as a message that is parsed by a server using protobufjs (nest in this case)

You get the error 
```
TypeError: MonitorHelloRequest.namePrefix: object expected
```

This is because protobufjs is actually expecting that interface to be 
```typescript

export interface MonitorHelloRequest {
  namePrefix: { value: string } | undefined
  limit: { value: number } | undefined;
}
``` 
it is unclear to me who is right here as I can see the argument for either. This was referenced on https://github.com/protobufjs/protobuf.js/issues/1042 and several other issues but there does not seem to be any update in a **very long time** from protobufjs on the matter

ssilve1989:
@stephenh I've opened up a reproduction repo here: https://github.com/ssilve1989/ts-proto-issue-69

stephenh:
Ah shoot. Yeah...

So, one of my pet goals with `ts-proto` was to get some resemblance of optional fields back into protobuf.

I did this by treating proto primitives as required (b/c they come back with default values if unset) and then the well-known types I purposefully make into `string | unknown`.

Then in the `Message.encode/decode` methods that `ts-proto` controls, I can turn these `string | unknown` types into the `{ value: string }` wrappers that protobuf wants on the wire.

However, you're using NestJS's protobuf support to do the actual serde, which doesn't know about my "haha well-known types are my hack to get optional fields back into protobuf".

So, not sure the best solution here. You either need to:

1) Teach ts-proto to _not_ output well-known types as the `string | optional`, which is "more correct", but also a generally worse API,
2) Teach NestJS's protobuf serializer about ts-proto's "different" approach to well-known types

[...]

PR #567, opened 2022-05-17
feat: added nestJsTimestampTypeWrapper

PhilipMantrov:
related issue #69


PhilipMantrov:
Added integration test and change 
`const wrappers = imp('wrappers@protobufjs/minimal');`
to
`const wrappers = imp('wrapper@protobufjs');`
because protobuf.js has custom wrapper only for Any ([wrappers.js](https://github.com/protobufjs/protobuf.js/blob/master/src/wrappers.js), and [issue](https://github.com/protobufjs/protobuf.js/pull/1258)), and if we dont have Any in our proto, wrappers has undefined

PhilipMantrov:
And I commit changes in genfiles, if it dont needed just revert this commit.

PhilipMantrov:
@stephenh can we merge this pr asap? Or just tell me what we need this for accepting this pr, thx. :)

integration/angular/simple-message.ts
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable */
+import { util, configure, Writer, Reader } from 'protobufjs/minimal';
```
stephenh, line 2: @PhilipMantrov do you know all of these other files changed?

stephenh:
@PhilipMantrov thanks for adding the test! I'm good to merge this once the build is passing.

integration/angular/simple-message.ts
PhilipMantrov: Newest version of protobufjs has deep types, so compiler dont need abstraction as _m0 anymore. 
In package.json we have '^', so we install newest version 6.11.3 (on 6.8.8 how i see compiler use _m0 for abstraction).

PhilipMantrov:
> @PhilipMantrov thanks for adding the test! I'm good to merge this once the build is passing. 

So you need to disable 'Diff check' for once because protobufjs was updated.

integration/angular/simple-message.ts
stephenh: Huh! I'd love to get rid of this, but I'm not following why this would change, b/c:

1) Isn't yarn.lock pinned to 6.8.8?
2) What controls the `_m0` output is this line:

https://github.com/stephenh/ts-proto/blob/main/src/options.ts#L183

I.e. the `* as _m0` is generated by ts-poet because we explicitly tell it to, and I'm like 99% sure nothing in `node_modules/protobuf/...` would affect ts-poet's behavior as to whether it does/does not use the `_m0` style import...

So, dunno, I'm still not following how this changed?

stephenh: @PhilipMantrov I checked out this PR locally, and when I run `yarn bin2ts`, AFAICT it puts all of the `_m0` back... I'm wondering if maybe when you ran codegen, you weren't on the latest `main` and/or had an outdated version of `ts-poet` installed?

Fwiw I'd be really happy to bump protobufjs and clean up these imports, but seems like something to do in a separate PR.

Can you try a new `yarn bin2ts` and see what happens? Thanks!

stephenh: @PhilipMantrov I got your ping about asking for a re-review, see this comment ^.

If we just can't figure this out, I can always create a new PR with your core code changes, and w/o all of the `_m0` changes, and get that merged.

PhilipMantrov: protobufjs now fixed on 6.8.8

n0isy:
Any updates?

integration/angular/simple-message.ts
PhilipMantrov: >I'd be really happy to bump protobufjs and clean up these imports, but seems like something to do in a separate PR  
I create another PR for this sooner.

flametuner:
I tried to run the assignment of ".google.protobuf.Timestamp" to `wrappers` with protobufjs 7.0.0 and it isn't working, I had to downgrade to 6.11.3.

It's worth to check if it's running in version 7 and what has changed.

PhilipMantrov:
@flametuner yea, its not working on `protobufjs: 7.0.0` and above. 
`ts-proto` was pinned `protobufjs` to `6.8.8` for some reason, and for `protobufjs 7.0.0` need some work in `ts-proto`.

TakeOver:
Whatsup guys when this will be merged? I need this feature ASAP

TakeOver:
Is this repo alive? Seems like dead perhaps I should fork it

TannerPerrien:
@TakeOver and @stephenh - Thanks for this contribution!

When generating output with `--ts_proto_opt=useDate=date`, I see the resulting protobufjs `wrappers/fromObject/toObject`. However, my nestjs application does not end up invoking this wrapper and my resulting `Timestamp` output is

```js
{
  "seconds": "0",
  "nanos": 0
}
```

I have carefully reviewed the [nestjs-simple-usedate](https://github.com/stephenh/ts-proto/tree/main/integration/nestjs-simple-usedate) integration test.

PR #572, opened 2022-05-26
fix: Use `fromJSON` methods to deserialize enum map values

haines:
Our grpc-gateway service returns enums as strings in JSON, but we want to convert those to numeric enums in the client. It works fine for enum fields in messages (which use the enum's `fromJSON` method), but not for enum values in maps (which currently just generate a type cast).

This PR changes the generator to use the enum's `fromJSON` method instead of a type cast for map values.


stephenh (approved):
Ah yeah, that makes sense, thanks @haines !

stephenh:
@haines huh, I just noticed it looks like the last commit on main wasn't fully rebased, and so it's codegen output isn't 100% correct b/c of a separate change made while it was in-flight.

Looks like you've noticed the same and already fixed it as well! :-) Thanks!

PR #583, opened 2022-06-02
Fix and simplified option handling

oyvindwe:
- Moved logic for required and implied options into options.ts
- Made onlyTypes=true behave as documented.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -396,7 +396,7 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=onlyTypes=true`, only types will be emitted, and imports for `long` and `protobufjs/minimal` will be excluded.
 
-  Note: _This is a combination_ of `outputJsonMethods=false,outputEncodeMethods=false,outputClientImpl=false,nestJs=false`
+  This implies `outputJsonMethods=false,outputEncodeMethods=false,outputClientImpl=false,nestJs=false`
 
 - With `--ts_proto_opt=usePrototypeForDefaults=true`, the generated code will wrap new objects with `Object.create`.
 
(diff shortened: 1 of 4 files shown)


src/options.ts
```diff
@@ -124,8 +123,18 @@ export function optionsFromParameter(parameter: string | undefined): Options {
     }
     Object.assign(options, parsed);
   }
-  // We should promote onlyTypes to its own documented flag, but just an alias for now
-  if (!options.outputJsonMethods && !options.outputEncodeMethods && !options.outputClientImpl && !options.nestJs) {
+  // onlyTypes=true implies outputJsonMethods=false,outputEncodeMethods=false,outputClientImpl=false,nestJs=false
+  if (options.onlyTypes) {
```
stephenh, line 127: ...huh, yeah, not sure why we weren't doing this before. Great fix!

If you want to add test cases to `options-test.ts` to make sure these don't regress, that'd be great!

stephenh (approved):

src/options.ts
oyvindwe: This should probably have been fixed when the option was documented. :)

I'll add test cases!

README.markdown
```diff
@@ -396,7 +396,7 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=onlyTypes=true`, only types will be emitted, and imports for `long` and `protobufjs/minimal` will be excluded.
 
-  Note: _This is a combination_ of `outputJsonMethods=false,outputEncodeMethods=false,outputClientImpl=false,nestJs=false`
+  This is the same as setting `outputJsonMethods=false,outputEncodeMethods=false,outputClientImpl=false,nestJs=false`
```
oyvindwe, line 399: Not sure if `nestJs=false` should be included here, as it is false by default, but  nestJs=true -> onlyTypes=false.

src/options.ts
oyvindwe: Tests added.

stephenh:
Looks great @oyvindwe , thanks!

Issue #330, opened 2021-06-27
import Long from 'long'; even when long option is number


doctorpangloss:
Importing long is emitted even though the long format is number.

Version: 1.81.3
options:
```
        option "outputClientImpl=grpc-web"
        option "esModuleInterop=true"
        option "returnObservable=false"
        option "addGrpcMetadata=true"
        option "forceLong=number"
        option "oneof=unions"
        option "outputJsonMethods=false"
        option "lowerCaseServiceMethods=true"
```

stephenh:
Yeah, that's surprising, can you paste a snippet of the code that has `Long` in it?

jake-tourtech:
Here's an outputted file including `Long` despite the inclusion of `--ts_proto_opt=forceLong=number`. All it appears to do it unnecessarily configure `protobufjs`.

```typescript
/* eslint-disable */
import { GrpcMethod, GrpcStreamMethod } from '@nestjs/microservices';
import Long from 'long';
import _m0 from 'protobufjs/minimal';
import { Observable } from 'rxjs';
import { Metadata } from '@grpc/grpc-js';

export const protobufPackage = 'hero';

export interface HeroById {
  id: number;
}

export interface Hero {
  id: number;
  name: string;
}

export const HERO_PACKAGE_NAME = 'hero';

export interface HeroesServiceClient {
  findOne(
    request: HeroById,
    metadata: Metadata,
    ...rest: any
  ): Observable<Hero>;
}

export interface HeroesServiceController {
  findOne(
    request: HeroById,
    metadata: Metadata,
    ...rest: any
  ): Observable<Hero>;
}

export function HeroesServiceControllerMethods() {
  return function (constructor: Function) {
    const grpcMethods: string[] = ['findOne'];
    for (const method of grpcMethods) {
      const descriptor: any = Reflect.getOwnPropertyDescriptor(
        constructor.prototype,
        method,
      );
      GrpcMethod('HeroesService', method)(
        constructor.prototype[method],
        method,
        descriptor,
      );
    }
    const grpcStreamMethods: string[] = [];
    for (const method of grpcStreamMethods) {
      const descriptor: any = Reflect.getOwnPropertyDescriptor(
        constructor.prototype,
        method,
      );
      GrpcStreamMethod('HeroesService', method)(
        constructor.prototype[method],
        method,
        descriptor,
      );
    }
  };
}

export const HEROES_SERVICE_NAME = 'HeroesService';

if (_m0.util.Long !== Long) {
[...]

PR #599, opened 2022-06-20
remove Long when unused

3846masa:
Somethimes (e.g. when outputEncodeMethods=false), Long is not required, but included in generated codes.

This PR ensures that unused Long is not included in generated codes.

Releated: https://github.com/stephenh/ts-proto/issues/330


src/main.ts
```diff
@@ -345,25 +343,24 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
...
     : `
     // If you get a compile-error about 'Constructor<Long> and ... have no overlap',
     // add '--ts_proto_opt=esModuleInterop=true' as a flag when calling 'protoc'.`;
 
-  // Kinda hacky, but we always init long unless in onlyTypes mode. I'd rather do
-  // this more implicitly, like if `Long@long` is imported or something like that.
-  const longInit = options.onlyTypes
-    ? code``
-    : code`
+  const Long = conditionalOutput(
```
stephenh, line 354: Ah neat! This is smart. @3846masa do you mind adding an in-source comment right before this:

```
// Instead of exposing `LongImp` directly, let callers think that they are getting the
// `imp(Long)` but really it is that + our long initialization snippet. This means the
// initialization code will only be emitted in files that actually use the Long import.
```

stephenh (approved):
Nice!

src/main.ts
3846masa: Looks good. I'll add it :+1:

3846masa:
@stephenh 
Thanks for the review. I added a comment to the code.

stephenh:
Looks great, thanks @3846masa !

PR #602, opened 2022-06-29
fix: add .js suffix to proto cross-reference imports

### LINKED ISSUE #601: Add .js suffix to imports
When importing a cross-referenced file, a line like this is generated:

```
import { MyMessage } from '../myother/myother'
```

.. assuming there's a proto at ../myother/myother.proto

But if we add the suffix '.pb.ts' to the generated files:

```
import { MyMessage } from '../myother/myother.pb'
```

Is not recognized as a TypeScript import by tsc because of the .pb suffix.

To fix this, we can just add .js, and the TypeScript compiler recognizes that we actually mean the .ts file:

```
import { MyMessage } from '../myother/myother.pb.js'
```

And this is also required if we are using esm mode.

The fix:

```
diff --git a/src/utils.ts b/src/utils.ts
index 60efc9e..a666e04 100644
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -203,6 +203,6 @@ export function impProto(options: Options, module: string, type: string): Import
   if (options.onlyTypes) {
     return imp('t:' + importString);
   } else {
-    return imp(importString);
+    return imp(importString + '.js');
   }
 }
```

paralin:
When importing a cross-referenced file, a line like this is generated:

import { MyMessage } from '../myother/myother'

.. assuming there's a proto at ../myother/myother.proto

But if we add the suffix '.pb.ts' to the generated files:

import { MyMessage } from '../myother/myother.pb'

Is not recognized as a TypeScript import by tsc because of the .pb suffix.

To fix this, we can just add .js, and the TypeScript compiler recognizes that we actually mean the .ts file:

import { MyMessage } from '../myother/myother.pb.js'

Fixes #601

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -199,7 +199,8 @@ export function getPropertyAccessor(objectName: string, propertyName: string, op
 }
 
 export function impProto(options: Options, module: string, type: string): Import {
-  const importString = `${type}@./${module}${options.fileSuffix}`;
+  // NOTE: TypeScript resolves importing foo.js to importing foo.ts.
+  const importString = `${type}@./${module}${options.fileSuffix}.js`;
   if (options.onlyTypes) {
     return imp('t:' + importString);
   } else {
--- a/integration/avoid-import-conflicts-types-only/simple.ts
+++ b/integration/avoid-import-conflicts-types-only/simple.ts
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import type { SimpleEnum as SimpleEnum1, Simple as Simple2 } from './simple2';
+import type { SimpleEnum as SimpleEnum1, Simple as Simple2 } from './simple2.js';
 
 export const protobufPackage = 'simple';
 
--- a/integration/avoid-import-conflicts/simple.ts
+++ b/integration/avoid-import-conflicts/simple.ts
@@ -4,7 +4,7 @@ import {
   Simple as Simple2,
   simpleEnumFromJSON as simpleEnumFromJSON3,
   simpleEnumToJSON as simpleEnumToJSON4,
-} from './simple2';
+} from './simple2.js';
 import * as _m0 from 'protobufjs/minimal';
 
 export const protobufPackage = 'simple';
--- a/integration/barrel-imports/foo.ts
+++ b/integration/barrel-imports/foo.ts
@@ -1,5 +1,5 @@
 /* eslint-disable */
-import { Bar } from './bar';
+import { Bar } from './bar.js';
 import * as _m0 from 'protobufjs/minimal';
 
 export interface Foo {
--- a/integration/bytes-node/point.ts
+++ b/integration/bytes-node/point.ts
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import * as _m0 from 'protobufjs/minimal';
-import { BytesValue } from './google/protobuf/wrappers';
+import { BytesValue } from './google/protobuf/wrappers.js';
 
 export const protobufPackage = '';
 
--- a/integration/fieldmask/fieldmask.ts
+++ b/integration/fieldmask/fieldmask.ts
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import * as _m0 from 'protobufjs/minimal';
-import { FieldMask } from './google/protobuf/field_mask';
+import { FieldMask } from './google/protobuf/field_mask.js';
 
 export const protobufPackage = '';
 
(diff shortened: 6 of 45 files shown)


src/utils.ts
```diff
@@ -203,6 +203,6 @@ export function impProto(options: Options, module: string, type: string): Import
   if (options.onlyTypes) {
     return imp('t:' + importString);
   } else {
-    return imp(importString);
+    return imp(importString + '.js');
```
stephenh, line 206: Huh! I didn't know that if you have a file called `foo.ts`, you're allowed to import it as `foo.js`.

That makes sense, I guess, since it will all wash out to `*.js` files in the end / post-compilation.

I dunno, that said, does it work to make this `.ts`? 

Personally I think that'd be less surprising to the reader, since they would normally be reading the files as pre-`tsc` TS code. And I assume `tsc` will make all of the `foo.ts` imports into `foo.js` in the output anyway (?).

paralin: > does it work to make this .ts?

No, this returns an error from the compiler (counter-intuitively)

stephenh: Also huh!

Wdyt about doing a `s,\./_/` on the file names, so that your `myother.pb.ts` files end up as `myother_bs` and imports like like:

```
import { MyMessage } from '../myother/myother_pb'
```

It seems like this would be less disruptive to existing users that aren't use the `.pb` naming convention.

Granted, it sounds like file extensions are becoming defacto required in ESM, but hopefully/surely they've fixed/allow `.ts` extensions by then.

paralin: I'm very interested in using .pb.js since I'm using .pb.go - and with this patch it's working correctly.

We're going to have to use the .js suffix for all non-package imports with ESM anyway - (for example protobufjs/minimal.js) so eventually this will be required. And already the TypeScript compiler ignores the .js suffix and finds the .ts file, so it shouldn't be backwards incompatible, right?

If it causes an issue I guess it could be another option (to add .js as the suffix everywhere we try to import a ts file).

stephenh: Yeah, totally get/agree we'll need it with ESM, albeit I'm ~80% sure that `*.ts` suffixes will be/are kosher with that tsconfig (I've only seen the TS release notes go by, haven't personally used it yet).

Maybe do a conditional here, like if `importString` has a `.` in it, add the `.js` export?

I'm just thinking, while this makes a ton of sense for you, for the other pre-ESM users of ts-proto, they might read the imports and just do an raised eyebrow about "why is that there?", even if I agree it won't break anything for them.

If the conditional-if-`.`-in-`importString` fixes your `.pb.js` use case, let's do that? But if not, yeah, I'm good with always having it.

paralin: I agree it's confusing and at first when I saw the .js imports in files I was like.. wat?

But like you said, it's the way the TypeScript developers decided to go for this, and I guess it's slightly less ambiguous than guessing to add the .ts suffix.

paralin:
Probably can be done in a separate PR but this is also necessary if ESM is enabled:

```
import * as _m0 from 'protobufjs/minimal.js';
```

paralin:
Seems like this requires a slightly newer typescript - 4.7.x vs the current. I'll untangle it and get the tests to pass.

src/utils.ts
paralin: With regards to importing with '.ts' suffix: https://github.com/microsoft/TypeScript/issues/37582

paralin:
This PR needs a few tweaks, looking at it now.

paralin:
Needs https://github.com/stephenh/ts-poet/pull/35 first

paralin:
@stephenh Fixed.

Ts-jest doesn't support the .js import alias yet, so it required adding `resolver: 'jest-ts-webcompat-resolver'` to the jest.config.js.

paralin:
Okay, I think it's good to go now 👍🏽

src/utils.ts
stephenh: Nice! Thanks for the link to the TS issue re `.ts` imports.

zsmatrix62:
I think this change is making my error like below today: " Module not found: Error: Can't resolve '.xxxx.js' in ...

image

Are there any options to control the output just remove `.js` extension?

paralin:
@zsmatrix62 if you use typescript 4.7 or later, it will correctly resolve the .js to the .ts

zsmatrix62:
> @zsmatrix62 if you use typescript 4.7 or later, it will correctly resolve the .js to the .ts

I'm using Angular13 which runs on ts 4.4 and only.

Issue #615, opened 2022-07-14
ESM packages still not working because of protobufjs/minimal


siendoricardo:
The package `protobufjs/minimal` is being import as `import * as _m0 from "protobufjs/minimal";`
(Produced the following error):

```
/home/.../WebstormProjects/.../node_modules/ts-node/dist-raw/node-internal-modules-esm-resolve.js:366
    throw new ERR_MODULE_NOT_FOUND(
          ^
```
And it is required to be imported as follows: `import _m0 from "protobufjs/minimal.js";`

paralin:
Good to go in #616  :)

PR #616, opened 2022-07-14
fix: internal packages now takes suffix and esm opt.

siendoricardo:
This PR fixes: #615

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -198,6 +198,10 @@ export function getPropertyAccessor(objectName: string, propertyName: string, op
     : `${objectName}${optional ? '?.' : ''}[${JSON.stringify(propertyName)}]`;
 }
 
+export function impFile(options: Options, spec: string) {
+  return imp(`${spec}${options.importSuffix}`);
+}
+
 export function impProto(options: Options, module: string, type: string): Import {
   const importString = `${type}@./${module}${options.fileSuffix}${options.importSuffix}`;
   if (options.onlyTypes) {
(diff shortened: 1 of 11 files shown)


src/utils.ts
```diff
@@ -198,6 +198,13 @@ export function getPropertyAccessor(objectName: string, propertyName: string, op
     : `${objectName}${optional ? '?.' : ''}[${JSON.stringify(propertyName)}]`;
 }
 
+export function impSlash(options: Options, spec: string) {
+  if (!spec.includes('/'))
```
paralin, line 202: There are a lot of imports like `@foo/bar` which have a slash but are not referencing a specific file.

I don't see why it's necessary to add another function `impSlash` as opposed to just appending `importSuffix` everywhere `protobufjs/minimal` is imported?

paralin (changes requested):

src/utils.ts
siendoricardo: That's what the function it's doing. I just thought it looks cleaner to have it as util. What might be redundant are the includes check.

paralin: Maybe call it `impFile` instead of `impSlash` and remove the `/` check ?

siendoricardo: That sounds great. 👍

siendoricardo:
I'll fix the lint.

paralin (approved):

paralin:
@siendoricardo Squashed your commits to 1 commit & added some more info in the commit description, hope that's ok w/ you!

@stephenh This looks good to me.

stephenh (approved):
Looks great, thanks @paralin and @siendoricardo !

Issue #539, opened 2022-03-23
A request should not be retried due to errors in these "upstream codes"


doctorpangloss:
https://github.com/stephenh/ts-proto/blob/a810b388777f464e79bcff51883f19c9d767e94f/src/generate-grpc-web.ts#L293

The error should be propagated to onError and the client should decide what to do

doctorpangloss:
To clarify my interpretation of the spec is sending these errors closes the stream. It isn't within spec that then the client retries the same exact request. It was a surprise giving me trouble for a while.

stephenh:
@doctorpangloss yeah, I agree it sounds like this behavior should not be hard-coded, with the disclaimer IANAE an expert on grpc-web / nor use it personally, so ts-proto has had to rely on users incrementally improving the grpc-web-specific output as their individual use cases have needed.

It'd be great great if you want to submit a PR that makes sense defaults configurable, with like a runtime constructor arg or something like that.

Thanks!

snyh:
only the status code 14 (UNAVAILABLE) may be retry automatically , all other status code hasn't any sense to retry. 

the code 
```
            } else if (upStreamCodes.includes(code)) {
              setTimeout(upStream, DEFAULT_TIMEOUT_TIME);
```
should be deleted directly.

stephenh:
@snyh I don't personally use this code, but if you want to submit a PR that makes this behavior configurable, with what you think are more sane defaults, that'd be great.

stephenh:
Fixed in #618 , thanks @hersentino !

PR #618, opened 2022-07-18
feat: add dynamic upStreamCodes option

hersentino:
This PR introduces dynamic upStreamCodes option following discussions on #539


doctorpangloss (changes requested):
The default upstream codes should be an empty array, it never makes sense for this code to retry. Let the application always decide what to do by default

hersentino:
@stephenh all done :)

stephenh:
Thanks @hersentino !

PR #619, opened 2022-07-20
docs: add `node_modules` exclusion

TonyPapousekLC:
resolves `Failure: node_modules/protobufjs/google/api/annotations.proto:5:8:google/api/http.proto: does not exist`

Note: This assumes `buf.yaml` is in the top-level project directory, next to `node_modules`


stephenh:
Thanks!

PR #620, opened 2022-07-21
feat: add meta-data to stream error

hersentino:
This PR introduces metadata support in Stream (invoke) during catch as in Unary. It's required to read gRPC richer errors.
S

### DIFF
--- a/integration/grpc-web/example.ts
+++ b/integration/grpc-web/example.ts
@@ -876,13 +876,16 @@ export class GrpcWebImpl {
           metadata: maybeCombinedMetadata,
           debug: this.options.debug,
           onMessage: (next) => observer.next(next),
-          onEnd: (code: grpc.Code, message: string) => {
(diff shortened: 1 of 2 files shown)


stephenh:
@hersentino cool, this looks good to me; can you run `yarn bin2ts` to update the generated code in `integration/`? Having the code output checked basically acts as our regression suite. Thanks!

hersentino:
@stephenh If you check pr changes we can see `integration/grpc-web/example.ts` has been updated. To be sur I ran `yarn bin2ts` again and nothing change

stephenh:
@hersentino ah! yep, I see it that now, sorry I was just expecting more files than just this one to change, but I guess this is the only integration test we have for this output right now. Which is fine, I'll merge. Thanks!

hersentino:
Ok perfect ! Thank you

PR #624, opened 2022-07-25
fix: fieldmask wrap method with type registry

stezu:
Fixes the following type error:
```
../protos/dist/google/protobuf/field_mask.ts:274:5 - error TS2741: Property '$type' is missing in type '{ paths: string[]; }' but required in type 'FieldMask'.

274     return { paths: paths };
        ~~~~~~~~~~~~~~~~~~~~~~~~

  ../protos/dist/google/protobuf/field_mask.ts:209:3
    209   $type: 'google.protobuf.FieldMask';
          ~~~~~
    '$type' is declared here.
```

When ts-proto is run with `outputTypeRegistry=true`.

### DIFF
--- a/integration/fieldmask/google/protobuf/field_mask.ts
+++ b/integration/fieldmask/google/protobuf/field_mask.ts
@@ -261,7 +261,11 @@ export const FieldMask = {
   },
 
   wrap(paths: string[]): FieldMask {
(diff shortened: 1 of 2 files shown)


stephenh:
Cool, thanks @stezu !

PR #629, opened 2022-07-28
chore: add regression test for #628

### LINKED ISSUE #628: Recent change broke client-streaming methods
After upgrading ts-proto something causes client->server streams to return AsyncIterable instead of Promise:

```
diff --git a/echo/echo.pb.ts b/echo/echo.pb.ts
index 406004f..2a7e00d 100644
--- a/echo/echo.pb.ts
+++ b/echo/echo.pb.ts
@@ -103,7 +103,7 @@ export interface Echoer {
   /** EchoServerStream is an example of a server -> client one-way stream. */
   EchoServerStream(request: EchoMsg): AsyncIterable<EchoMsg>
   /** EchoClientStream is an example of client->server one-way stream. */
-  EchoClientStream(request: AsyncIterable<EchoMsg>): Promise<EchoMsg>
+  EchoClientStream(request: AsyncIterable<EchoMsg>): AsyncIterable<EchoMsg>
   /** EchoBidiStream is an example of a two-way stream. */
   EchoBidiStream(request: AsyncIterable<EchoMsg>): AsyncIterable<EchoMsg>
   /** RpcStream opens a nested rpc call stream. */
@@ -138,7 +138,7 @@ export class EchoerClientImpl implements Echoer {
     return EchoMsg.decodeTransform(result)
   }
 
-  EchoClientStream(request: AsyncIterable<EchoMsg>): Promise<EchoMsg> {
+  EchoClientStream(request: AsyncIterable<EchoMsg>): AsyncIterable<EchoMsg> {
     const data = EchoMsg.encodeTransform(request)
     const promise = this.rpc.clientStreamingRequest(
       'echo.Echoer',
diff --git a/package.json b/package.json
index a838df4..2810cf2 100644
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "eslint-config-prettier": "^8.5.0",
     "prettier": "^2.7.1",
     "rimraf": "^3.0.2",
-    "ts-proto": "^1.117.0",
+    "ts-proto": "^1.121.0",
     "typescript": "^4.7.4",
     "utf-8-validate": "^5.0.9"
   },
```

```
		--ts_proto_opt=esModuleInterop=true \
		--ts_proto_opt=fileSuffix=.pb \
		--ts_proto_opt=importSuffix=.js \
		--ts_proto_opt=forceLong=long \
		--ts_proto_opt=oneof=unions \
		--ts_proto_opt=outputServices=default,outputServices=generic-definitions \
		--ts_proto_opt=useAsyncIterable=true \
		--ts_proto_opt=useDate=true \
```

I'll try to figure out which commit did it.

paralin:
"responsePromiseOrObservable" was changed to return the Observable type if clientStreaming is set:

https://github.com/stephenh/ts-proto/commit/d3e7f1f4aac12db87c54d6357e557a351c96a2ca#r79607248

And reverted recently.

Adds a couple regression tests to be sure we don't break client -> server streams in future.

Fixes #628


paralin:
Not sure exactly where but this still needs a couple tweaks to pass the tests

paralin:
@stephenh Added the regression test here, I think we should merge it separately from the other grpc-web changes.

stephenh:
Yep, +1

PR #631, opened 2022-07-28
feat: grpc-web support promise stream

luhuaei:


src/generate-services.ts
```diff
@@ -57,11 +58,18 @@ export function generateService(
       params.push(code`ctx: Context`);
     }
 
-    // the grpc-web clients auto-`fromPartial` the input before handing off to grpc-web's
-    // serde runtime, so it's okay to accept partial results from the client
-    const partialInput = options.outputClientImpl === 'grpc-web';
-    const inputType = requestType(ctx, methodDesc, partialInput);
-    params.push(code`request: ${inputType}`);
+    // grpc-web client stream or bidirectional stream not request params when use Promise.
+    if (!options.returnObservable && options.outputClientImpl === 'grpc-web' && methodDesc.clientStreaming) {
```
paralin, line 62: Just invert the if statement here.

if (options.returnObservable || options.outputClientImpl !== "grpc-web" || !methodDesc.clientStreaming)

src/generate-services.ts
```diff
@@ -20,6 +20,7 @@ import {
 import SourceInfo, { Fields } from './sourceInfo';
 import { contextTypeVar } from './main';
 import { Context } from './context';
+import { grpcWebResponsePromiseOrObservable } from './generate-grpc-web';
```
paralin, line 23: How does this compare to the solution in #629

paralin (changes requested):

src/generate-services.ts
luhuaei: This pull request more than to support grpc-web Promise stream. 

After this pull request, fix incorrect return type( #629 ) can do more simple.

paralin: I don't understand what you mean by "can do more simple"

paralin: My question is - why add a new function `grpcWebResponsePromiseOrObservable` instead of adding a parameter to the previous function?

luhuaei: I want to support `Promise Stream`, need to add some specifical response type, but this only using in `grpc-web`.

paralin: Ok, I see.

Please `git fetch` and then `git rebase 557a5a186fbc375f09c3b3ceea14fe502c5ce34d`

This will bring in the regression test to ensure that the issue from #629  is also fixed.

src/generate-services.ts
paralin: "grpc-web client stream or bidirectional stream not request params when use Promise."

I can't understand this sentence, could you make it more clear?

src/generate-services.ts
luhuaei: Ok, I updated it

paralin:
@luhuaei Are you still working on this? Am happy to help with getting it ready to merge.

luhuaei:
> @luhuaei Are you still working on this? Am happy to help with getting it ready to merge.

@paralin  thanks!

I make other pull request #633 to support grpc-web stream method.

src/generate-services.ts
luhuaei: When using the `Promise` return type, the input parameter does not require the `request` parameter.

See https://github.com/stephenh/ts-proto/pull/631/files#diff-cf75bd2b26f57126e3c440377e55dfa55c406359a04d2029355db727cac33c3eR44

In this implementation, the `request` parameter is not required. Use `callback` to get the relevant data stream, use the `write` method to send data streams.

PR #638, opened 2022-08-03
fix: remove Record word conflict

stezu:
When a protobuf file contains a message named `Record`, it causes conflicts with the utility types at the bottom of the file which were using the [built-in TS `Record` utility](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type). This obviously causes the file to contain invalid Typescript since we'd expect `Record` to be the utility generic, not the non-generic type generated by this library.

To alleviate this issue, I added a test to treat `Record` as a reserved word and fixed the implementation to avoid using it and instead used the [mapped types](https://www.typescriptlang.org/docs/handbook/2/mapped-types.html) to emulate what `Record` does. The alternatives I considered seemed to be more trouble than they were worth given that there was only the single usage of `Record` in the generated output.


stephenh:
@stezu this looks good!

Can you run `yarn proto2bin` to re-create the `*.bin` files? In theory none of those should have changed, and the `proto2bin` script will use docker to make sure any host-dependent / `protoc`-version-dependent changes are not included in the `*.bin` files.

stezu:
@stephenh I was wondering about that, thanks for the tip!

stephenh:
Looks great, @stezu , and thanks for the regression test on `Record`!

PR #649, opened 2022-08-06
perf: Faster base64FromBytes & bytesFromBase64 on Node.JS

coyotte508:
I didn't change the browser code but `for .. of` would probably be more performant than `forEach` on it

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -43,7 +43,7 @@ var globalThis: any = (() => {
 
 function bytesFromBase64(b64: string): Uint8Array {
   if (globalThis.Buffer) {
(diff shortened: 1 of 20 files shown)


stephenh:
@coyotte508 looks great! Looks like there are a few test failures/need the `fix` / `feat` prefix for your PR, but otherwise lgtm. Thanks!

coyotte508:
The failing test was caused because expecting a Uint8Array instead of a Buffer (even though Buffer is a subclass of Uint8Array and is type-compatible since node 4.x).

I used `Uint8Array.from()` to convert the buffer back to an `Uint8Array`

stephenh:
Thanks @coyotte508 , I'll hit merge!

Issue #630, opened 2022-07-28
Convert enum zero values to undefined when using string enum


Celend:
Assume we have enum in proto:
```protobuf
enum Product {
  PRODUCT_UNSPECIFIED = 0;

  PRODUCT_A = 1;

  PRODUCT_B = 2;
}

message Example {
  Product product = 1; // allow undefined
}
```

And generate code with options `stringEnums=true` and `enumsAsLiterals=true`,
when `product` field is unset, the deserializer will output `field = 'PRODUCT_UNSPECIFIED'`,
in actual development, is not so handy and I need to convert it again to `undefined`.

So is good idea to add an option which auto convert zero enum value to `undefined` or I should't use `stringEnums`?

If it's ok to add an option, I'm glad to make a PR 😄

stephenh:
Huh, I was going to say that, because of proto3's default values are not serialized:

https://developers.google.com/protocol-buffers/docs/proto3#default

That we can't really know if the client "did not set `product` at all" or "they set product to unspecified".

But your ask is, for either one, you don't really care, you really want unspecified to "go away" and just become undefined...

So yeah, I get what you're saying. Maybe a new option like `enumUnspecifiedAsUndefined=true`...

Wouldn't this new option work for both `stringEnums=true` and `false`?

Celend:
TFTR! 

I think the new option only available for `stringEnums=true`,
because number enum have easy way to check the field is unset or unspecified, like `msg.enum === 0` but string enum can't.

Actually my current workaround is use number enum then use toJSON like below:
```typescript
for (const key of ['enumfield1', 'enumfield2']) {
  if (data[key] === 0) {
    delete data[key]
  }
}

return Message.toJSON(data)
```

`toJSON` / `fromJSON` may also changed.  

Have any other problems or side effects?

PR #650, opened 2022-08-07
feat: add enumUnspecifiedAsUndefined option

Celend:
see #630

But I'm not sure the value of `Struct.NulLValue` should be `null` or `undefined`, current is `undefined`, it's looks a little weird.


src/enums.ts
```diff
@@ -72,12 +81,22 @@ export function generateEnumFromJson(ctx: Context, fullName: string, enumDesc: E
...
   for (const valueDesc of enumDesc.value) {
     chunks.push(code`
-      case ${valueDesc.number}:
+      case ${valueDesc.number}:`);
+
+    if (options.enumUnspecifiedAsUndefined && valueDesc.number === 0) {
+      chunks.push(code`
+      case undefined:
+        return undefined`);
+      continue;
```
stephenh, line 96: Shouldn't we leave this `continue` out? Because if some other system (that is not ts-proto) sends us a enum with the 0 value, but as the string "FOO_UNKNOWN", we'd want to still treat it as undefined.

I.e. something like:

```
  switch (object) {
    case 0:
    case 'UNKNOWN':
    case undefined:
      return undefined;
    case 2:
    case 'ON':
```

Granted, if the other system is itself ts-proto, or using the proto3 rules to not include the enum value if its the 0th value anyway, this extra "case 'UNKNOWN'" would be strictly required, but it seems like a good idea to include just in case?

Also, do we need to include "case undefined"? I don't think its part of the proto3 JSON spec to send `state: undefined` on the wire.

I see that you've changed `toJSON` to also return `undefined`:

```

export function stateEnumToJSON(object: StateEnum | undefined): string | undefined {
  switch (object) {
    case undefined:
      return undefined;
    case StateEnum.ON:
      return 'ON';
    case StateEnum.OFF:
      return 'OFF';
    case StateEnum.UNRECOGNIZED:
    default:
      return 'UNRECOGNIZED';
  }
}
```

But I don't think we should do this; i.e. the `enumUnspecificAsUndefined` should only control how ts-proto messages are represented internally and not change how messages are encoded/decoded from JSON on the wire.

src/enums.ts
```diff
@@ -110,26 +129,32 @@ export function generateEnumToJson(ctx: Context, fullName: string, enumDesc: Enu
...
-    if (ctx.options.useNumericEnumForJson) {
+    if (options.useNumericEnumForJson) {
       chunks.push(code`case ${fullName}.${valueDesc.name}: return ${valueDesc.number};`);
-    } else {
-      chunks.push(code`case ${fullName}.${valueDesc.name}: return "${valueDesc.name}";`);
+      continue;
     }
+
+    if (options.enumUnspecifiedAsUndefined && valueDesc.number === 0) {
+      chunks.push(code`case undefined: return undefined;`);
```
stephenh, line 146: I.e. per my other comment, I think this should be `case undefined: return "${valueDesc.name}"`.

stephenh (commented):
Hey @Celend , thanks for the PR; apologies for the late code review. This looks pretty good, just some questions around the JSON encoding.

src/options.ts
```diff
@@ -190,6 +192,13 @@ export function optionsFromParameter(parameter: string | undefined): Options {
     }
   }
 
+  if (options.enumUnspecifiedAsUndefined) {
+    // enumUnspecifiedAsUndefined requires stringEnums=true
```
stephenh, line 196: I know that, at runtime, `msg.enumValue === 0` is easy to check, but b/c during codegen time we can easily pick up the `value.number === 0` enum, i.e.:

```
      const enumProto = typeMap.get(field.typeName)![2] as EnumDescriptorProto;
```

It seems like we should be able to support this for all enums, and not just `stringEnums`?

Celend:
@stephenh I'll fix in this week, please wait :D

src/options.ts
Celend: yeah, indeed that is not hard to support

src/enums.ts
Celend: You're right, I have not thought about that👍

Celend: > Also, do we need to include "case undefined"? I don't think its part of the proto3 JSON spec to send state: undefined on the wire.

Just in case for some other use case, I think we should not send 'UNRECOGNIZED' or throw an error (when unrecognizedEnum=false) when value is undefined?

Celend: > I see that you've changed toJSON to also return undefined:

Because I'd removed the default(unspecified) value from enum definition, if `toJSON` output `XXX_UNSPECIFIED`, that is inconsistent behavior, unless don't remove the default value from enum?
But also strange, the default value already decoded as `undefined`, why keep the default value in enum?

src/options.ts
Celend: I found we don't need to do any thing to support that, fs proto3 really did not serialize default values, just change createBaseXXX and make enum field default value to undefined, the decode method wouldn't read value to message field

Celend:
@stephenh I pushed new code, there are some question/changes need to confirm:
1. number enum support
2. toJSON output `undefined` or not
3. fromJSON received `undefined` or not

Please take a look and confirm, thanks!

PR #651, opened 2022-08-08
fix: resolve import collisions for services

doochik:


stephenh:
Cool, thanks @doochik !

PR #654, opened 2022-08-17
feat: Add optional json field names map to message type

GregVS:
Protobuf messages may be stored as JSON objects, for example in a database. Performing query or lookups on these objects often requires a string representation of the message's field names (keys). Currently, the programmer needs to hardcode these strings in their code, but if the underlying protobuf definition changes, there will be no warning that those field names are invalid. Providing an optionally generated mapping of each key to its corresponding JSON name allows Typescript to flag if that field does not exist. This can be enabled using `outputJsonFieldNames=true` which adds a map object to the message type named `$jsonFields`.

Example Proto:
```
message Foo {
  string bar = 1;
}
```

Typescript Generated Class Member:
```
const Foo = {
  $jsonFields: {
    bar: "bar"
  } as const,
  ...
}
```

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -397,6 +397,8 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=outputTypeRegistry=true`, the type registry will be generated that can be used to resolve message types by fully-qualified name. Also, each message will get extra `$type` field containing fully-qualified name.
 
+- With `--ts_proto_opt=outputJsonFieldNames=true`, each message will get an extra `$jsonFields` object containing a mapping from each message key to it's corresponding JSON name.
+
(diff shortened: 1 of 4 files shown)


stephenh:
@GregVS thanks for the PR! My initial thought is that this seems like a slippery slope to the existing `outputTypeRegistry` feature.

Granted, I assume you've seen that and don't need _everything_ that it does?

I also get that it's nice to have `$jsonFields` directly on the `Foo` const...

I dunno, maybe we could generalize this a little bit and do like:

```
const Foo = {
  $fields = { bar: { jsonName: "bar" } }
}
```

With an option like `fieldsMetadata=jsonName`? Then maybe eventually someone wants `fieldsMetadata=jsonName,type` and we could do `$fields = { bar: { jsonName: "bar", type: "int[]" } }` or something like that.

Basically make it a little bit more generic so that `$fields` could have more things added to it in the future.

Also, if you could include an `integration-test/...` with a tiny `*.proto` and then `parameters.txt` set to use the new flag you're introducing, that would be great as those serve as our regression tests.

Issue #494, opened 2022-01-25
Add better return types for toJson methods


stephenh:
See #492 but it'd be nice to `.toJson()` to return `JsonOf<FooMessage>` or `FooMessageJson` so that API layers that want "basically JSON, but with a few custom mappings" could reuse `.toJson()` instead of missing out b/c of the `unknown`.

boukeversteegh:
Just adding some thoughts: Would `JsonOf` be new generic type? In several cases, the generated Json has a different interface than the native typescript interface, so I guess this implies that a generic type would not work for all cases. It may require a dedicated interface per message, or only for those cases where there are canonical representations.

stephenh:
@boukeversteegh yeah, good point that, if `JsonOf` was a mapped type, it would need to have conditionals to recognize the various "this TS native type --> this one-off JSON type"...

Which, yeah, that sort of complexity is typically handled better by just doing codegen.

Issue #492, opened 2022-01-23
Next.js does not serialize values that are undefined


m8vago:
First of all thanks for your hard work, ts-proto is really awesome!

Recently we'd updated to the latest version and noticed, the grpc response has `undefined` values as a default for fields which aren't sent in the wire. 

I think the change was made in this commit:
`https://github.com/stephenh/ts-proto/commit/70832d33bae82ecb3c5f87845d14e992a13437e4#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3d`

However this broke our Next.js frontend, because _next_ forbids to send an object in `getServerSideProps()` as a response that has any `undefined` value.

As far as i see, in the current code-base there is no option to disable the default `undefined` value for optional fields.

I am aware of `Type.toJson()` calls are dealing with this issue, but those are returning `any` types, which would make our code less type-safe.

Would it be possible to have a flag that restores the old behavior?

stephenh:
Huh, that's interesting. I guess the next js restriction makes sense, insofar as it's expected a valid JSON value (I assume) and undefined as a value is not valid JSON.

As you pointed out, this is b/c you're putting "not meant to be JSON" objects on the wire, and previously it had just happened to work out. I'd be a little worried that this is a slippery slope, i.e. that besides the `undefined` issue that eventually you'll run into other mismatches where instead of just "proto messages that are implicitly JSON.stringified", you might want other "better/actual JSON" output that comes from the canonical protobuf JS encoding, which again as you mention is what the `.toJson`s do...

For example, how do you handle `Date`s? By default, ts-proto will use js `Date`s in messages, so the message type will be `{ startDate: Date }` but the "actual JSON type" is `{ startDate: string }`. Does your `getServerSideProps` in this case have to return `startDate: string` i.e. the literal JSON-on-the-wire type?

Fwiw, the reason we introduced this change is that setting all keys directly on instantiation is supposed to help (handwaving) with v8 optimizations, so I'd at least want to keep it as the default behavior.

Seems like there are a few options:

1. As you said, we could add an option to not set keys as `undefined` by default
[...]

PR #656, opened 2022-08-18
feat: omit optional fields in base instance

robot9706:
Releated to #494 and #492.
Should fix the issue where nextjs is unable to serialize server side props.

By default the feature is disabled so nothing changes in the genererated code, optional fields can be omited using the "omitOptionalUndefineds=true" option.


stephenh:
Hey @robot9706 , thanks for the PR!

I'm remembering this issue now... I think this sounds good, to fix it via a flag. A few asks:

* And you run `yarn proto2bin` in the top-level directory? This should use docker to generate the `*.bin` files and so be host & protoc-version independent. Basically they shouldn't show up as changed in your PR.

* Can you add a new `integration-tests/*` directory with a super-small `*.proto` and `parameters.txt` that shows the generated output with the new flag enabled?

* Maybe call this `initializeFieldsAsUndefined=true/false`? Wdyt of that name?

* Should we change `initializeFieldsAsUndefined` to `false` if `nestJs=true` is set?

* Can you add a note about the new flag to the readme?

Apologies for the ~3-4 asks, but overall this lgtm and we can work on getting it merged. Thanks!

robot9706:
Hi! Thanks for checking out the PR!

> * And you run `yarn proto2bin` in the top-level directory? This should use docker to generate the `*.bin` files and so be host & protoc-version independent. Basically they shouldn't show up as changed in your PR.

I did run `yarn proto2bin` in the top-level directory, I'll look into fixing the test files then.

> * Can you add a new `integration-tests/*` directory with a super-small `*.proto` and `parameters.txt` that shows the generated output with the new flag enabled?

Yes, sure.

> * Maybe call this `initializeFieldsAsUndefined=true/false`? Wdyt of that name?

In my opinion this option should be disabled by default since it changes some optimisations, that's why I wanted to name it something which reflects that it's off by default, and setting to "true" enables it.
But `initializeFieldsAsUndefined` works for me as well if you think that's a better name. 

> * Should we change `initializeFieldsAsUndefined` to `false` if `nestJs=true` is set?

That's a good question, I think it yes that could be the default behaviour but I'll think about it and get back to you.

> * Can you add a note about the new flag to the readme?

Oh I forgot to add the new flag to the readme, my bad.

robot9706:
I pushed some changes:

- Changed the name of the option to the one you suggested, also made sure to set it to `false` if `nextJs` is enabled.
- Added a new integration test.

After rebuilding the tests loads of other `*.bin` files changes as well, I'm not sure if I should commit these files... any input on these?

stephenh:
Hey @robot9706 , this looks great! The CI build failed with a very minor failure I fixed by just merging main in #669 , and so your work got merged via that PR and I'll close this one out. Thanks!

robot9706:
Thanks for your help!

Issue #643, opened 2022-08-04
numerical enum first value not being picked up


khi195:
hi, any enum which is equal to 1 is not being picked up.  For example

export enum AccountType {
  CURRENT = 1,
  SAVING = 2,
  JOINT = 3,
  UNRECOGNIZED = -1,
}

in the above example CURRENT = 1 is not being picked up when we make the network call? What am i doing wrong, how can I remedy this.

khi195:
I think the problem could be related to this encoding function,
    if (message.accountType !== 1) {
      writer.uint32(72).int32(message.accountType);
    }

this is not allowing enums which are equal to 1, which obv incorrect. in the above example, when CURRENT=1 this is not being picked up.

stephenh:
Ah yeah, you are probably running into this:

https://github.com/stephenh/ts-proto/blob/main/src/types.ts#L254

You could probably make `notDefaultCheck` return `Code | undefined`, instead of `Code`, and in the "there is no v.number === 0" case, return `undefined`, and have that tell `generateEncode` line 1107 to not put in that `if` check.

If you'd like to submit a PR, that'd be great. Thanks!

sefaphlvn:
is there any update in here?

sefaphlvn:
in my case I dont want to this condition `${place} !== ${enumType}.${enumValue}`

Also I dont have experience for this.

https://github.com/stephenh/ts-proto/blob/75ba9073744e0c086540a8e3ae1edad629bdf85a/src/types.ts#L272C7-L275C84

`      if (options.stringEnums) {
        const enumType = messageToTypeName(ctx, field.typeName);
        const enumValue = getEnumMemberName(ctx, enumProto, zerothValue);
        return code`${maybeNotUndefinedAnd} ${place} !== ${enumType}.${enumValue}`;`

sefaphlvn:
I did some hardcoding in type.ts file, but it's a workaround. I'm writing this for you to understand my needs. 



Replace the following line:
```
case ts_proto_descriptors_1.FieldDescriptorProto_Type.TYPE_BOOL:
    return (0, ts_poet_1.code) `${place} === true`;
```
with:
```
case ts_proto_descriptors_1.FieldDescriptorProto_Type.TYPE_BOOL:
    return (0, ts_poet_1.code) `${place} !== undefined`;
```

-------------------------------------------------------------------

Replace the following block:
```
if (options.stringEnums) {
    const enumType = messageToTypeName(ctx, field.typeName);
[...]

PR #657, opened 2022-08-19
disabled the default check for the upstream encoder

khi195:
this is a PR for the following issue
https://github.com/stephenh/ts-proto/issues/643

for the upstream encoder, this PR has the ability to disable the check for default enums, because for proto2 we may not have a default enum, so will have to disable the default behaviour.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1125,11 +1125,29 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
         }
       `);
     } else if (isScalar(field) || isEnum(field)) {
-      chunks.push(code`
-        if (${notDefaultCheck(ctx, field, messageDesc.options, `message.${fieldName}`)}) {
-          ${writeSnippet(`message.${fieldName}`)};
+      /**
+       *
+       * this is needed for proto2 when a default value may not be set on the server
+       * so remove the if statement around the snippet
+       */
+
+      const snippet = writeSnippet(`message.${fieldName}`);
+
+      const { disableDefaultCheck, disableDefaultEnumCheck } = options;
+
+      if (disableDefaultCheck) {
+        chunks.push(code`${snippet};`);
+      } else {
+        if (isEnum(field) && disableDefaultEnumCheck) {
+          chunks.push(code`${snippet};`);
+        } else {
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -1125,11 +1125,29 @@ function generateEncode(ctx: Context, fullName: string, messageDesc: DescriptorP
         }
       `);
     } else if (isScalar(field) || isEnum(field)) {
-      chunks.push(code`
-        if (${notDefaultCheck(ctx, field, messageDesc.options, `message.${fieldName}`)}) {
```
stephenh, line 1129: Hey @khi195 ; thanks for the PR! Instead of adding more options, I think we should try and solve this by being smarter at handling "there is no enum with value === 0".

I.e. this line is calling `notDefaultCheck`, which inside of that method has some code that looks like:

```
      if (options.stringEnums) {
        const enumType = messageToTypeName(ctx, field.typeName);
        return code`${maybeNotUndefinedAnd} ${place} !== ${enumType}.${zerothValue.name}`;
      } else {
        return code`${maybeNotUndefinedAnd} ${place} !== ${zerothValue.number}`;
      }
```

Note the `|| enumProto.value[0]`. That is the line that is tripping up your code, b/c it means ts-proto is treating the "the enum value 1" as the default, which it just shouldn't do.

Can you try removing that `|| enumProto.value[0]` and instead doing something like:

```
      const zerothValue = enumProto.value.find((v) => v.number === 0);
      if (!zerothValue) {
        return undefined;
      } else if (options.stringEnums) {
        const enumType = messageToTypeName(ctx, field.typeName);
        return code`${maybeNotUndefinedAnd} ${place} !== ${enumType}.${zerothValue.name}`;
      } else {
        return code`${maybeNotUndefinedAnd} ${place} !== ${zerothValue.number}`;
      }
```

And then here in the `generateEncode` method, you could do:

```
const maybeDefaultCheck = notDefaultCheck(ctx, field, messageDesc.options, `message.${fieldName}`);
```

And if `maybeDefaultCheck` is undefined, then leave the `if` statement out, just like you're currently doing by looking at the new `disableDefaultCheck` options.

I'm pretty sure that should work.

PR #672, opened 2022-09-20
feat(options): adds protoc-gen-go-like M option

### LINKED ISSUE #596: Feature: Importmap option

iamricard:
This commit adds a new feature that aims to provide the same functionality as `M` for the go compiler (https://developers.google.com/protocol-buffers/docs/reference/go-generated#package). I opted for not modifying the type map, because we'll also want to figure out whether or not we need to generate the file, so we'll be accessing the `options.M` everywhere and I found it less confusing to always get mapping information from `options.M` instead of `options.M` or the type map depending on the decision that we're making (we need to know if we need to do a relative import or just the raw provided import too).

@stephenh there was an alternative approach I considered: change the type registry module to include information about whether or not the module had been remapped, but that seemed to add more complexity than the option I went with. I'm happy to revisit it though, if you think some other approach would be a better fit for the future of the codebase.

Fixes #596


stephenh:
This looks great, thanks @iamricard !

leandro-branco:
@iamricard great feature! Thanks for it!

I'm just wondering why it was needed to filter the filesToGenerate in the `plugin.ts` file.

```typescript
    filesToGenerate
      .filter((file) => !options.M[file.name]) // THIS LINE
      .map(async (file) => {
        const [path, code] = generateFile(ctx, file);
        const content = code.toString({ ...getTsPoetOpts(options), path });
        return { name: path, content };
      })
  );
```

 In a mono repo like `nrwl/nx`, it's not uncommon to  generate the typescript files and also apply the import mapping linking to another module. 

Perhaps, the reason is the filter is preventing the plugin to generate google protobuf well known types , like empty. If it's the case, the option `emitImportedFiles` cloud be used to do the job. 

Is there any impact if we remove the filter?

Thanks again.

PR #673, opened 2022-09-21
fix(options): initializes M opt to empty object

iamricard:
So because the test for the `M` opt is really the output of `yarn build:test`
I didn't catch a bug I introduced in my last iteration of refactoring.
In `options`, `M` is `undefined` so passing `M` opts doesn't work
because it'll try `options.M['...']` and `options.M` is undefined.
`options` should start out with the default `M`.

Apologies for the bug!


stephenh (approved):

stephenh:
Np!

iamricard:
wow that was a fast review 👀

PR #680, opened 2022-09-29
feat: CLI option: forcePartials

fizx:
This allows you to turn on the grpc-web-like behavior of having clients and interfaces accept partials, without fully commiting to something like useOptionals=all where you have question-marks on all of your interfaces.


README.markdown
```diff
@@ -362,6 +362,9 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=outputPartialMethods=false`, the `Message.fromPartial` methods for accepting partially-formed objects/object literals will not be output.
 
+- With `--ts_proto_opt=forcePartials=true`, the generated interfaces and clients for services will accept partials.  This is less type-safe, and similar to `optionalTypes=all` without adding ?'s to all of your interfaces.
```
stephenh, line 365: Wdyt about `useOptionals=services` as a way to ask for services to use partials but not the messages themselves?

fizx:
I think I'm closing this one because we have another angle on the problem.  Please look forward to a future typeregistry patch.

PR #683, opened 2022-10-11
feat(client): allow overriding the service identifier

paralin:
Currently the service identifier is the fully qualified "fullName" of the service, for example, "simple.Echoer"

This identifier is hard-coded in the auto-generated client with string literals.

Instead, allow overriding the service ID with a new optional argument in the client constructor, default the service ID to the fullName, and store the ID in a read-only field on the client object.

This allows the user to override the service ID if desired while still maintaining the old behavior otherwise.

### DIFF
--- a/integration/async-iterable-services/simple.ts
+++ b/integration/async-iterable-services/simple.ts
@@ -101,7 +101,9 @@ export interface Echoer {
 
 export class EchoerClientImpl implements Echoer {
   private readonly rpc: Rpc;
-  constructor(rpc: Rpc) {
+  private readonly service: string;
+  constructor(rpc: Rpc, opts?: { service?: string }) {
+    this.service = opts?.service || "simple.Echoer";
     this.rpc = rpc;
     this.Echo = this.Echo.bind(this);
     this.EchoServerStream = this.EchoServerStream.bind(this);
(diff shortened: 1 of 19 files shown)


integration/async-iterable-services/simple.ts
```diff
@@ -101,7 +101,9 @@ export interface Echoer {
 
 export class EchoerClientImpl implements Echoer {
   private readonly rpc: Rpc;
-  constructor(rpc: Rpc) {
+  private readonly service: string;
+  constructor(rpc: Rpc, service?: string) {
```
stephenh, line 105: @paralin wdyt about a `constructor(rpc: Rpc, opts?: { service?: string })` so that we could add more opts in the future w/o a breaking api change?

I'm pretty happy to ship as-is and then just make the opts change later if ever needed.

stephenh (approved):

integration/async-iterable-services/simple.ts
paralin: I'm happy with either, opts works for me.

paralin: @stephenh I updated the PR to match your suggestion! 🚀

paralin:
Updated syntax:

```
  constructor(rpc: Rpc, opts?: { service?: string }) {
    this.service = opts?.service || "simple.PingService";
```

stephenh:
Looks great!

Issue #547, opened 2022-04-08
Feat: proto map generates a Hashmap instead of an object


jayp0521:
currently, 

`map<uint32, X> x = 2;` in proto outputs `{ [key: number]: X }`. 

What are your thoughts of it generating a HashMap instead?

stephenh:
I dunno, not 100% against it, but I think the existing output is ~slightly more idiomatic JS/TS where "maps" are, IMO anyways, almost always just objects with keys and not the actual `Map` instances.

If you want to add an option for it, that seems fine; I'm a little concerned about having two types of `map` output being a little tedious in the code, but I'd merge a PR if you put one together.

Thanks!

stephenh:
This was just fixed in #686 and will be in a release soon; thanks @moznion !

PR #686, opened 2022-10-15
feat: support `useMapType` option

moznion:
This option controls the way to generate the code for protobuf's `map<key_type, value_type>`:

- true: it generates the code with using JavaScript Map type like `Map<key_type, value_type>`
- false (default): it generates the code with key-value pair like `{[key: key_type]: value_type}`

This fixes https://github.com/stephenh/ts-proto/issues/547


stephenh (approved):
Looks great, thanks @moznion !

moznion:
I appreciate your work, @stephenh!

PR #688, opened 2022-10-19
add a no-tags option to the type-registry

fizx:
The $types are getting in the way of usability of the code we're writing, but the registry itself is useful.


src/types.ts
```diff
@@ -688,15 +692,15 @@ export function responseType(
   methodDesc: MethodDescriptorProto,
   typeOptions: { keepValueType?: boolean; repeated?: boolean } = {}
```
stephenh, line 693: Looks like the `typeOptions` param is not used anymore? Lets just drop it if so?

src/generate-services.ts
```diff
@@ -121,11 +121,11 @@ function generateRegularRpcMethod(
...
   let encode = code`${rawInputType}.encode(request).finish()`;
-  let decode = code`data => ${outputType}.decode(new ${Reader}(data))`;
+  let decode = code`data => ${rawOutputType}.decode(new ${Reader}(data))`;
 
-  if (options.useDate && rawOutputType.toString().includes("Timestamp")) {
-    decode = code`data => ${utils.fromTimestamp}(${rawOutputType}.decode(new ${Reader}(data)))`;
-  }
+  // if (options.useDate && rawOutputType.toString().includes("Timestamp")) {
+  //   decode = code`data => ${utils.fromTimestamp}(${rawOutputType}.decode(new ${Reader}(data)))`;
+  // }
```
stephenh, line 128: I assume this is related to https://github.com/stephenh/ts-proto/pull/689 , just checking that you wanted to include commenting out this code in this PR?

stephenh (commented):
I think other than the `*.bin` issue this PR lgtm! Let me know if running `yarn proto2bin` with dockers gets them fixed up. If not, we can either figure out how to get the M1 docker output consistent, or I can just revert the *.bin changes directly on this branch/something like that.

PR #689, opened 2022-10-19
fix: Adding a failing regression test for wrapper types

fizx:
If a GRPC service returns a wrapper type (e.g. StringValue), everything breaks.  This used to work in 1.110.2 at least.  I'm investigating the cause.  Feels related to #534, but the solution is quite unclear as yet.

```
return promise.then((data) => boolean | undefined.decode(new _m0.Reader(data)));
```


fizx:
So #534 introduced a unique way of handling Timestamps.  That incredibly complicates things and I believe should be reverted.

The problem at the moment is that while we're great at `StringValue => string | undefined` if the message in question is a field of another message, doing this magic on GRPC requests or responses is completely unaccounted for in the code.

Given the current state, we should default to "no magic", rather than "deeply broken magic."

fizx:
Also, I have no idea why the test generation is inconsistent with upstream. 😬

stephenh:
> Also, I have no idea why the test bin generation is inconsistent with upstream.

I haven't looked into the rest of this yet, but that's typically due to different versions of `protoc` on host machines. To avoid that, the `yarn proto2bin` target runs fixed version of `protoc` via docker, so you shouldn't see any `.bin` changes after running that.

Hopefully that works?

stephenh:
@fizx fwiw I pushed the `yarn proto2bin` updates to this PR and every is clean/passes now; so thanks! Going to go ahead and merge this.

PR #691, opened 2022-10-20
feat: support `useReadonlyTypes` option

### LINKED ISSUE #163: Option for readonly types
Would it be possible to add an option to generate the types with readonly?

For example instead of generating this:

```ts
export interface GetFullGraphResponse {
  nodes: GraphNode[];
  edges: GraphEdge[];
}
``` 
I would like to have this generated:

```ts
export interface GetFullGraphResponse {
  readonly nodes: readonly GraphNode[];
  readonly edges: readonly GraphEdge[];
}
``` 

The reason behind this is that our team is working in a functional style and therefore trying to keep everything immutable as far as possible. For example if we have a function to generate an array of items, that array is always readonly. So we cannot assign our function's return values directly to ts-proto generated types today .

I'm not sure it it would be possible to have readonly types in all cases but if it is feasible and in line with the goals of this project I may be able to work on a PR for it.

jonaskello:
Fixes #163 

This PR will add the option `useReadonlyTypes`. When set to `true` this option will generate code like this:

```ts
export interface Entity {
  readonly intVal: number;
  readonly stringVal: string;
  readonly intArray: readonly number[];
  readonly stringArray: readonly string[];
  readonly subEntity: SubEntity | undefined;
  readonly subEntityArray: readonly SubEntity[];
  readonly optionalIntVal?: number | undefined;
  readonly fieldMask: readonly string[] | undefined;
  readonly listValue: ReadonlyArray<any> | undefined;
  readonly structValue: { readonly [key: string]: any } | undefined;
}
```


jonaskello:
@stephenh I went with the ` as any` approach as discussed in #163, do you think that is OK?

stephenh (approved):
@jonaskello this looks great! Thanks for the PR, and also for testing the various `google.protobuf.*` well known types.

jonaskello:
@stephenh Thank you for the quick review and merge. Also, thanks for creating this awesome package in the first place. It really makes a huge difference for working with protobuf in typescript :-).

PR #699, opened 2022-11-09
fix: Extend `global.Error` to avoid import collisions with Error proto msgs

MorganIsBatman:
We have an `Error` proto (I assume reasonably common) that causes typescript issues when imported into the proto file of a service definition, as the `export class GrpcWebError extends Error` added to the output .ts files for the service now references the imported Error type, not javascript's Error class.

Change here is to simply use `global.Error` to avoid the issue


stephenh:
Ah yeah, @MorganIsBatman this is great! We've definitely that this issue in the past, with `Error` message types overlapping with `Error`-the-error.

Can you run `yarn bin2ts` to re-generate our integration suite's code with your change?

That will get the build passing + be a regression against the prefix accidentally going away in the future.

MorganIsBatman:
Oh! sorry @stephenh - I missed that step in my haste.  

I've run through building the tests and running them via both the standard and :local variants on my intel Mac and M1 but both have a few tests fail for reasons I'm not totally understanding.  They seem to fail when run on the master branch too?

I've committed the results of running `yarn build:test:local` which I see has re-generated the integration suite, but I would love any advice you may have on resolving the broken tests? 🙏

stephenh:
Ah yeah, @MorganIsBatman didn't have a chance to look in to this now, but sorry, the issue was an admittedly non-obvious foot gun that the `:local`-suffixed targets were not really meant as "run these version locally" but more like "our legacy non-docker build setup".

The non-/pre-docker setup lead to build output flakiness b/c each contributors version of `protoc` would have _slightly_ different output (even for very trivial things like the comments of the core protobuf types being updated across protoc version).

I went ahead and pushed the `yarn bin2ts`/docker-based output to this PR and it all looks great, so I'll go ahead and merge.

I'm also going to make a follow up PR that removes the `:local` targets; we'd kept them with the idea that the docker setup was new/wip, but they've outlived their usefulness at this point.

Thanks!

MorganIsBatman:
You're awesome! Thanks for the quick turnaround and for taking care of that last step for me @stephenh

PR #704, opened 2022-11-15
feat: change channel options to client options in generate grpc/js

### LINKED ISSUE #693: grpc-js generated client constructor missing some allowed options
Small type generation bug (I hope to contribute a fix, but wanted to make sure to open an issue in case I forget!). For grpc-js generated clients, the following constructor [is used](https://github.com/stephenh/ts-proto/blob/main/src/generate-grpc-js.ts#L239):

```
new (
  address: string,
  credentials: ${ChannelCredentials},
  options?: Partial<${ChannelOptions}>,
): ${serviceDesc.name}Client;
```

However, `options` here is not correct: it should use the `ClientOptions` type (exported from grpc-js) instead, which is `Partial<ChannelOptions>` plus some extra fields that you can pass to a client constructor (see https://grpc.github.io/grpc/node/grpc.Client.html).

Very minor bug in the grand scheme of things and easy to work around since it's just a type bug, but I think an easy fix unless I'm missing something.

Focadecombate:
Changes Channel Options type to Client Options and removes 2 not used imports
Fixes #693

### DIFF
--- a/integration/grpc-js/simple.ts
+++ b/integration/grpc-js/simple.ts
@@ -2,9 +2,9 @@
 import {
   CallOptions,
   ChannelCredentials,
-  ChannelOptions,
   Client,
   ClientDuplexStream,
+  ClientOptions,
   ClientReadableStream,
   ClientUnaryCall,
   ClientWritableStream,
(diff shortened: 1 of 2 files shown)


stephenh:
Looks great, thanks @Focadecombate !

PR #706, opened 2022-11-15
fix: use-readonly-types for oneof unions

jonaskello:
This is a follow up PR to #691 that adds missing `readonly` modifiers for `oneof` fields when the `oneof=unions` option is active.

Previously it would generate code like this:

```ts
export interface Entity {
  readonly intVal: number;
  readonly stringVal: string;
  oneOfValue?: { $case: "theStringValue"; theStringValue: string } | {
    $case: "theIntValue";
    theIntValue: number;
  };
}
```

With this PR it now properly generates like this:

```ts
export interface Entity {
  readonly intVal: number;
  readonly stringVal: string;
  readonly oneOfValue?: { readonly $case: "theStringValue"; readonly theStringValue: string } | {
    readonly $case: "theIntValue";
    readonly theIntValue: number;
  };
}
```


src/main.ts
```diff
@@ -770,13 +772,15 @@ function generateOneofProperty(
...
+      return code`{ ${ctx.options.useReadonlyTypes ? "readonly " : ""}$case: '${fieldName}', ${
+        ctx.options.useReadonlyTypes ? "readonly " : ""
+      }${fieldName}: ${typeName} }`;
     }),
     { on: " | " }
   );
 
   const name = maybeSnakeToCamel(messageDesc.oneofDecl[oneofIndex].name, options);
-  return code`${name}?: ${unionType},`;
+  return code`${ctx.options.useReadonlyTypes ? "readonly " : ""}${name}?: ${unionType},`;
```
stephenh, line 783: @jonaskello wdyt of a helper method like:

```
code`${maybeReadonly(ctx)}${name}...`

function maybeReadonly(ctx: ...): string {
  return useReadonlyTypes ? "readonly " : "";
}
```

If you'd like to try that, that'd be cool, but otherwise lgtm! Feel free to merge and release (just added you as a collaborator so I believe you should be able to do that; lmk if not).

stephenh (approved):
Thanks!

src/main.ts
jonaskello: Yes, good idea that would make the code easier to read. Replaced many ternary expressions with two helpers,  ´maybeReadonly()` and `maybeAsAny()`-

jonaskello:
@stephenh Thanks for the collaborator status :-). Should I use "Squash and merge" when merging or what do you prefer?

PR #710, opened 2022-11-22
feat: conditionally add "Service" to nice-grpc's generated service interface name

krugi:
Considering the fact that [Buf](https://buf.build) is gaining popularity and that their [default lint](https://docs.buf.build/tour/lint-your-api) configuration expects a "Service" suffix, it might be nice to avoid adding it to `nice-grpc`'s interface name automatically.

Also, in Google's Resource Oriented Design documentation, services are suffixed with "Service" ([here](https://cloud.google.com/apis/design/resource_names#resource_name_as_string), `service LibraryService`).

If you prefer, I can implement it using an option (`--ts_proto_opt=niceGrpcServiceSuffix`).


stephenh:
Makes sense! Thanks @krugi !

PR #714, opened 2022-11-26
feat: Bump ts poet for perf increase

gwer:
Pulling ts-poet updates. I ran the tests locally and they are still green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "dataloader": "^1.4.0",
     "object-hash": "^1.3.1",
     "protobufjs": "^6.11.3",
-    "ts-poet": "^6.1.0",
+    "ts-poet": "^6.2.0",
     "ts-proto-descriptors": "1.7.1"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -6735,10 +6735,10 @@ ts-jest@^28.0.5:
     semver "7.x"
     yargs-parser "^21.0.1"
 
-ts-poet@^6.1.0:
-  version "6.1.0"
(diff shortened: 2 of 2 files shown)


stephenh:
Great, thanks!

Fwiw is the cost of calling dprint a large remaining hot spot for you?

If so, ts-poet has a ~cute optimization to use the hash of pre-formatted content to avoid calling dprint just to realize the new output vs. old output on disk hasn't changed.

The logic is handled by this saveFiles method:

https://github.com/stephenh/ts-poet/blob/585fc389a822eb909ddfeed8b30592024d4eb899/src/saveFiles.ts#L46

As a disclaimer, dprint is quick, so I'm not sure if this is worth while or not; I've got another ORM project using this `saveFiles` method and it seems to save ~half a second for ~150 files that are ~30k LOC combined. So, nice, but not super-noticeable for that specific project.

Also to leverage `saveFiles`'s conditional behavior from ts-proto, we'd have to stop sending files unconditionally back (already formatted) in the `CodeGeneratorResponse` and maybe just have ts-proto's `plugin.ts` write the files directly to disk. Which, dunno, should be fine? But is probably not what tooling like buf/protoc expect.

Anyway, just mentioning that as an idea if you wanted to keep optimizing things. But there could definitely be other things to fix/optimize as well.

Thanks again!

gwer:
Hmm, I just keep the hashes of the input files at the time of the last generation. Thus, if the inputs have not changed, then protoc does not start at all.

But this logic is implemented in the protoc startup script. I'm not sure that it can be correctly implemented in the plugin.

stephenh:
Ah nice! Agree that sounds even better.

I guess for my ORM use case, one of the inputs is the database schema, so its harder to have that has an input to hash. That and handling incremental changes, i.e. 99 tables are all the same, but 1 is different, so the overall schema hash would have changed, but only a handful of individual output files will change. :thinking:

PR #721, opened 2022-12-09
fix: Add functionality for grpc camel case to respect splitting by word

iangregsondev:
I am using the exact same functionality found in the grpc nodejs library.

I also added a small test.

It now converts GetAPIKey -> getApiKey

Before the change it would be -> getAPIKey

This obviously fails inside the grpc nodejs library.

I have run the tests and of course that works but I was unable to run

```
yarn build:test 
```

I am on a M1 and get 

```
 => ERROR [7/8] RUN protoc --version                                                                                                          0.2s 
------                                                                                                                                             
 > [7/8] RUN protoc --version:
#0 0.144 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
------
failed to solve: executor failed running [/bin/sh -c protoc --version]: exit code: 255
error Command failed with exit code 17.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 17.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I know that this issue was raised and fixed previously but it doesn't work for me :-(


iangregsondev:
@stephenh Could you take a look for me, please? Thanks.

iangregsondev:
There is a failing check :-( Not sure what that is.

iangregsondev:
Oh, I see the checks are all passing now, but before the 16.x was failing, but all seems good now.

iangregsondev:
This is where the grpc library does its conversion.


https://github.com/grpc/grpc-node/blob/f457a9bfa102fc4c7a7213ac0f926346baa49036/packages/proto-loader/src/index.ts#L19

and

https://github.com/grpc/grpc-node/blob/f457a9bfa102fc4c7a7213ac0f926346baa49036/packages/proto-loader/src/index.ts#L242

stephenh:
@iangregsondev I just added a `.yarnrc.yml` file that I think will fix the docker issue for you:

https://github.com/stephenh/ts-proto/blob/main/.yarnrc.yml

Can you try running/rebasing on `main` and see if that works?

package.json
```diff
@@ -61,6 +62,7 @@
   "dependencies": {
     "@types/object-hash": "^1.3.0",
     "dataloader": "^1.4.0",
+    "lodash": "^4.14.191",
```
stephenh, line 65: Given `lodash` is kind of big, can we depend only on the camelcase package:

https://www.npmjs.com/package/lodash.camelcase

src/case.ts
```diff
@@ -32,3 +33,7 @@ export function capitalize(s: string): string {
 export function camelCase(s: string): string {
   return s.substring(0, 1).toLowerCase() + s.substring(1);
 }
+
+export function camelCaseGrpc(s: string): string {
```
stephenh, line 37: Cool, I like wrapping the `lodashCamelCase` in this `camelCaseGrpc` as a call out that we're doing this specifically to match the GRPC semantics.

stephenh (approved):

src/case.ts
iangregsondev: Fixed, I think :-) Take a look.

iangregsondev:
> 

This did not work I am afraid.

iangregsondev:
Oh Wait, I think I updated the wrong tsconfig. Let me have another play :-)

iangregsondev:
@stephenh Sorry mate, no go, it causes me issues if import lodash.camel as I have to enable esModuleInterop but all tests start to fail.

I am probably missing something :-(

Help ?

Would this be a major concern as this is a build tool anyway and keep the size down for front end use is probably not a concern here.

Let me know your feedback, maybe you see the issue ?

Failing that I could extract the lodash internal implementation :-)

Speak soon.

stephenh:
Ah damn, yeah it's been awhile since I've fought with esModuleInterop but
yeah it can be finicky...

I dunno, I don't actually hate the idea of just copy/pasting in their camel
case; assuming that it's a short/self-contained method, yeah let's try that?

Otherwise yeah I'm fine with the full lodash dependency; thanks!

iangregsondev:
The internal implementation for camelCase inside of lodash was a mess :-)

A new library found that is a native typescript library and only 1.4K in size compared to lodash which was 71k :-)

It also supports precisely what we need to do.

Also updated the comment to remove the reference to lodash as we are no longer using it.

iangregsondev:
@stephenh Sometimes the build seems to fail, for example, before it failed all of them :-(

Maybe an issue of Github Actions, I don't think its anything I have done. I had this before.

```
Error: 14 UNAVAILABLE: Stream refused by server
```

Could you force a new build please, I don't think I have rights.

iangregsondev:
@stephenh I updated a comment and it forced a build :-) It has now passed.

Over to you :-) 

Cheers.

stephenh:
This looks great! Thanks @iangregsondev !

Hoppingmad9:
I believe this change breaks service names that use snake_case.

With version 1.135.0 the following proto
```
service General
{
    rpc general_analysis_calculation(GeneralInputs) returns (GeneralOutputsList) {}
}
```
generates
```ts
export interface GeneralServer extends UntypedServiceImplementation {
  general_analysis_calculation: handleUnaryCall<GeneralInputs, GeneralOutputsList>;
}

export interface GeneralClient extends Client {
  general_analysis_calculation(
    request: GeneralInputs,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
  general_analysis_calculation(
    request: GeneralInputs,
    metadata: Metadata,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
  general_analysis_calculation(
    request: GeneralInputs,
    metadata: Metadata,
    options: Partial<CallOptions>,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
}
```
maintaining the snake_case (general_analysis_calculation).

Then with version 1.135.1 the same proto generates
```ts
export interface GeneralServer extends UntypedServiceImplementation {
  generalAnalysisCalculation: handleUnaryCall<GeneralInputs, GeneralOutputsList>;
}

export interface GeneralClient extends Client {
  generalAnalysisCalculation(
    request: GeneralInputs,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
  generalAnalysisCalculation(
    request: GeneralInputs,
    metadata: Metadata,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
  generalAnalysisCalculation(
    request: GeneralInputs,
    metadata: Metadata,
    options: Partial<CallOptions>,
    callback: (error: ServiceError | null, response: GeneralOutputsList) => void,
  ): ClientUnaryCall;
}
```
which has converted the name to camelCase (generalAnalysisCalculation).

I'm using the following options when generating the TypeScript.
```ts
const PROTO_DIR = path.join(__dirname, '../proto_files')
const MODEL_DIR = path.join(__dirname, '../src/models/protos')
const PROTOC_PATH = path.join(__dirname, '../node_modules/grpc-tools/bin/protoc')
const PLUGIN_PATH = path.join(__dirname, '../node_modules/.bin/protoc-gen-ts_proto')
const protoConfig = [
  `--plugin=${PLUGIN_PATH}`,

  // https://github.com/stephenh/ts-proto/blob/main/README.markdown
  '--ts_proto_opt=outputServices=grpc-js,env=node,useOptionals=messages,exportCommonSymbols=false,esModuleInterop=true,snakeToCamel=false',

  `--ts_proto_out=${MODEL_DIR}`,
  `--proto_path ${PROTO_DIR} ${PROTO_DIR}/*.proto`
]
execSync(`${PROTOC_PATH} ${protoConfig.join(' ')}`)
```

It's not a huge issue as a quick find/replace across the workspace fixes it, and when building the JS my compiler suggested the "new" function name as I guess snake/camel difference was easy enough to detect.

stephenh:
@Hoppingmad9 FWIW from @iangregsondev 's fix in this PR, I had assumed the method names of `generalAnalysisCalculation` had to line up with what the grpc-js node library assumed, b/c if we got the name translation wrong (i.e. `getAPIKey` instead of `getApiKey`) things would break (i.e. that's why he opened this PR and fixed it).

But your assertion is that grpc-js works fine with `general_analysis_calculation` names? Did you also have to configure it with a similar "snake to camel false" config option?

Hoppingmad9:
We don't do any config for grpc-js.

So
 - our proto file
 -  the C++ code the requests go to
 - my TypeScript code that references the ts-proto generated code 

all use `general_analysis_calculation`.

Everything worked when ts-proto generated it as snake_case.
Then when I updated ts-proto and generated it with camelCase it broke, but I only had to rename the functions in the TypeScript code to make it all work again.
So camelCase in typescript and snake_case in the proto and C++ works, but this doesn't seem like behaviour that should be relied upon.

I'm happy to change our code/proto to use camelCase everywhere so don't worry about this for us.

PR #727, opened 2022-12-11
refactor: export createBaseMessage function

paralin:
It's useful to be able to construct a message:

const message = createBaseMessage()

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -7,7 +7,7 @@ export interface SimpleMessage {
   numberField: number;
 }
 
-function createBaseSimpleMessage(): SimpleMessage {
+export function createBaseSimpleMessage(): SimpleMessage {
   return { numberField: 0 };
 }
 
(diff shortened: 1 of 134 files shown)


paralin:
I guess it's possible to do MessageType.fromPartial({}) as well. Up to you if you'd like to merge this @stephenh

paralin:
Meh, I'm OK with fromPartial({}).

stephenh:
Cool, yeah, I'm good with sticking with `fromPartial`.

Fwiw one of the "ts-proto 2.x" ideas is to remain `fromPartial` to something "less solution-oriented" (i.e. that it accepts partial types), to be something like just `MessageType.from` or `MessageType.create`.

paralin:
@stephenh I think it might make sense to add MessageType.create today, accepting an optional parameter which, if set, is passed to fromPartial.

```
  create<I extends Exact<DeepPartial<MyMessage>, I>>(obj?: I): MyMessage {
    return MyMessage.fromPartial(obj || {})
  }
```

Thoughts?

stephenh:
@paralin yep, sgtm

paralin:
@stephenh submitted here: https://github.com/stephenh/ts-proto/pull/760

Issue #636, opened 2022-08-03
grpc-web messages have an extra toObject method


lededje:
I'm using redux to manage state and need the responses to be serializeable. I found the flag to change dates into strings which solved half the issue, now it complains that toObject is in every response but it's not in the generated types. How best should I work around this?

stephenh:
@lededje normally ts-proto `decode`-d data does not have any methods, so would be perfect to put into redux, however, for grpc-web, we currently have to add it here:

```
  responseType: {
    deserializeBinary(data: Uint8Array) {
      return {
        ...DashCred.decode(data),
        toObject() {
          return this;
        },
      };
    },
```

Because of how the improbable-eng/grpc-web library works.

If you can play with the generated code, and get a `deserializeBinary` working that doesn't need this `toObject` hack, that'd be great!

(Maybe there is a feature/flag in the the improbable-eng library that means we don't need `toObject`, or maybe you could submit a PR to improbable-eng that would make it detect `toObject`-is-not-available and just use the return value of `deserializeBinary` as-is.)

(Also, as a disclaimer, this `toObject` approach is a little old at this point, so maybe it could be something that is already fixed / not necessary in the latest versions of improbable-eng/grpc-web.)

The wrinkle is that I'm not personally using grpc-web at this point, so you'll have to poke around a bit to figure out a fix, but let us know what you find! Thanks!

lededje:
Thanks for the explaination. I think I'm going to stick with a function that strips that function out when I get a response.

stephenh:
Hey @lededje sure, np, I'll leave the issue open just for others to find & maybe someone can find a way to fix it.

PR #728, opened 2022-12-11
fix: return grpc-web response without toObject method

relaxgo:
fix stephenh/ts-proto#636


stephenh:
Ah great! Thanks @relaxgo !

PR #731, opened 2022-12-13
feat: RPC: add useAbortSignal option

### LINKED ISSUE #730: RPC: use AbortController and AbortSignal to cancel requests
Related to #665 but not exactly the same.

TypeScript/JavaScript has AbortController now as a general way of canceling things:

https://developer.mozilla.org/en-US/docs/Web/API/AbortController

Here's an example of how it's used in yamux in javascript:

https://github.com/ChainSafe/js-libp2p-yamux/blob/c396f8c1b99f3c68104c894a1ac88a805bff68a3/src/muxer.ts#L49
https://github.com/ChainSafe/js-libp2p-yamux/blob/c396f8c1b99f3c68104c894a1ac88a805bff68a3/src/muxer.ts#L215

I propose adding an optional parameter to the RPC methods: `AbortSignal` - 

```
const abortController = new AbortController()
const responsePromise = rpcClient.DoSomething(request, abortController.signal)
abortController.abort()
// responsePromise is rejected w/ Aborted
```

This can be used to stop a request before it completes.

paralin:
Adds a new option "useAbortSignal" which adds an optional AbortSignal parameter to RPC functions. AbortController and AbortSignal are built-ins in both Node.JS and all web browsers, which implement aborting long-lived processes.

For example:

```typescript
const abortController = new AbortController()
const responsePromise = rpcClient.DoSomething(request, abortController.signal)
// abort the RPC call early
abortController.abort()
```

Fixes #730


paralin:
@stephenh This should be good to go now. I wonder too about passing an AbortSignal on the server side.

PR #737, opened 2022-12-16
fix: Initialize fields when decoding. (#736)

### LINKED ISSUE #736: Invalid code generation for `decode` with `useOptionals=all` + `initializeFieldsAsUndefined=false`
When using the options `useOptionals=all` and `initializeFieldsAsUndefined=false` in combination, `ts-proto` generates the following code:

```test.proto
message Test {
  repeated string values = 1;
}
```

```test.ts
function createBaseTest(): Test {
  return {};
}

export const Test = {
  decode(input: _m0.Reader | Uint8Array, length?: number): Test {
    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
    let end = length === undefined ? reader.len : reader.pos + length;
    const message = createBaseTest();
    while (reader.pos < end) {
      const tag = reader.uint32();
      switch (tag >>> 3) {
        case 1:
          message.values!.push(reader.string());  // TypeError: Cannot read properties of undefined (reading 'push')
          break;
        default:
          reader.skipType(tag & 7);
          break;
      }
    }
    return message;
  }
};
```

This fails at runtime with `TypeError: Cannot read properties of undefined (reading 'push')`.

To fix this, the affected code snippet should probably not use a non-null assertion operator and instead look as follows:

```test.ts
          message.values = message.values ?? [];
          message.values.push(reader.string());
```

This is also true for maps, which might either use `new Map()` or `{}`.

fxb:
Fixes #736

Currently missing tests, which I will add next week.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -944,32 +944,39 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
 
     // and then use the snippet to handle repeated fields if necessary
     if (isRepeated(field)) {
-      const maybeNonNullAssertion = ctx.options.useOptionals === "all" ? "!" : "";
+      const initialValue = isMapType(ctx, messageDesc, field) ? (ctx.options.useMapType ? "new Map()" : "{}") : "[]";
+      const maybeInitalize =
+        ctx.options.useOptionals === "all" ? `message.${fieldName} = message.${fieldName} ?? ${initialValue}` : "";
 
       if (isMapType(ctx, messageDesc, field)) {
         // We need a unique const within the `cast` statement
         const varName = `entry${field.number}`;
 
         const valueSetterSnippet = ctx.options.useMapType
-          ? `message.${fieldName}${maybeNonNullAssertion}.set(${varName}.key, ${varName}.value)`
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -944,32 +944,38 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
 
     // and then use the snippet to handle repeated fields if necessary
     if (isRepeated(field)) {
-      const maybeNonNullAssertion = ctx.options.useOptionals === "all" ? "!" : "";
+      const initialValue = isMapType(ctx, messageDesc, field) ? (ctx.options.useMapType ? "new Map()" : "{}") : "[]";
+      const maybeInitalize = `message.${fieldName} = message.${fieldName} ?? ${initialValue}`;
```
fxb, line 948: Just noticed that it's always initializing now, but this code can be omitted if `ctx.options.useOptionals !== "all"`

fxb: Fixed and force-pushed.

davidzeng0:
Fixed in https://github.com/stephenh/ts-proto/pull/802

stephenh:
Great, thanks @davidzeng0 ! Fyi @fxb going to close this out then.

Issue #739, opened 2022-12-20
google.protobuf.Struct is invalid when generated with useMapType=true


roboslone:
Generated code:
```typescript
  wrap(object: { [key: string]: any } | undefined): Struct {
    const struct = createBaseStruct();
    if (object !== undefined) {
      Object.keys(object).forEach((key) => {
        struct.fields[key] = object[key]; // <----------------
      });
    }
    return struct;
  },

  unwrap(message: Struct): { [key: string]: any } {
    const object: { [key: string]: any } = {};
    Object.keys(message.fields).forEach((key) => {
      object[key] = message.fields[key]; // <----------------
    });
    return object;
  },
```

```
google/protobuf/struct.ts:163:9 - error TS7052: Element implicitly has an 'any' type because type 'Map<string, any>' has no index signature. Did you mean to call 'struct.fields.set'?

163         struct.fields[key] = object[key];
            ~~~~~~~~~~~~~~~~~~

google/protobuf/struct.ts:172:21 - error TS7052: Element implicitly has an 'any' type because type 'Map<string, any>' has no index signature. Did you mean to call 'message.fields.get'?

172       object[key] = message.fields[key];
                        ~~~~~~~~~~~~~~~~~~~
```

stephenh:
Fixed by #740, thanks @roboslone !

PR #740, opened 2022-12-20
fix: Fix codegen for google.protobuf.Struct with useMapType=true

roboslone:
Possible fix for #739. I'm not sure how to generate code using `buf` and my fork. Can someone help with this?


roboslone:
@stephenh could you take a look at this? Thanks.

stephenh:
Hey @roboslone yep, this looks good! I'll open a follow up PR with a really simple integration test to make sure this doesn't regress, but otherwise looks great! Thank you!

PR #742, opened 2022-12-28
feat: added bigint force long option

azizghuloum:
Also added simple tests based on string option.


src/main.ts
```diff
@@ -417,6 +417,15 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
     `
   );
 
+  const longToBigint = conditionalOutput(
+    "longToBigint",
+    code`
+      function longToBigint(long: ${Long}) {
```
stephenh, line 423: I was wondering if long.js had a built-in `toBigInt` and didn't find anything, but I did find a github issue where someone linked to their long.js fork that has a `toBigInt`:

https://github.com/consento-org/longfn/blob/main/index.mjs#L654

Wdyt about adapting their `toBigInt` and maybe even `fromBigInt`?

https://github.com/consento-org/longfn/blob/main/index.mjs#L100

Seems like those would surely be faster than going through `toString`s, so seems like a good idea, but also willing to leave it as a follow-up issue if you'd like to ship this as-is.

stephenh (approved):
Nice! This looks great @azizghuloum! 

I was assuming we'd have to wait on protobufjs to support bigint natively, but great call that we can just do the adapting on the way in/out.

I had one question about potentially using more mathy to/from conversion methods, from the longfn repo, but definitely optional and happy to merge this as-is. Just let me know, thanks!

src/main.ts
azizghuloum: Honestly I was just trying to get it to work correctly with zero effort towards optimizing anything.  This, and a few other things in the timestamps can definitely be improved, but I think eventually protobuf's `Reader` and `Writer` will support bigints and drop `Long`.  Regardless, I think it should be merged now and any improvements be done later.

PR #743, opened 2022-12-29
fix: Additional fix for structs with useMapType

stephenh:


integration/use-map-type/use-map-type.proto
```diff
@@ -13,4 +16,5 @@ message Maps {
   map<string, bytes> string_to_bytes = 3;
   map<int64, int64> int64_to_int64 = 4;
   map<string, google.protobuf.Timestamp> map_of_timestamps = 5;
+  google.protobuf.Struct struct = 6;
```
stephenh, line 19: cc @roboslone I verified that `struct.ts` was failing before your PR, and with your PR, the `struct.ts` compiles cleanly.

PR #751, opened 2023-01-04
fix: repeated uint64 fields do not encode properly with bigint option

### LINKED ISSUE #749: --forceLong=bigint doesn't handle array of int64 correctly
```
message XXX {
  repeated uint64 yyy = 4;
}
```

will generate code like:
```
  encode(message: EvmGetHeaderRequest, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
    for (const v of message.blockNumbers) {
      writer.uint64(v);
    }  },
```

which doesn't work since `uint64` can't take bigint

azizghuloum:
fixes #749


stephenh:
This is great, thanks @azizghuloum !

stephenh:
(I also went ahead and added a regression test before merging)

PR #755, opened 2023-01-05
fix: grpc-js timestamp conversion

### LINKED ISSUE #754: Invalid code generation for grpc-js Timestamp conversion
Following proto schema
```
  rpc GetTime(google.protobuf.Empty) returns (google.protobuf.Timestamp);
```
generates following client code with `outputServices=grpc-js`
```
  getTime(request: Empty, callback: (error: ServiceError | null, response: Date) => void): ClientUnaryCall;
```
But actual response type is Timestamp, not Date
```
    client.getTime({}, (error, response) => {
      console.log(response) // { seconds: 1672938038, nanos: 388586128 }
    })
```

It seems the generated service code for deserialization is incorrect.
```
  getTime: {
    path: "/namespace.Service/GetTime",
    requestStream: false,
    responseStream: false,
    requestSerialize: (value: Empty) => Buffer.from(Empty.encode(value).finish()),
    requestDeserialize: (value: Buffer) => Empty.decode(value),
    responseSerialize: (value: Date) => Buffer.from(Timestamp.encode(toTimestamp(value)).finish()),
    responseDeserialize: (value: Buffer) => Timestamp.decode(value),  // <- should be wrapped by fromTimestamp()
  },
```

Also `useDate=false` doesn't generate valid code for serialization.
* true
```
    responseSerialize: (value: Date) => Buffer.from(Timestamp.encode(toTimestamp(value)).finish()),  // correct
```
* string
```
    responseSerialize: (value: string) => Buffer.from(Timestamp.encode(toTimestamp(value)).finish()),  // correct
```
* false
```
    responseSerialize: (value: Timestamp) => Buffer.from(Timestamp.encode(toTimestamp(value)).finish()),  // compile error: shouldn't be wrapped by toTimestamp()
```

I'll open a PR to fix them.

aokht:
Fix #754


stephenh:
This is great! Thanks @aokht !

PR #760, opened 2023-01-10
feat: add create utility function to message defintions

paralin:
Similar to `fromPartial` except it allows an empty argument:

 - create() -> {object}
 - create({partial object}) -> {object}

Calls fromPartial internally.

Callers of the library can now use create() instead of fromPartial({}) which, while functionally equivalent, is much more readable.

Context: https://github.com/stephenh/ts-proto/pull/727#issuecomment-1374975543

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -167,6 +167,10 @@ Along with `encode`/`decode` factory methods:
 
 ```typescript
 export const Simple = {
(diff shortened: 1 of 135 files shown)


stephenh (approved):
Looks great!

Issue #188, opened 2021-01-22
addGrpcMetadata=true without NestJS


jonaskello:
I would like to pass some meta data in my service calls but I am not using NestJS but instead building my own framework. Since the interfaces for service calls do not generate any parameter for metadata this is not really possible with the types generated today. I found the option `addGrpcMetadata=true ` which is exactly what I want but I discovered it does not work unless using NestJS specific codegen. Is there anything stopping this option to be available without the NestJS generation being active? If not I would request it be added for other users too.

stephenh:
Sure, I've moved the lines in config around so that these should work. Disclaimer I didn't add an integration test for your setup, but it should work. Let me know if not.

zouyixiong:
Have you publish this feature in 1.76.0 version ? I got some error output.

``` proto
service DemoService {
    rpc GetBasic (QueryById) returns (BasicInfo) {}
    rpc Insert (Demo) returns (Demo) {}
}
```

```sh
	protoc -I ./proto \
	--plugin=node_modules/ts-proto/protoc-gen-ts_proto \
	--ts_proto_opt=outputEncodeMethods=false,outputJsonMethods=false,outputClientImpl=false,lowerCaseServiceMethods=true,addGrpcMetadata=true, \
	--ts_proto_out=${OUT_DIR} \
    ./proto/*/*.proto
```

output  [ without grpc import in the output file ]

```ts
import { util, configure } from 'protobufjs/minimal';
import * as Long from 'long';

export interface DemoService {
   getBasic(request: QueryById,metadata?: Metadata@grpc): Promise<BasicInfo>;
   insert(request: Demo,metadata?: Metadata@grpc): Promise<Demo>;
}

```

if more infos required, please tell me

stephenh:
@zouyixiong thanks for the report and reproduction; you're right the import wasn't correct. That's been fixed in the latest release.

vojty:
Hello @stephenh , I'm trying to use `addGrpcMetadata=true` without NestJS as well it doesn't seem to be working. The client types are correct, for example:
```
export interface MyClient {
...
  ListRegions(
    request: ListRegionsRequest,
    metadata?: Metadata
  ): Promise<ListRegionsResponse>;
...
}
```
but the actual implementation looks like this:
```
export class MyClientImpl implements MyClient {
...
  ListRegions(request: ListRegionsRequest): Promise<ListRegionsResponse> {
                                        ^^ ----- metadata are missing here
[...]

PR #761, opened 2023-01-19
fix: fixed addGrpcMetadata option

vojty:
This PR should fix this issue https://github.com/stephenh/ts-proto/issues/188

I've removed `addNestjsRestParameter` parameter completely from `src/generate-services.ts` since `NestJS` seems to be handled in `generate-nestjs` file.

- [x] tests


vojty:
Hey, @stephenh would you be so kind and have some time to look at this, please?

integration/generic-metadata/hero.ts
```diff
@@ -287,9 +287,9 @@ export const Villain = {
 };
 
 export interface HeroService {
-  FindOneHero(request: HeroById, metadata?: Foo): Promise<Hero>;
```
stephenh, line 290: I think the `metadataType=Foo...` should probably still be used here/take precedence over `addGrpcMetadata`? Or how do you think using both options at the same time should behave?

integration/generic-metadata/hero.ts
```diff
@@ -303,21 +303,21 @@ export class HeroServiceClientImpl implements HeroService {
     this.FindOneVillain = this.FindOneVillain.bind(this);
     this.FindManyVillain = this.FindManyVillain.bind(this);
   }
-  FindOneHero(request: HeroById): Promise<Hero> {
+  FindOneHero(request: HeroById, metadata?: Metadata): Promise<Hero> {
```
stephenh, line 306: Just checking my understanding, but given this is the `...ClientImpl`, is the idea of exposing `metadata` here that it allows clients to pass in something like `{ userId: 123, authToken: 345 }` into the client calls?

Fwiw/if so, I think historically I'd assumed that sort of metadata would be added by the `Rpc` implementation, instead of having to be passed on every single client method invocation.

But curious to hear if that makes sense/how you're using the `metadata` param on client calls.

stephenh (commented):
Hi @vojty ! Thanks for the PR, just a few questions, thanks!

integration/generic-metadata/hero.ts
vojty: I completely missed `metadataType=Foo` parameter. I think there should be `metadata?: Foo`, I'll fix that and probably add another test case

integration/generic-metadata/hero.ts
vojty: We generally pass a metadata factory to `Rpc` implementation but we have a use case, where the backend implementation expects additional metadata, so the idea here is to be able to override them.

vojty:
I've added 2 separate tests. The expected output:

1. `metadataType=Foo` - this should fix the current behavior and add optional `metadata?: Foo` to `*Impl` methods as well (the metadata parameter is currently present only in the type definition https://github.com/stephenh/ts-proto/blob/main/integration/generic-metadata/hero.ts#L289-L293)
2. `metadataType=Foo,addGrpcMetadata=true` - the same as above
3. `addGrpcMetadata=true` - adds optional `metadata?: Metadata`

PR #763, opened 2023-01-25
Implemented optional visitor pattern for messages

ArthurvdKnaap:
I have a larger message structure that I have to walk through several times, so I implemented on optional [visitor pattern ](https://en.wikipedia.org/wiki/Visitor_pattern ) for messages. would appreciate it if you would accept this to the main branch.

### DIFF
--- a/integration/simple-visitorPattern/simple.ts
+++ b/integration/simple-visitorPattern/simple.ts
@@ -1,0 +1,224 @@
+/* eslint-disable */
+import * as _m0 from "protobufjs/minimal";
+
+export const protobufPackage = "simple";
+
+export interface PBVisitable {
+  visit(visitor: Visitor): void;
+}
+
+export interface Simple extends PBVisitable {
+  /** Name field */
+  name: string;
+  /** Age field */
+  age: number;
+  child: Child[];
+  testField: string;
+  testSecondField: string;
+}
+
+export interface Child extends PBVisitable {
+  name: string;
+  child: Child[];
+}
+
+function createBaseSimple(): Simple {
+  return {
+    visit(visitor: Visitor) {
+      visitor.visitSimple(this);
+    },
+    name: "",
+    age: 0,
+    child: [],
+    testField: "",
+    testSecondField: "",
+  };
+}
+
+export const Simple = {
+  encode(message: Simple, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
+    if (message.name !== "") {
+      writer.uint32(10).string(message.name);
+    }
+    if (message.age !== 0) {
+      writer.uint32(16).int32(message.age);
+    }
+    for (const v of message.child) {
+      Child.encode(v!, writer.uint32(26).fork()).ldelim();
+    }
+    if (message.testField !== "") {
+      writer.uint32(34).string(message.testField);
+    }
+    if (message.testSecondField !== "") {
+      writer.uint32(42).string(message.testSecondField);
+    }
+    return writer;
+  },
+
+  decode(input: _m0.Reader | Uint8Array, length?: number): Simple {
+    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
+    let end = length === undefined ? reader.len : reader.pos + length;
+    const message = createBaseSimple();
+    while (reader.pos < end) {
+      const tag = reader.uint32();
+      switch (tag >>> 3) {
+        case 1:
+          message.name = reader.string();
+          break;
+        case 2:
+          message.age = reader.int32();
+          break;
+        case 3:
+          message.child.push(Child.decode(reader, reader.uint32()));
+          break;
+        case 4:
+          message.testField = reader.string();
+          break;
+        case 5:
+          message.testSecondField = reader.string();
+          break;
+        default:
+          reader.skipType(tag & 7);
+          break;
+      }
+    }
+    return message;
+  },
+
+  fromJSON(object: any): Simple {
+    return {
+      visit(visitor: Visitor) {
+        visitor.visitSimple(this);
+      },
+      name: isSet(object.name) ? String(object.name) : "",
+      age: isSet(object.age) ? Number(object.age) : 0,
+      child: Array.isArray(object?.child) ? object.child.map((e: any) => Child.fromJSON(e)) : [],
+      testField: isSet(object.testField) ? String(object.testField) : "",
+      testSecondField: isSet(object.testSecondField) ? String(object.testSecondField) : "",
+    };
+  },
+
+  toJSON(message: Simple): unknown {
+    const obj: any = {};
+    message.name !== undefined && (obj.name = message.name);
+    message.age !== undefined && (obj.age = Math.round(message.age));
+    if (message.child) {
+      obj.child = message.child.map((e) => e ? Child.toJSON(e) : undefined);
+    } else {
+      obj.child = [];
+    }
+    message.testField !== undefined && (obj.testField = message.testField);
+    message.testSecondField !== undefined && (obj.testSecondField = message.testSecondField);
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -35,13 +35,6 @@
     "@nestjs/common": "^8.2.2",
     "@nestjs/core": "^8.2.2",
     "@nestjs/microservices": "^8.2.2",
-    "@semantic-release/changelog": "^5.0.1",
```
stephenh, line 38: These should stay.

src/main.ts
```diff
@@ -1304,13 +1325,13 @@ function generateFromJson(ctx: Context, fullName: string, fullTypeName: string,
       } else if (isValueType(ctx, field)) {
         const valueType = valueTypeName(ctx, field.typeName)!;
         if (isLongValueType(field) && options.forceLong === LongOption.LONG) {
-          return code`${capitalize(valueType.toCodeString())}.fromValue(${from})`;
+          return code`${capitalize(valueType.toCodeString([]))}.fromValue(${from})`;
         } else if (isLongValueType(field) && options.forceLong === LongOption.BIGINT) {
           return code`BigInt(${from})`;
         } else if (isBytesValueType(field)) {
-          return code`new ${capitalize(valueType.toCodeString())}(${from})`;
+          return code`new ${capitalize(valueType.toCodeString([]))}(${from})`;
```
stephenh, line 1332: Huh, are these changes required?

stephenh (commented):
Hey @ArthurvdKnaap , sorry took me awhile to get to the code review, but yeah I'm opening to adding support for this. Added a few comments, primarily about using a static method + auto-traversal. Thanks!

integration/simple-visitorPattern/simple.ts
```diff
@@ -0,0 +1,224 @@
+/* eslint-disable */
+import * as _m0 from "protobufjs/minimal";
+
+export const protobufPackage = "simple";
+
+export interface PBVisitable {
+  visit(visitor: Visitor): void;
```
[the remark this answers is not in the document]
ArthurvdKnaap: The philosophy in it's core is conflicting with the Visitor pattern that is an object oriented pattern that uses indirection based on classes. Could you elaborate your idea on how this would function?

integration/simple-visitorPattern/visitorPattern-test.ts
```diff
@@ -0,0 +1,52 @@
...
+    level = 0;
+    buildTree(obj: Simple): string {
+        this.simpleTree = "";
+        obj.visit(this);
+        return this.simpleTree;
+    }
+
+    visitSimple(obj: Simple): void {
+        this.simpleTree = this.simpleTree + obj.name + "\r\n";
+        for (const child of obj.child) {
```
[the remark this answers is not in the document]
ArthurvdKnaap: I would like to leave iteration to the visitor, it is up to the visitor to decide what are the interesting parts of the message. Also I would like to keep it as simple as possible

package.json
ArthurvdKnaap: I got errors on these, there are conflicting dependencies, with "semantic-release"

src/main.ts
ArthurvdKnaap: These are calls to an internal API that  shouldn' t be used. The API has changed, I tried to track down where the error originated from (how I got a different dependency), but after some trying I decided to give the methods the array they needed.

integration/simple-visitorPattern/simple.ts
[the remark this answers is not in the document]
stephenh: Combining the two comment threads...

> it is up to the visitor to decide what are the interesting parts of the message

Imo that is not traditionally what the visitor pattern is, i.e. see the [wikipedia](https://en.wikipedia.org/wiki/Visitor_pattern) implementations:

```java
    @Override
    public void accept(CarElementVisitor visitor) {
        for (CarElement element : elements) {
            element.accept(visitor);
        }
        visitor.visit(this);
    }
```

```python
class Car(CarElement):
    def __init__(self):
        self.elements = [
            Wheel("front left"), Wheel("front right"),
            Wheel("back left"), Wheel("back right"),
            Body(), Engine()
        ]

    def accept(self, visitor):
        for element in self.elements:
            element.accept(visitor)
        visitor.visitCar(self)
```

They all have the parent's `accept` method auto-traversing into the child by auto-calling the child `accept`s within it's own `accept` method.

That is what the visitor pattern fundamentally is, b/c the idea comes from compilers/ASTs/etc. where the number/complexity/types of children is large, non-trivial, and would be easy for consumers to forget/not get right, so the visitor impl handles auto-traversal.

I get it reduces control, and b/c of that visitors can just be complicated to use in practice, but :shrug: afaiu that is just pros/cons of the pattern.

To make it concrete in terms of this PR, if all `accept` does is call `visitor.visitWhateverIAm()`, that's so simple, couldn't this code:

```ts
  visitSimple(obj: Simple): void {
    this.simpleTree = this.simpleTree + obj.name + '\r\n';
    for (const child of obj.child) {
      child.accept(this);
    }
  }
```

Just be written as:

```ts
  visitSimple(obj: Simple): void {
    this.simpleTree = this.simpleTree + obj.name + '\r\n';
    for (const child of obj.child) {
      this.visitChild(child);
    }
  }
```

Where I just manually changed `child.accept` --> `this.visitChild(child)`. If that is all that `child.accept` is going to do, I don't think it's carrying it's weight as an abstraction, so you might as well just call `this.visitChild(child)` directly, and it would work without any changes to ts-proto's output.

Personally, given you explicitly want to control iteration, that's my recommendation for what you should do. That said, if you really want to see visitors-the-pattern implemented in ts-proto, that's great too and we can keep discussing, but I think, for me, having the visitor control iteration will be a required requirement, b/c again w/o that, why bother/you don't really need ts-proto's help to call `child.accept(this)` instead of `this.visitChild(child)`.

> The philosophy in it's core is conflicting with the Visitor pattern that is
> an object oriented pattern that uses indirection based on classes.

I agree that the visitor pattern is typically implemented as a polymorphic `accept` method, but that is really only required for situations that have _hetergenous_ collections, i.e. something like:

```
class FruitBasket {
  fruits = [new Apple(), new Banana()]
  accept(visitor) {
    // this is heterogeneous, we don't know to call f.acceptApple or f.acceptBanana
    // so we just call f.accept and let OO figure it out
    this.fruits.forEach(f => f.accept(visitor))
  }
}
```

However, protobuf fundamentally does not support heterogeneous collections; when we have a repeated list of Apples in a message, we know all of them are Apples. When we have a repeated list of Bananas in a message, we know all of them are Bananas. Because of this, we don't have the "I don't know whether to call `acceptApple` or `acceptBanana`, and so our generated code can "just know" which static method (Accept.accept or Banana.accept) to call and it will always be correct.

ArthurvdKnaap:
Hi @stephenh ,
I implemented the visitor pattern as first descibed by the "great four",as they used to be called, in their book "Design patterns", of which I still have a hardcopy, and used as reference.
I actually have a polymorphic structure, and it is true that protobuf doesn't support this, but being new to Typescriipt, I thought I could build my object polymorphic model directly on top of the generated code, which would allow for a lot less boiler plate code. I now have come to the conclusion that I cannot use this approach to provide a clean interface, and have a separate object model.
I will not develop this fork anymore, but use the main branch for generating code. Thank you for your valuable input, and for your work on ts-proto.

stephenh:
> I implemented the visitor pattern as first descibed by the "great four"

Yep, I have that book on my shelf too, but admittedly I have not read it in 20 years... :-/ :-)

> I will not develop this fork anymore, but use the main branch for
> generating code.

Cool, definitely no problem, thanks for diving in and spiking out your thoughts!

PR #767, opened 2023-01-31
perf: generate switch statement for oneof union encode

antsluts:
Updates the `encode` method to generate a `switch` statement for oneof unions.  These are currently generating as `if` statements for every possible `$case` so not great for performance as the number of cases grows.

In our usage scenario we have a oneof with 160+ cases, which is currently generating 160+ `if` statements and during `encode` each is evaluated!!!

From benchmarking against our internal `.proto` message definition the `switch` version achieves ~2.8x better performance over the current `if` version.

**Before**

```typescript
export const Value = {
  encode(message: Value, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
    if (message.kind?.$case === "null_value") {
      writer.uint32(8).int32(message.kind.null_value);
    }
    if (message.kind?.$case === "number_value") {   // not an 'else if' so evaluate regardless if matched already
      writer.uint32(17).double(message.kind.number_value);
    }
    if (message.kind?.$case === "string_value") {   // not an 'else if' so evaluate regardless if matched already
      writer.uint32(26).string(message.kind.string_value);
    }
    ...
```

**After**

```typescript
export const Value = {
  encode(message: Value, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
    switch (message.kind?.$case) {
      case "null_value":
        writer.uint32(8).int32(message.kind.null_value);
        break;
      case "number_value":
        writer.uint32(17).double(message.kind.number_value);
        break;
      case "string_value":
        writer.uint32(26).string(message.kind.string_value);
        break;
     ...
```

### DIFF
--- a/integration/oneof-unions-snake/google/protobuf/struct.ts
+++ b/integration/oneof-unions-snake/google/protobuf/struct.ts
@@ -244,23 +244,25 @@ function createBaseValue(): Value {
 
 export const Value = {
   encode(message: Value, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
-    if (message.kind?.$case === "null_value") {
-      writer.uint32(8).int32(message.kind.null_value);
-    }
-    if (message.kind?.$case === "number_value") {
-      writer.uint32(17).double(message.kind.number_value);
-    }
-    if (message.kind?.$case === "string_value") {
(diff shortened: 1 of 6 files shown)


stephenh:
Looks great, thanks @antsluts !

PR #777, opened 2023-02-10
feat: implementation of useAbortSignal option for grpc-web

hersentino:
https://github.com/stephenh/ts-proto/pull/731 introduces the `useAbortSignal` option but it's currently not implemented for grpc-web.

This PR add useAbortSignal for grpc-web, fully working with unary/invoke

### DIFF
--- a/integration/grpc-web-no-streaming-observable/example.ts
+++ b/integration/grpc-web-no-streaming-observable/example.ts
@@ -321,7 +321,11 @@ export const Empty = {
  * but with the streaming method removed.
  */
 export interface DashState {
-  UserSettings(request: DeepPartial<Empty>, metadata?: grpc.Metadata): Observable<DashUserSettingsState>;
+  UserSettings(
+    request: DeepPartial<Empty>,
+    abortSignal?: AbortSignal,
+    metadata?: grpc.Metadata,
+  ): Observable<DashUserSettingsState>;
 }
 
 export class DashStateClientImpl implements DashState {
@@ -332,8 +336,12 @@ export class DashStateClientImpl implements DashState {
     this.UserSettings = this.UserSettings.bind(this);
   }
 
-  UserSettings(request: DeepPartial<Empty>, metadata?: grpc.Metadata): Observable<DashUserSettingsState> {
-    return this.rpc.unary(DashStateUserSettingsDesc, Empty.fromPartial(request), metadata);
+  UserSettings(
+    request: DeepPartial<Empty>,
+    abortSignal: AbortSignal | undefined,
+    metadata?: grpc.Metadata,
+  ): Observable<DashUserSettingsState> {
+    return this.rpc.unary(DashStateUserSettingsDesc, Empty.fromPartial(request), abortSignal, metadata);
   }
 }
 
@@ -373,6 +381,7 @@ interface Rpc {
   unary<T extends UnaryMethodDefinitionish>(
     methodDesc: T,
     request: any,
+    abortSignal: AbortSignal | undefined,
     metadata: grpc.Metadata | undefined,
   ): Observable<any>;
 }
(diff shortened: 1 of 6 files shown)


hersentino:
I'm trying to figure out why `build (16.x)` is failing

stephenh:
Hi @hersentino , thanks for the PR! I think this looks great!

> I'm trying to figure out why build (16.x) is failing

Ah yeah, sorry for not replying sooner but oddly enough CircleCL will cancel all jobs once the first one fails, so a lot of times the issue will look (say) specific to the 16.x build, but really it just happened to fail first, and the issue does actually exist in all versions of node.

I think your changes look good; as a regression test, can you find an example `integration/*` directory that is using grpc-web and add the `useAbortSignal=true` method to its `parameters.txt`? After running `yarn bin2ts`, you should see the generated code from your changes, which we include in PRs + as part of CI. Thanks!

src/generate-grpc-web.ts
```diff
@@ -332,7 +378,20 @@ function createInvokeMethod(ctx: Context) {
               }
             },
           });
-          observer.add(() => client.close());
+          observer.add(() => {
+            if (!observer.closed) return client.close()
+          });
+
+          ${
```
stephenh, line 385: These inline ternaries are _just_ long enough to be a little hard to ready; could you pull the out to look like:

```
const maybeAbortSignal = useAbortSignal ? code`
  const abortHandler ...
` ? "";
```

And then inline / like right you can do just `${maybeAbortSignal}`.

stephenh:
Fwiw @hersentino I think landing this PR is still great; fwiw I went to google "improbable-eng abort signal" to sanity check how it works, and I came across nice-grpc-web:

https://www.npmjs.com/package/nice-grpc-web

Which is a sister project/built on top of ts-proto.

I was scanning their impl and nothing stuck out as me as something we should borrow/replicate for your PR:

https://github.com/deeplay-io/nice-grpc/blob/master/packages/nice-grpc-web/src/client/createUnaryMethod.ts

Which is cool, we can ship your PR as-is, but just thought I'd mention it, since they are actively using ts-proto + grpc-web, it might be worth checking out.

hersentino:
>  can you find an example integration/* directory that is using grpc-web and add the useAbortSignal=true method to its parameters.txt ?

I chose `integretion/grpc-web-no-streaming-observable` because it contains unary method. But might it be better if I create a new example directory just for useAbortSignal option ?  It could test the 3 methods `createPromiseUnaryMethod`, `createObservableUnaryMethod` and `createInvokeMethod`

hersentino:
> since they are actively using ts-proto + grpc-web, it might be worth checking out.

It seems to be handle here [execute
](https://github.com/deeplay-io/abort-controller-x/blob/master/src/execute.ts).

They don't use the `close` method from the client. I'm trying to understand how it works without, I don't a have lot a time right now but I should test `nice-grpc-web` to fully understand the behavior

hersentino:
I'm sorry @stephenh I forgot to ping you

integration/grpc-web-no-streaming-observable/example.ts
```diff
@@ -321,7 +321,11 @@ export const Empty = {
  * but with the streaming method removed.
  */
 export interface DashState {
-  UserSettings(request: DeepPartial<Empty>, metadata?: grpc.Metadata): Observable<DashUserSettingsState>;
+  UserSettings(
+    request: DeepPartial<Empty>,
+    abortSignal?: AbortSignal,
```
stephenh, line 326: @hersentino this seems fine, but it's probably a breaking API change to current codes that are passing `foo.UserSettings(request, metadata)`?

Can we push the `abortSignal` last so that it's not a breaking change?

hersentino: Yep no problem :). 

I initially put it in second argument because in this it was set as 2nd argument in this [PR](https://github.com/stephenh/ts-proto/pull/731/files#diff-c74bac0693e72660a78a574521c3dadd5dfa74fecb9d2dba513c7407897cb39bR65)

So I had to also push abortSignal to last argument for the function `generateService` inside `src/generate-services.ts`.  Not sur if this modification break something somewhere else.

stephenh:
@hersentino looks great! Thanks!

PR #779, opened 2023-02-17
feat: `removeEnumPrefix` option

superclaw:
Based on PR by @Toofifty: https://github.com/stephenh/ts-proto/pull/477


README.markdown
```diff
@@ -352,6 +352,12 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=unrecognizedEnum=false` enums will not contain an `UNRECOGNIZED` key with value of -1.
 
+- With `--ts_proto_opt=removeEnumPrefix=all` or `members` generated enums will have the enum name removed from both members and values (using `all`), or just members (using `members`)
```
stephenh, line 355: @superclaw I know this is asking for more work, but do you mind simplifying this PR by changing this to just support `removeEnumPrefix=true`?

So remove the all/members distinction, and really just keep the "members" where we rename the keys but leave the values alone.

And we can have just a single integration test, `remove-enum-prefix`.

(That was my feedback on the original PR that I'd been waiting to have implemented.)

superclaw: I see, seems I did not understand last time :) Ok, I'll do it

stephenh:
Looks great @superclaw ! Thanks for the PR!

PR #783, opened 2023-02-23
docs: outputServices is ignored if nestjs=true

lautarodragan:


stephenh:
Thanks @lautarodragan !

PR #785, opened 2023-02-24
fix: implement abort grpc web

hersentino:
https://github.com/stephenh/ts-proto/pull/777 introduce abort for grpc web but I forgot to change argument order for one fonction, here is the fix. I'm really sorry about that, everything tested locally and working this time.


hersentino:
hey @stephenh could you review review this pr, or could I do something to get it reviewed ?

stephenh:
Hey @hersentino yeah, sorry about that! I typically only have time to triage ts-proto issues about every/other weekend or so. Thanks for the ping and the fix!

Issue #780, opened 2023-02-17
Why are outputted protos not classes


Ekrekr:
I'm trying out migrating to this from a different proto library because in general this looks much better! However we have hit a bit of a roadblock. We have variables that are of types of multiple kinds of protos. For example, we have a function that looks like:

```
function(a: proto1 | proto2 | proto3) {
  if (a instanceof proto1) {
    ...
  }
}
```

Because `ts-proto` generates an output of:

```
interface proto1 {...}
const proto1 = {...}
...
```

rather than classes, there doesn't seem to be a way of validating the proto type in this scenario - `instanceof` doesn't work with interfaces. Is there an expected way of accomplishing this?

See https://github.com/dataform-co/dataform/pull/1462/files#diff-8ac9e2a1b048a5e2289c9d7b9fcf1955150ce19191b86fc76c0e6cd7a643148a for full WIP.

stephenh:
Avoiding classes was just one of the original "idiomatic TS / just POJOs" design goals for ts-proto, and I think that has generally worked out + is what most users prefer / like about the library.

It is not `instanceof`, but there is an option to add a `$type` field to every message that would let you do similar checks but more like ADTs instead of classes:

```
--ts_proto_opt=outputTypeRegistry=true
```

I kinda thought we had an option to output `$type` but not literally the whole type registry infra...but maybe not.

stephenh:
Good! Fwiw I think adding a `outputTypeFlag` or something that did "just `$type`" from the type registry code would be a pretty simple PR; if you wanted to take a crack at it, that'd be great. Thanks!

PR #786, opened 2023-02-24
feat: Add type annotations flag

Ekrekr:
Fixes https://github.com/stephenh/ts-proto/issues/780


stephenh:
@Ekrekr this looks great! Looks like a build failure, hopefully minor but I haven't had a chance to take a look, can you see if it's an easy fix? Thanks!

Ekrekr (commented):
Have fixed this now - culprit was a `options.outputTypeRegistry || options.outputTypeRegistry` rather than `options.outputTypeAnnotations || options.outputTypeRegistry`.

Was having trouble with having to do `docker-compose` with no `--rm` vs `docker compose` previously so couldn't run tests locally, but got it working now.

Issue #773, opened 2023-02-07
Give the ability to export encode and decode methods in a separate way


Frabat:
Hi!
Currently I'm working on a project where we're using TS-proto to generate interfaces from protobuf files.
We are also using the encode methods generated by ts-proto to serialize binary messages.
Since we don't really need decode methods, and since those take up a lot of space, could be reasonable to insert a flag to generate only the encode (or decode) methods instead of both?

I've already did an edit for this, if this feature makes sense to you I could open a PR.

stephenh:
Hi @Frabat ; sure if you wanted to modify the existing `outputEncodeMethods` to support values like:

```
outputEncodeMethods = true | false | "encode-only" | "decode-only"
```

I think that'd be fine. Thanks!

Frabat:
Ok awesome! I'll finish and open a pull request

akornatskyy:
consider to look at this from the client vs server standpoint, e.g. in case of client I am interested in encode for request but decode for response... for the server it is the other way around.

olejorgenb:
This issue seems to have been solved by #787

PR #787, opened 2023-03-01
feat: output-encode-only methods

Frabat:
As requested by me in https://github.com/stephenh/ts-proto/issues/773, I've changed the outputEncodeMethods to accept also "encode-only" and "decode-only" string values.
Since my use case also didn't need the creation methods, I've added a third option "encode-no-creation" that doesn't export that. 
In theory, this could be included with the "encode-only" flag, since the creation method is used just by the encoder, but somebody might need it.


stephenh (approved):
@Frabat looks like the build failed with a prettier error, but otherwise looks good to me! Thanks!

Frabat:
Hi @stephenh can I solve the prettier error or is something CI related?

stephenh:
@Frabat if you run `yarn format` in the root directory, it should resolve the prettier errors, and then just push those changes up to the PR.

Frabat:
@stephenh done!

stephenh:
Erg, sorry for the flakey tests, I think we had several tests sharing the same port. I believe I've fixed it in:

https://github.com/stephenh/ts-proto/pull/790

Can you rebase on top of that / on top of main? Thank you!

stephenh:
Looks great! Thanks @Frabat !

PR #796, opened 2023-03-17
Fixing generator when both `forceLong=bigint` and `outputTypeRegistry=true` options are used

yflander:
Fixing generator when both `forceLong=bigint` and `outputTypeRegistry=true` options are used

Generated `BuiltIn` type was missing `bigint` type.


stephenh:
Looks great, thanks!

PR #797, opened 2023-03-17
Several feature additions and fixes

davidzeng0:
1. Adds group encoding and decoding support
2. Uses Reader.create for faster decoding
3. Generate _unknownFields for generic service options (extensions are planned)
4. useOptionals=all, initializeFieldsAsUndefined=false, initialize maps and repeated fields on use

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -20,7 +20,7 @@ export const SimpleMessage = {
   },
 
   decode(input: _m0.Reader | Uint8Array, length?: number): SimpleMessage {
-    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
+    const reader = input instanceof _m0.Reader ? input : _m0.Reader.create(input);
     let end = length === undefined ? reader.len : reader.pos + length;
     const message = createBaseSimpleMessage();
     while (reader.pos < end) {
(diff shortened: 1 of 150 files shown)


davidzeng0:
Since the rest of the repo is not updated for ts-proto 6.4.1, neither are the commits in this pull request

stephenh:
Hi @davidzeng0 ; thanks for the PR!

Do you mind breaking each of your changes out as a separate PR? Fwiw it looks like your commits are already very clean, so hopefully that would be too bad?

Of your changes:

1. Adds group encoding and decoding support
2. Uses Reader.create for faster decoding
3. Generate _unknownFields for generic service options (extensions are planned)
4. useOptionals=all, initializeFieldsAsUndefined=false, initialize maps and repeated fields on use

I think 2-4 all sound great and would be obvious PRs to land.

I'm a little hesitant about 1, b/c a) afaict groups are basically deprecated, and b) just scanning the PR I'm kinda surprised more work was not necessary to make them work; could you add an integration test that shows groups encoding/decoding working as expected?

I'm still on the fence about supporting groups, but if the code changes are minimal, and we have an integration test showing that it does actually work, then I'd be fine accepting it.

Thanks!

davidzeng0:
Sure thing

I added group support because I'm using some very old .proto files that uses groups and I've already thoroughly tested that they do in fact work (I basically copied the generating format from protoc's C++ code, then tested on --decode_raw)

davidzeng0:
Going to create separate pull requests

PR #799, opened 2023-03-17
feat: group encoding and decoding support

davidzeng0:
All tests pass


src/main.ts
```diff
@@ -957,6 +962,12 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
     const fieldName = maybeSnakeToCamel(field.name, options);
     chunks.push(code`case ${field.number}:`);
 
+    const tag = ((field.number << 3) | basicWireType(field.type)) >>> 0;
+    const tagCheck = code`
+      if(tag != ${tag})
```
stephenh, line 967: @davidzeng0 is this check necessary as part of the grouped message support? If so, can you give a brief overview/explanation? Thanks!

davidzeng0: Not necessarily, but because the same field number may be a string or a varint, it's useful to check that the tag matches before parsing to avoid errors. Google's C++ library does this aswell, and adds the result as an unknown field if the exact tag does not match.

stephenh: Cool, thanks for the explanation

stephenh (approved):

PR #800, opened 2023-03-17
perf: use Reader.create

davidzeng0:
Allows protobufjs/minimal to detect the config and optionally use BufferReader if supported.

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -20,7 +20,7 @@ export const SimpleMessage = {
   },
 
   decode(input: _m0.Reader | Uint8Array, length?: number): SimpleMessage {
-    const reader = input instanceof _m0.Reader ? input : new _m0.Reader(input);
+    const reader = input instanceof _m0.Reader ? input : _m0.Reader.create(input);
     let end = length === undefined ? reader.len : reader.pos + length;
     const message = createBaseSimpleMessage();
     while (reader.pos < end) {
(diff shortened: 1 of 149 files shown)


stephenh:
@davidzeng0 this looks great, but there is a failure in the build:

```
Error: illegal buffer
    at create_typed_array (/home/runner/work/ts-proto/ts-proto/node_modules/protobufjs/src/reader.js:47:15)
    at create_buffer (/home/runner/work/ts-proto/ts-proto/node_modules/protobufjs/src/reader.js:63:23)
    at Function.create_buffer_setup (/home/runner/work/ts-proto/ts-proto/node_modules/protobufjs/src/reader.js:64:15)
    at /home/runner/work/ts-proto/ts-proto/integration/simple/simple.ts:2072:78
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Disclaimer I've not looked in to it this at all, but can you tell if it's something wrong we're doing in the test, or maybe some wrinkle about `Reader.create` that we'll have to handle? Thanks!

davidzeng0:
@stephenh It seems like the test is returning the mock value `{}` and it failing  because an object isn't a buffer. I'm not entirely sure how tests before this commit passed, but I'll look into it

davidzeng0:
Yeah, so new Reader({}) succeeds because there are no type checkings, however Reader.create fails

image

Root cause on line 6
image

stephenh:
@davidzeng0 Ah good find! Do you mind pushing a fix up to this PR?

davidzeng0:
@stephenh Fixed

Also, any progress on the other 2 PRs?

Thanks.

Issue #788, opened 2023-03-02
error: repeated field is not iterable


akornatskyy:
`encode` fails if repeated field is `undefined`.

Expected behavior: wrap `for` loop iterating over repeated field with `undefined` check.

stephenh:
Fixed by https://github.com/stephenh/ts-proto/pull/802, thanks @davidzeng0 !

PR #802, opened 2023-03-17
fix: initialize undefined optional fields upon use

davidzeng0:


stephenh:
@davidzeng0 I've merged your other two PRs, so this one has a conflict now; could you rebase it? Thanks!

davidzeng0:
Rebased

stephenh:
@davidzeng0 Ah great, thanks for rebasing; can you try and find an existing `integration/*` test that matches the `parameters.txt` you're using and add a reproduction of what would previously blow up, but now does not?

Just to keep this behavior from regressing.

Also, do you think this PR will fix this issue:

https://github.com/stephenh/ts-proto/issues/788

My guess is yes?

Thanks!

davidzeng0:
This should fix https://github.com/stephenh/ts-proto/issues/788 and https://github.com/stephenh/ts-proto/issues/736

davidzeng0:
As expected, the test fails without https://github.com/stephenh/ts-proto/pull/802/commits/6f6f97d78d0322f9d43b8e56e4dcfe00af855ad3

stephenh:
Great integration test, thank you!

PR #806, opened 2023-03-24
feat: include _unknownFields as a field

davidzeng0:
Makes _unknownFields a part of messages for easier access
Field left as optional as to not break existing code
Also uses `for in` as it is marginally faster than `Object.keys`
Conformant to initializeFieldsAsUndefined:
- false -> _unknownFields always created
- true -> _unknownFields created on use

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -101,12 +101,8 @@ export function generateSchema(ctx: Context, fileDesc: FileDescriptorProto, sour
 
   let fileOptions: Code | undefined;
   if (fileDesc.options) {
-    fileOptions = encodedOptionsToOptions(
-      ctx,
-      ".google.protobuf.FileOptions",
-      (fileDesc.options as any)["_unknownFields"]
-    );
-    delete (fileDesc.options as any)["_unknownFields"];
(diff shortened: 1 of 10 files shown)


src/schema.ts
```diff
@@ -298,12 +279,8 @@ function resolveMessageOptions(ctx: Context, message: DescriptorProto): Code | u
...
   let messageOptions: Code | undefined;
   if (message.options) {
-    messageOptions = encodedOptionsToOptions(
-      ctx,
-      ".google.protobuf.MessageOptions",
-      (message.options as any)["_unknownFields"]
-    );
-    delete (message.options as any)["_unknownFields"];
+    messageOptions = encodedOptionsToOptions(ctx, ".google.protobuf.MessageOptions", message.options._unknownFields);
+    delete message.options._unknownFields;
```
stephenh, line 283: It took me awhile to figure out why this wasn't failing CI....

It looks like our CI process right now doesn't actually type-check until the release step, which is unintentional.

That said, I like this change, so :shrug: I'm going to try and push it through and then follow up with a ts-proto-descriptors bump, as you mentioned in the PR description.

stephenh (approved):
Nice! This is a good idea. Thanks!

Issue #682, opened 2022-10-10
Optionally omit defaults in JSON


alexhayes:
I'd really love a way of calling `toJSON` that omits the defaults from the output. Ideally this would be an argument supplied to the `toJSON` call because I have a use case for both.

1. **Use Case 1: Snapshots**
    In this use case I'd like to omit defaults to remove noise from Jest snapshots.
    
2. **Use Case 2: Browser/Workshop Output**
    In this use case I'd like to include defaults as it's easier when running a workshop if all defaults are included - rather than having to describe to attendees "oh, your attribute is not being set because it's the default"

Thus, my proposal would be that the signature for `toJSON` would take an optional second parameter `omitDefaults` which defaults to false (to maintain current behaviour).

This relates to [this section in the docs](https://github.com/stephenh/ts-proto/blame/main/README.markdown#L663-L681);

> When writing JSON, `ts-proto` currently does **not** normalize message when converting to JSON, other than omitting unset fields, but it may do so in the future.
> 
> ```typescript
> // Current ts-proto behavior
> Foo.toJSON({}); // => { }
> Foo.toJSON({ bar: undefined }); // => { }
> Foo.toJSON({ bar: '' }); // => { bar: '' } - note: this is the default value, but it's not omitted
> Foo.toJSON({ bar: 'baz' }); // => { bar: 'baz' }
> ```
> 
> ```typescript
> // Possible future behavior, where ts-proto would normalize message
> Foo.toJSON({}); // => { }
> Foo.toJSON({ bar: undefined }); // => { }
> Foo.toJSON({ bar: '' }); // => { } - note: omitting the default value, as expected
> Foo.toJSON({ bar: 'baz' }); // => { bar: 'baz' }
> ```
> 
> - Please open an issue if you need this behavior.

Given it appears some thought has already gone into this, would this use case be supported or is the preference to use a `--ts_proto_opt=` value?

stephenh:
Hey @alexhayes ; I don't have strong opinions about whether to just always add the `omitDefaults` flag, or to only output it behind a flag.

It seems fine to me to just always output it, with the default behavior as-is, and see if anyone complains.

If you'd like to submit a PR that'd be great!

migmartri:
This feature would be very helpful for me too! 

[...]

PR #807, opened 2023-03-24
feat: option fromPartialDefaultInitialize=false

davidzeng0:
Should be merged after https://github.com/stephenh/ts-proto/pull/806


stephenh:
@davidzeng0 wdyt about instead of adding a new option, we just use `initializeFieldsAsUndefined` and assume that if the user has set `initializeFieldsAsUndefined=false` then they also want `fromPartial` to leave fields uninitialized? 

Would that work, or am I missing something?

stephenh:
Also, just in general thanks for all the PRs! I don't always have time to respond to them during the week, but will try to catch up with you on the weekends. :-) 

(...also do you mind rebasing your PRs as each prior one is merged? For some reason GitHub doesn't update the diffs/list of commits to exclude your prior work; would be great if it just did that automatically imo.)

Thanks!

davidzeng0:
Yes, I don't mind rebasing. I will be PRing as long as I come across missing features that I need to use :).

I only added a new option because I thought that "initializeFieldsAsUndefined" is a decoding option, and it might make sense for an encoding option, but they can be the same option. Something else I can also fix is https://github.com/stephenh/ts-proto/issues/682

davidzeng0:
https://github.com/davidzeng0/ts-proto/commit/fa2d5df25c37d2acb5370c0f2b3b06bfd2a5719f uses the existing option
https://github.com/davidzeng0/ts-proto/commit/b5481ceb77f527e2d55725e4a685000b29c238c5 removes default initialization fir json

PR #808, opened 2023-03-26
feat: extensions

davidzeng0:
Can be automatically merged


davidzeng0:
I hope you like the way I implemented it, if you want any changes let me know

davidzeng0:
Any thoughts on this?

stephenh:
Hey @davidzeng0 ; from my initial scan of the tests, it looks pretty great, very standard approach to protobuf extensions AFAIU (I've not actually used them personally before). It'll probably be the weekend before I dive in a bit more, but its great to have this feature. Thank you!

davidzeng0:
Rebased and fixed tests

stephenh (approved):
This looks great!

Awhile ago I'd thought about maybe trying to figure out an alternative API to the stock `Extension.set(message, someConst, value)`, like something like:

```
const message = ...
const extended = Extension.wrap(message)
extended.someField = 1
extended.otherField = 2
```

But I don't think that's even possible b/c the proto file doesn't even have a name for what the wrapped type would be, i.e. something like:

```
extension MyNiceName extends Extendable {
  string someField = 1 
}
```

Would let us know to use `MyNiceName` as the type name for all of the `someField` / etc. fields ... but that is just not there/something they thought to include in the proto file. :thinking: So :shrug: I think that puts a damper on any cute/alternative APIs.

davidzeng0:
Thanks for being patient with me :)

davidzeng0:
A ts-proto-descriptors bump with outputExtensions=true would be great

davidzeng0:
ts-proto-descriptors bump: https://github.com/stephenh/ts-proto/pull/814

PR #811, opened 2023-03-27
feat: Update fromPartial and fromJson to respect initializeFieldsAsUndefined

stephenh:


davidzeng0:
Seems like there's an issue with ts-poet where it generates ?? instead of a ? b : c
https://github.com/davidzeng0/ts-proto/blob/b5481ceb77f527e2d55725e4a685000b29c238c5/src/main.ts#L1895

davidzeng0:
Reproduced

stephenh:
Thanks for fixing the build @davidzeng0 ! 

I'm thinking in retrospect, we're not _exactly_ following the intent of `initializeFieldsAsUndefined=false`, which would be for those keys to not even be set in the first place; so your suggested `fromPartialDefaultInitialize` is technically more correct.

That said, the option proliferation in ts-proto is just kind of excessive :-D so at this point I prefer _slightly_ less accurate semantics to another flag, and then if users end up complaining a lot about this in-exactness, we can always change it later.

Thank you!

davidzeng0:
No problem

Issue #212, opened 2021-02-11
Feature request: autogenerate index files


willclarktech:
With `protobuf.js` we used to generate all our type definitions in one big file and then we could easily re-export everything from a single index file. With ts-proto we now have nested files reflecting the structure very nicely, but manually creating index files that reflect this structure is going to be a pain. It would be nice to have an option for ts-proto to automatically generate index files that make each subdirectory a module that just re-exports everything from the files/subdirectories it contains.

stephenh:
I worry about the inevitable namespace collisions from messages with the same name being in different proto files, but otherwise yeah this seems like a good idea.

stephenh:
@alexisvisco can you clarify that with a concrete example?

kallangerard:
@alexisvisco please don't use Typescript namespaces. They're a completely unnecessarily relic since ES6 Modules, and support for them is very patchy.

stephenh:
An `outputIndex` option was added in #821

PR #821, opened 2023-05-03
feat: Generate Index Files

blorgon1:
[#212] These changes add an `outputIndex` option for generating index files, based on the proto package namespaces.

An index file is generated for each sub-namespace of a package, so that they can be imported using the same path as the proto package, without relying on TS namespaces.

Example - Given the following proto files:
```
/// a.proto
syntax = "proto3";
package com.test;

message A {
    string a = 1;
}

/// b.proto
syntax = "proto3";
package com.test;

message B {
    string b = 1;
}
```
Three index files will be generated:
```
/// index.ts
export * as com from "./index.com";

/// index.com.ts
export * as test from "./index.com.test";

/// index.com.test.ts
export * from "./a";
export * from "./b";
```

The index can then be used as:
```
import protos from './protos'; // the pb output directory
type A = protos.com.test.A;
```


stephenh:
Hi @blorgon1 , thanks for the PR!

Can you pick either an existing `integration/` test, or create a new one, that uses the new flag, and check in the generated output as part of this PR? This will basically act as a regression test to make sure we don't break the option in the future.

stephenh:
Hi @blorgon1 ! I had a chance to review the PR. The nested exports initially threw me off, but reading your PR description, I get your goal, i.e. the `package com.test.domain` w/`AuthorMessage` in it turns into:

```
const AuthorMessage = protos.com.test.domain.AuthorMessage;
```

It is nice that this approach avoids conflicts, b/c each package gets its own output.

Fwiw I have two musings:

1) I can anticipate some users wanting a "flat" / package-less export, i.e. even at the risk of conflicts, if they _don't_ have conflicts, wanting to do something like:

```
imports { AuthorMessage } from "src/ts-proto-output";
```

No need to build that in your PR, just thinking that maybe someday we'd have an `outputIndex=flat` or `outputIndex=packageless` option someday.

2) I'm wondering how strongly you feel about the dotted package syntax, b.c. if we snake or camel cased the proto package name, we could probably have just 1 index file, i.e.:

* Given `a.proto` w/`package com`
* And `b.proto` w/`package com.test`

We could generate a single index.ts that looked like:

```
export * as com from "./a";
export * as comTest from "./b";
```

Then use it as:

```
import { comTest } from "./src/ts-proto-outout";
const a = comTest.Author...
```

Granted, you've already got the export-N-levels of working, but, dunno, I think the camel-case-the-name seems simpler to implement and possibly consume, just b/c there are less "shell packages" that exist solely to mimic namespaces.

I'm not completely against the current approach, but do have a general preference for using something like camelized package names to avoid the N-level exports; what do you think?

stephenh:
Hey @blorgon1 , thanks for the pings! Yep, the tests pass, and I'll go ahead and merge.

Just fwiw I am still a little hesitant about the "imports to mimic namespaces" approach, so on the very low chance that I end up having to debug/support this in more esoteric cases, I can't promise to not fallback to a less-Russian-doll-of-imports approach, but hopefully we won't need to do that.

Thanks for the PR! This has been a popular ask for a long time!

PR #823, opened 2023-05-04
Only return Observavle types for NestJS if the option is switched on

alexkb:
We're using `--ts_proto_opt=nestjs=true` and wanted to remove `Observable`'s being returned, so we tried `--ts_proto_opt=returnObservable=false` but it had no effect.

This PR fixes that issue so it does behave as expected. If it is incorrect, then I suggest you update the documentation to make it clear that `returnObservable` has no effect when `nestjs=true`. Thanks.

### DIFF
--- a/src/generate-nestjs.ts
+++ b/src/generate-nestjs.ts
@@ -55,13 +55,13 @@ export function generateNestjsServiceController(
     let returns: Code;
     if (isEmptyType(methodDesc.outputType)) {
       returns = code`void`;
-    } else if (options.returnObservable || methodDesc.serverStreaming) {
+    } else if (methodDesc.serverStreaming) {
       returns = code`${responseObservable(ctx, methodDesc)}`;
     } else {
       // generate nestjs union type
       returns = code`
         ${responsePromise(ctx, methodDesc)}
-        | ${responseObservable(ctx, methodDesc)}
+        ${options.returnObservable ? "\n| " + responseObservable(ctx, methodDesc) : ''}
         | ${responseType(ctx, methodDesc)}
       `;
     }


tmthecoder:
Bump on this, using nestJS currently doesn't allow me to generate methods returning a promise

alexkb:
@tmthecoder if you've got `--ts_proto_opt=nestJs=true` set then you should get Promise's along with Observable's, which you don't have to use. You won't get them though, if you've also set `ts_proto_opt=returnObservable=true` OR if your methods involve streams. Whatever the case, I've just re-synced this branch with `main` but I'm not sure what the maintainer wants to do here sorry.

tmthecoder:
@alexkb I see the types be promise or observable for the controller implementations but the clients only have observable in the definitions, not sure why that is.

integration/nestjs-simple/hero.ts
```diff
@@ -49,13 +49,13 @@ export interface HeroServiceClient {
...
-  findOneHero(request: HeroById): Promise<Hero> | Observable<Hero> | Hero;
+  findOneHero(request: HeroById): Promise<Hero> | Hero;
 
-  findOneVillain(request: VillainById): Promise<Villain> | Observable<Villain> | Villain;
+  findOneVillain(request: VillainById): Promise<Villain> | Villain;
 
   findManyVillain(request: Observable<VillainById>): Observable<Villain>;
 
-  findManyVillainStreamIn(request: Observable<VillainById>): Promise<Villain> | Observable<Villain> | Villain;
+  findManyVillainStreamIn(request: Observable<VillainById>): Promise<Villain> | Villain;
```
stephenh, line 58: @alexkb Disclaimer that I don't use NestJS, but wouldn't just `Promise<Villain>` make more sense here as a return type?

I agree that your PR is already better than the previous `Observable<T> | Promise<T> | T`, which imo seems just kind of maddening to use as a caller, b/c you'd have to constantly be checking what the return value was.

rathod-sahaab: @stephenh NestJS can also have non async function for controllers and services hence returning T makes sense
this will valid implementation of function within nest framework
```ts
   createPaymentLink(
      request: CreatePaymentLinkRequest,
   ): CreatePaymentLinkResponse {
      //...

      return {
         url: 'Payment Juspay: https://lesgo.tealfeed.com',
      }
   }

```

This wont be allowed if we go with `Promise<T>`

rathod-sahaab:
@alexkb shouldn't this also effect the client part which only returns the observable

Issue #825, opened 2023-05-06
Import Observable as type


spheenik:
Hi,

first of all, great project, using it with a lot of joy.

I have a small feature request. I am using the generated code in a svelte project, where the `tsconfig.js` does have

`"importsNotUsedAsValues": "error"`

This gives me a tslint error in the generated api.ts, because it imports `Observable` as

`import { Observable } from "rxjs";`

instead of

`import type { Observable } from "rxjs";`

I was trying to come up with a pull request, but I can't seem to find where it happens. I saw in `tspoet` that it supports importing as type, so maybe it'll take you only a minute. Not really a big deal though, I can simply add it everytime after compiling the API.

stephenh:
Fixed by #826 , thanks @spheenik !

spheenik:
My pleasure!

PR #826, opened 2023-05-06
import Observable as a type

spheenik:
I think I found it.
Possibly fixes issue #825

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -694,7 +694,7 @@ export function observableType(ctx: Context): Code {
   if (ctx.options.useAsyncIterable) {
     return code`AsyncIterable`;
   }
-  return code`${imp("Observable@rxjs")}`;
+  return code`${imp("t:Observable@rxjs")}`;
 }
 
 export function requestType(ctx: Context, methodDesc: MethodDescriptorProto, partial: boolean = false): Code {


stephenh:
Ah yep, this seems reasonable; thanks for the PR! Let's see what the test suite says.

Fwiw to get the PR green, you'll need to update the `integration/` test output, and push it up as part of this PR, since that serves as our integration tests; there should be instructions in the readme, but let me know if you have questions. Thanks!

spheenik:
Hi Stephen,
did as described in the readme.

It turned out that you use `observableType()` in `types.ts` both for creating the "generated api" as well as your test cases.
So I opted to introduce a new parameter `asType` to it, which is `false` by default, and only set it to `true` when the api generation happens.

Hope I did everything right.

stephenh:
@spheenik fwiw I'm not 100% following; it's been awhile since I've looked at the `Observable` test suites, but can you link to where adding the `t:` broke a test? 

Also fwiw it looks like, at least in the current PR, the `asType` is never passed `true`, so I don't believe there is any behavior change at the moment.

spheenik:
Lol, I set it to false... I just corrected it.

spheenik:
It breaks a single test when it is imported by type:

```
Summary of all failing tests
 FAIL  integration/grpc-web/example-test.ts
  ● Test suite failed to run

    integration/grpc-web/example.ts:983:16 - error TS1361: 'Observable' cannot be used as a value because it was imported using 'import type'.

    983     return new Observable((observer) => {
                       ~~~~~~~~~~

      integration/grpc-web/example.ts:5:15
        5 import type { Observable } from "rxjs";
                        ~~~~~~~~~~
        'Observable' was imported here.
```

stephenh:
Ah great! Thanks for including the snippet that failed; that makes sense, and verifies we've got existing regression coverage for this change, so don't need any new tests for this PR. Thanks for the fix!

PR #828, opened 2023-05-10
refactor: use the AbortSignal reason for rejection

bouk:
The AbortSignal [has a 'reason' property](http://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) that can be used to pass through an exception. We should pass that on


stephenh:
Looks great, thanks @bouk !

PR #832, opened 2023-05-11
fix: ensure generated fromTimestamp works when useOptionals=all

miselin:
The generated `fromTimestamp` raises a Typescript compilation error `'t.seconds' is possibly 'undefined'` when `useOptionals=all` as all the fields in the generated timestamp message are optional.

This PR changes the generated function to handle an undefined value by defaulting to zero.

(drafting while I get the test files updated)

### DIFF
--- a/integration/file-suffix/parent.pb.ts
+++ b/integration/file-suffix/parent.pb.ts
@@ -113,8 +113,8 @@ function toTimestamp(date: Date): Timestamp {
 }
 
 function fromTimestamp(t: Timestamp): Date {
-  let millis = t.seconds * 1_000;
-  millis += t.nanos / 1_000_000;
+  let millis = (t.seconds || 0) * 1_000;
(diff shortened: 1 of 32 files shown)


stephenh:
Hi @miselin ; ah yeah, this is a good fix! Updating the test files should work with `yarn bin2ts`, but lmk if that's not working. Thanks!

miselin:
> Hi @miselin ; ah yeah, this is a good fix! Updating the test files should work with `yarn bin2ts`, but lmk if that's not working. Thanks!

I think I'm fighting against something with Docker and mount permissions somewhere:
```
$ yarn build:test
./map-long-optional/test.proto
protoc-gen-dump: line 18: file.bin: Permission denied
mv: overwrite './map-long-optional/test.bin'? ^C
```

The scripts/tests don't have write access to the directory but it's a rw mount... very strange 🤔

Issue #836, opened 2023-05-19
It maybe !observer.closed => observer.closed when set returnObservable=true


KiyomiHan:
https://github.com/stephenh/ts-proto/blob/75f9f4a674a6cb402e6e2070da4174416069007e/src/generate-grpc-web.ts#L386

it seems like

``` javascirpt
if (observer.closed)  return client.close();
```

because when call the `unsubscribe()`  the `closed` set be true.

current code, make me can not close the grpc streamimg response

stephenh:
Hi @KiyomiHan , I'm not sure and honestly am not really following; @hersentino I think wrote the code you linked to; Francois lmk if you have any thoughts on ^. Thanks!

hersentino:
Hi,

I'm not sure if this condition is the real issue here. I've tried removing it locally and used the `unsubscribe` function, but it seems like the `add` callback is never triggered. @KiyomiHan Can you test this on your end to confirm whether it works without the condition or not? You can use [yalc](https://www.npmjs.com/package/yalc) to easily link your local ts-proto project to your project.

I initially added this condition to prevent the close function from being called again if it has already been closed by the abort signal system. So, the condition can safely be removed, but as I mentioned earlier, I don't think this will resolve your problem.

I'm not an expert with observables, but I believe the issue here is that the `add` callback is not being called even when the subscription is unsubscribed. I'm trying to figure out why. The `add` callback was already implemented when I introduced the abort signal system. I'm not sure if it was working or not at that time 

Why don't you use the abort signal system instead of the unsubscribe method? The abort signal system is functioning properly

hersentino:
The added condition to the pr is required, If I remove the condition, an error is triggered when the stream is aborted. This is why I initially wrote the faulty condition

<img width="653" alt="Screenshot 2023-05-22 at 14 32 43" src="https://github.com/stephenh/ts-proto/assets/26276665/fc4a36e8-fc50-4878-a739-929cf01cd3a9">

This error occurs because the close event is triggered twice: once by the abort event and once by the unsubscribe event.

With this pull request, everything works fine, both with and without the useAbortSignal option. I have tested both cases locally.

PR #837, opened 2023-05-22
Fix: add callback close condition

hersentino:
In order to solve https://github.com/stephenh/ts-proto/issues/836


src/generate-grpc-web.ts
```diff
@@ -383,7 +383,12 @@ function createInvokeMethod(ctx: Context) {
             },
           });
           observer.add(() => {
-            if (!observer.closed) return client.close()
+           ${
+             !useAbortSignal
+               ? `return client.close();`
+               : `if (!abortSignal || !abortSignal.aborted)
```
stephenh, line 389: Ah gotcha; @hersentino it looks like we don't have an `integration/` test that generates the `!abortSignal || !abortSignal.aborted` case? If not, do you mind adding one? Just like copy/paste the `grpc-web` directory to like `grpc-web-abort-signal` and add the abort signal flag to `parameters.txt` I think? 

No need to actually have a runtime test suite that covers this behavior, b/c I assume that is hard to test (unless it's easy to test, then feel free! :-), but at least having some generated code checked in that shows this code being generated will help us keep the behavior from regressing.

Thanks!

hersentino: >  it looks like we don't have an integration/ test that generates the !abortSignal || !abortSignal.aborted case?

We have one, it is located in `integration/grpc-web-no-streaming-observable`, but since it does not contain streaming-observable and my fix is specifically related to streaming, there was no difference when I ran `bin2ts`.

> do you mind adding one?

@stephenh Tell me what you prefer, keeping it in `integration/grpc-web-no-streaming-observable` or creating a new one?

hersentino: It was quick, so I added it

stephenh: That's great, thanks @hersentino !

PR #848, opened 2023-06-12
docs: update README with protobufPackage note

haakonfp:
I suggest adding the note that `protobufPackage` is also skipped in the export when using the `exportCommonSymbols=false`  flag. I spent a bit too long looking for this information, and found it in some old PR or issue discussion.

This should make it easier for users to find this information, as they can just search the page for `protobufPackage`.


stephenh:
Ah shoot, good suggestion; thank you!

PR #850, opened 2023-06-13
feat: support "lib: es6"

f0rmat1k:
es2017 doesn't require array length argument. But 2015 and below needs it https://github.com/microsoft/TypeScript/blob/2a37eb2e471e01e510849f0dc108b3699ab869db/src/lib/es2015.iterable.d.ts#L268

any way according documentation zero – is default initializer https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array

So if you have `lib` param in tsconfig below `es2017`, you've got a problem with typechecking.
I just replaced every empty call to `(0)`. It must work on any es lib configuration.

### DIFF
--- a/integration/bytes-as-base64/message.ts
+++ b/integration/bytes-as-base64/message.ts
@@ -7,18 +7,18 @@ export interface Message {
 }
 
 function createBaseMessage(): Message {
-  return { data: new Uint8Array() };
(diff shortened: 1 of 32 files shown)


stephenh:
Cool, thanks @f0rmat1k !

PR #851, opened 2023-06-20
feat: expose service name as separate exported constant

paralin:
The service ID (for example, "simple.Echoer") is currently hidden within the constructor of the service implementation.

While it's possible to override the service ID with a constructor parameter, we should also export this ServiceID constant so that it can be used elsewhere.

### DIFF
--- a/integration/async-iterable-services-abort-signal/simple.ts
+++ b/integration/async-iterable-services-abort-signal/simple.ts
@@ -108,11 +108,12 @@ export interface Echoer {
   EchoBidiStream(request: AsyncIterable<EchoMsg>, abortSignal?: AbortSignal): AsyncIterable<EchoMsg>;
 }
 
+export const EchoerServiceName = "simple.Echoer";
 export class EchoerClientImpl implements Echoer {
   private readonly rpc: Rpc;
   private readonly service: string;
   constructor(rpc: Rpc, opts?: { service?: string }) {
-    this.service = opts?.service || "simple.Echoer";
+    this.service = opts?.service || EchoerServiceName;
     this.rpc = rpc;
     this.Echo = this.Echo.bind(this);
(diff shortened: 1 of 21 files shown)


integration/async-iterable-services-abort-signal/simple.ts
```diff
@@ -108,11 +108,12 @@ export interface Echoer {
   EchoBidiStream(request: AsyncIterable<EchoMsg>, abortSignal?: AbortSignal): AsyncIterable<EchoMsg>;
 }
 
+export const EchoerServiceID = "simple.Echoer";
```
stephenh, line 111: Kind of picky, but can we call these `EchoerServiceName` instead of `...ID`?

I'd remembered that services have names, i.e. from the `serviceDesc.name` field, and so initially thought the ID might be something else/different.

(At least I assume "name" is the most canonical term to use here--lmk if you think not.)

paralin: Sure, that's fine with me. But I named it service id to be consistent with other places I'm using the value in starpc (RPC) which is mirroring drpc which I believe also uses the term service ID.

Either way is fine with me, I'll rename it if you don't comment otherwise

stephenh: Nice, I hadn't seen drpc before; I really like/agree with their rationale of "the out-of-the-box grpc runtime seems huge/overly complex".

I tried to do some more googling across the grpc docs/etc of "service id" vs. "service name", and am not finding an obvious winner, so yeah would appreciate going with name.

Thanks!

stephenh (approved):

paralin:
Refactored -> ...ServiceName

stephenh:
Looks great, thanks @paralin !

Issue #853, opened 2023-06-22
Nested importMapping files are generated (but unused)


marcushultman:
Example:
```
// a.proto
import "b.proto"
```
```
// b.proto
import "c.proto"
```
When compiling `a.proto` with options:
`-I/src/b -I/src/c --ts_proto_opt=Mb.proto=foo-b`

Expectation:
Only `a.ts` is generated, with `import ... from 'foo-b';`

Reality:
Both `a.ts` **and** `c.ts` are generated - but `c.ts` is not used (transitively).

Let me know if this expectation is correct, or is there is a way block transitive dependencies from being generated.

marcushultman:
@stephenh What do you think? :)

marcushultman:
I had a look at the source, and it seems like the file resolution doesn't catch this case:
https://github.com/stephenh/ts-proto/blob/07830bd9019ce4b119d46f194a4e52b086c8dea3/src/plugin.ts#L22-L31

PR #854, opened 2023-06-23
fix: don't generate transitively imported files for mapped imports

marcushultman:
Closes https://github.com/stephenh/ts-proto/issues/853.

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -18,16 +18,33 @@ async function main() {
   const typeMap = createTypeMap(request, options);
   const utils = makeUtils(options);
   const ctx: Context = { typeMap, options, utils };
+  let filesToGenerate;
 
-  const filesToGenerate = options.emitImportedFiles ? request.protoFile : protoFilesToGenerate(request);
+  if (options.emitImportedFiles) {
+    const fileSet = new Set();
+    const addFilesUnlessAliased = (filenames: string[]) => {
+      filenames
+        .filter((name) => !options.M[name])
+        .forEach((name) => {
+          fileSet.add(name);
+          const file = request.protoFile.find((file) => file.name === name);
+          if (file && file.dependency.length > 0) {
+            addFilesUnlessAliased(file.dependency);
+          }
+        });
+    };
+    addFilesUnlessAliased(request.fileToGenerate);
+    filesToGenerate = request.protoFile.filter((file) => fileSet.has(file.name));
+  } else {
+    filesToGenerate = protoFilesToGenerate(request).filter((file) => !options.M[file.name]);
+  }
+
   const files = await Promise.all(
-    filesToGenerate
-      .filter((file) => !options.M[file.name])
-      .map(async (file) => {
-        const [path, code] = generateFile(ctx, file);
-        const content = code.toString({ ...getTsPoetOpts(options), path });
-        return { name: path, content };
-      })
+    filesToGenerate.map(async (file) => {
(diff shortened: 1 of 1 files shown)


marcushultman:
@stephenh I took a stab at fixing https://github.com/stephenh/ts-proto/issues/853, if the expectation in that issues is indeed correct.

src/plugin.ts
```diff
@@ -18,16 +18,32 @@ async function main() {
...
+    addFilesDeep(request.fileToGenerate);
+    filesToGenerate = request.protoFile.filter((file) => fileSet.has(file.name));
+  } else {
+    filesToGenerate = protoFilesToGenerate(request);
+  }
 
-  const filesToGenerate = options.emitImportedFiles ? request.protoFile : protoFilesToGenerate(request);
   const files = await Promise.all(
-    filesToGenerate
-      .filter((file) => !options.M[file.name])
```
stephenh, line 25: Won't removing this line mean that with `emitImportedFiles=false`, we are not filtering the `protoFilesToGenerate` against `options.M` anymore?

Or are you thinking that is necessary, b/c `protoFilesToGenerate` already has only the files directly from the CLI, so should be okay w/o this filter?

src/plugin.ts
```diff
@@ -18,16 +18,32 @@ async function main() {
   const typeMap = createTypeMap(request, options);
   const utils = makeUtils(options);
   const ctx: Context = { typeMap, options, utils };
+  let filesToGenerate;
+
+  if (options.emitImportedFiles) {
+    const fileSet = new Set();
+    const addFilesDeep = (filenames: string[]) => {
+      filenames = filenames.filter((name) => !options.M[name]);
```
stephenh, line 26: Just as a style thing, can you avoid the reassignment? Like do:

```
filenames.filter(...).forEach(name => ...)
```

src/plugin.ts
```diff
@@ -18,16 +18,32 @@ async function main() {
   const typeMap = createTypeMap(request, options);
   const utils = makeUtils(options);
   const ctx: Context = { typeMap, options, utils };
+  let filesToGenerate;
+
+  if (options.emitImportedFiles) {
+    const fileSet = new Set();
+    const addFilesDeep = (filenames: string[]) => {
```
stephenh, line 25: Wdyt of renaming this to `addFilesUnlessAliased`? Just to highlight that we're filtering those out.

stephenh (commented):
Hi @marcushultman , thanks for the PR! A few small stylistic asks + one question, but otherwise this lgtm!

src/plugin.ts
marcushultman: To me, it makes sense that it would be an error to map a file that is explicitly being passed as input to `protoc`. Emitting this error could be a separate PR. So it was intentional not to add filtering in the `emitImportedFiles=false` case - maybe I should add it anyway, to keep previous odd behavior?

marcushultman: I added it back. Thanks!

marcushultman:
@stephenh Addressed your comments. I added back the alias filtering to the `emitImportedFiles=false` case - but I would probably consider this a bug (it doesn't make sense to import map a file explicitly passed to `protoc` IMO).

stephenh:
> it doesn't make sense to import map a file explicitly passed to protoc IMO

Yeah, that's true...that said doesn't hurt to keep backwards compatibility I guess. Thanks for the change!

PR #858, opened 2023-06-30
Add const-only variant to to outputTypeAnnotations option

bouk:
Hi, I want to write a generic function like such:

```
function Encode<D>(desc: {$type: string, encode(value: D): Writer}, value: D)
```

So I need to have type annotations emitted but also still have the interfaces generated without the `$type` field. In this PR I'm adding a const-only variant to `outputTypeAnnotations` that does this.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -602,7 +602,10 @@ function makeDeepPartial(options: Options, longs: ReturnType<typeof makeLongUtil
   );
 
   // Based on https://github.com/sindresorhus/type-fest/pull/259
-  const maybeExcludeType = options.outputTypeAnnotations || options.outputTypeRegistry ? `| '$type'` : "";
+  const maybeExcludeType =
+    (options.outputTypeAnnotations || options.outputTypeRegistry) && options.outputTypeAnnotations !== "const-only"
+      ? `| '$type'`
+      : "";
   const Exact = conditionalOutput(
     "Exact",
     code`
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -616,7 +619,9 @@ function makeDeepPartial(options: Options, longs: ReturnType<typeof makeLongUtil
 
   // Based on the type from ts-essentials
   const keys =
-    options.outputTypeAnnotations || options.outputTypeRegistry ? code`Exclude<keyof T, '$type'>` : code`keyof T`;
+    (options.outputTypeAnnotations || options.outputTypeRegistry) && options.outputTypeAnnotations !== "const-only"
```
stephenh, line 622: Hm, this logic is doing in enough places, can you create like a `addTypeToMessages(options)` helper function?

stephenh (commented):
Hi @bouk ! This sounds like a good idea...

Two asks:

1) do you mind renaming this to `outputTypeAnnotations=staticOnly`? The "const-only" at first made me think of the TypeScript `const` feature, like the output would become readonly or `as const` or what not--but makes senses, you just want only the "static" `$type` field, and not one added to the actual message interface.

2) Can you create a new `integration/*` directory with a `parameters.txt` that uses the `staticOnly` param, just to get some regression coverage of this feature? 

Thanks!

bouk:
I've renamed it to `static-only` (to match other options), added a helper and added an integration test

stephenh:
Cool, thanks @bouk ! Looks like there is a build failure; could you :eyes: , but otherwise is great, and I'll merge once that's passing.

stephenh (approved):

bouk:
@stephenh forgot to add a parameter file!

PR #863, opened 2023-07-05
fix: generate different MessageType when using static-only

### LINKED ISSUE #861: Combining outputTypeAnnotations=static-only with outputTypeRegistry=true doesn't type check
Specifying these options:

```
--ts_proto_opt=outputTypeAnnotations=static-only \
--ts_proto_opt=outputTypeRegistry=true \
```

Doesn't type check because of this line:

https://github.com/stephenh/ts-proto/blob/14014207ff9159bb53b239a3ed60597929b17ec4/src/generate-type-registry.ts#L28

That expects the Message type to have a `$type` field, which it doesn't if `outputTypeAnnotations=static-only`. So it should handle that case by being slightly less type safe.

bouk:
Fixes #861


stephenh (approved):
Great, thanks!

Issue #859, opened 2023-07-01
Options for remove toJSON


f0rmat1k:
There is no options, to remove toJSON from output. Only both with `--ts_proto_opt=outputJsonMethods=false`. We are using tool on the client and we need only fromJSON. It would be nice to have option about removing toJSON. I can try to send PR, if you don't mind

stephenh:
Hi @f0rmat1k , sure if you wanted to add support for `outputJsonMethods=fromJSON`, so it becomes like `outputJsonMethods=true|false|fromJSON|toJSON"` that sgtm. Thanks!

Frabat:
Hi! I didn't read this issue but I've managed to do the same thing!
[I've opened a PR for this](https://github.com/stephenh/ts-proto/pull/865)

PR #865, opened 2023-07-05
feat: added encode-only mode for JSONs

Frabat:
Hi! There's an [issue](https://github.com/stephenh/ts-proto/issues/859). I've made a PR to fix this

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -7174,11 +7174,11 @@ __metadata:
   linkType: hard
 
 "semver@npm:2 || 3 || 4 || 5":
-  version: 5.7.1
-  resolution: "semver@npm:5.7.1"
+  version: 5.7.2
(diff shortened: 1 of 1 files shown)


stephenh (commented):
Hi @Frabat , thanks for the PR! It looks great, with a few minor asks; can you:

- Revert the change to `yarn.lock`, it seems it got downgraded to yarn v1
- Revert the newline change to `ts-proto.Dockerfile`
- Remove any `*.bin` files that aren't directly related to `*.proto` changes that you're making in the PR (...per this last point, somehow we got out-of-date `*.bin` files on main, which generally shouldn't happen but I guess it did--I've pushed updated versions of those, so if you rebase, they should get dropped from your PR)

Thanks!

Frabat:
@stephenh yep, I didn't notice those changes. I will do it today and update the PR

PR #868, opened 2023-07-07
feat: Ensure strict(er) TS compliance for the generated code

threema-lenny:
Changes:

- Add a new `tsc:check` script to ensure that there are no TS files with undetected errors.
- Split the TS configs in `integrations` so the compiled proto TS files can underly much stricter configuration requirements than the test files themselves.
- Add `DOM` to `compilerOptions.lib` for integration tests to have DOM's `AbortSignal` in scope.
- Apply the strictest TS preset there is to the compiled proto TS files (with a few exceptions that may be resolved later).
- Update the generated code to adhere to the stricter TS config.
- Modernise a few areas of the generated code.
- Add `tsc:check` to the CI script

---

There were a bunch of type errors in the integration tests that weren't caught by anything. I honestly don't know Jest enough to understand why that happens but at least the new `tsc:check` script catches them correctly.

I've applied the new `unwrap` utility in a couple of instances where I'm not precisely sure what the intended behaviour is, so this is a best effort guess. In any case, it doesn't make the experience any worse because those were uncaught error cases.

Resolving unused parameters `noUnusedParameters: true` is probably a Sisyphean task but it would be great to resolve the very few areas where there were unused locals `noUnusedLocals: true` in the future.

Note: Bumping Node types from 14 to 16 was kinda arbitrary. Node 16 is almost EOL, too.


threema-lenny:
I noticed that the CI still runs Node 14. Considering it's EOL and I've bumped the Node types, should it be removed from the test matrix?

stephenh:
@threema-lenny sure, I'm good dropping Node 14 support, thanks!

integration/extensions/test.ts
```diff
@@ -44,19 +44,19 @@ export function enumToJSON(object: Enum): string {
 }
 
 export interface Extendable {
-  field?: string;
-  _unknownFields?: { [key: number]: Uint8Array[] };
+  field?: string | undefined;
```
stephenh, line 47: @threema-lenny is this just a stylistic preference? I thought `field?: string` and `field?: string | undefined` were semantically equivalent.

integration/batching-with-context-esModuleInterop/batching.ts
```diff
@@ -770,3 +770,10 @@ function isObject(value: any): boolean {
 function isSet(value: any): boolean {
   return value !== null && value !== undefined;
 }
+
+function unwrap<T>(value: T | undefined | null): T {
```
stephenh, line 774: This makes sense, but we already use the term wrap/unwrap for dealing with wrapper types like `google.protobuf.StringValue`.

So can we use a name like `ensure`, `ensureSet`, `failIfUnset`, `check`...

Fwiw typically instead of a function, I'd write:

```
return ids.map(key => res.entites[key] ?? fail());
```

And define `fail` as:

```
export function fail(message?: string): never {
  throw new Error(message || "Failed");
}
```

Given that would most match what I'd write by hand, wdyt? Don't have a strong preference if you feel strongly about it being a function that takes a value.

integration/generic-metadata/some-file.ts
```diff
@@ -0,0 +1,3 @@
+export interface Foo {
+  foo: "bar";
```
stephenh, line 2: Is this an extra file that just snuck in?

stephenh (commented):
Hi @threema-lenny , this looks really great, thanks for catching these issues and the clean up! I'm still going through the PR and might have another question or two, hopefully should review the rest of it tomorrow. Thanks!

integration/tsconfig.json
```diff
@@ -1,12 +1,15 @@
...
-    "strict": true,
-    "outDir": "build",
-    "skipLibCheck": true,
-    "experimentalDecorators": true
+    "lib": ["es2018", "DOM"],
   },
-  "include": ["./"]
-}
+  "include": [
+    "**/*-test.ts"
```
stephenh, line 7: How about renaming this file to `tsconfig.tests.json` to highlight it's now only for the test files?

integration/extensions/test.ts
threema-lenny: Unfortunately they aren't because there's a semantic difference between _presence with `undefined`_ and _no presence_ of a property. [The TS config reference for `exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) enforces that semantic difference and also has a short example for it.

integration/generic-metadata/some-file.ts
threema-lenny: Yep. I added that file because `some-file` is directly referenced in the generated code but was not provided. This bug was picked up by the `tsc:check` script.

integration/extensions/test.ts
stephenh: Huh, I knew about that difference wrt runtime behavior (observing lack-of-presence vs. present-and-undefined), but didn't know about `exactOptionalPropertyTypes`.

I've also just used `strict: true` and assumed I was getting the most strict checking, so am surprised to learn about this `strictest` setting that turns on even a few more things--but it explains why I'd never been tripped up by `foo?: string` vs `foo?: string | undefined` before.

Sounds good then, thanks!

integration/tsconfig.json
stephenh: Ah gotcha. Yeah, truly restructuring the generated code from the tests does sound like a good idea, i.e. its what I do in real projects anyway, but would be a large effort for probably just a small convenience/wrinkle at the moment, so I'm good with the current setup you've got. Thanks!

.github/workflows/lint.yml
```diff
@@ -21,7 +21,16 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - name: install dependencies
-        run: yarn install || echo "ignore failure"
+        run: yarn install
+      - name: Prepare Integration
+        run: ./integration/pbjs.sh
+      # This will fail if any git-tracked file has changed
```
stephenh, line 27: It will fail? That would be great if so, but I don't think we already had that implemented, and I'm not seeing any new code here or in `yarn tsc:check` that would detect it?

Fwiw we do this for our internal project, and it looks like it's as simple as adding a `git status && git diff --exit-code`, but I hadn't gotten around to verifying + copy/pasting that over.

Do we need something like that `git diff --exit-code`, or am I missing how this comment is already implemented?

stephenh:
Looks great @threema-lenny ! Fwiw I changed the PR title over to "feat:" to trigger a release when this merges. Going to go ahead and merge this.

.github/workflows/lint.yml
stephenh: Ah I see! Yep, looks like the comment was in slightly the wrong place in `build.yml`; it should have been applied to the next step & not the tsc compile step.

I went ahead and moved the comment in `build.yml` to the more-correct spot + updated the `lint.yml` to do the same `git diff` check for good measure, and just pushed those up as a small chore commit.

Thanks!

Issue #834, opened 2023-05-15
Update to protobufjs 7


dbrgn:
protobufjs 7 has been out for almost a year now: https://github.com/protobufjs/protobuf.js/blob/master/CHANGELOG.md#700-2022-07-08 It contains quite a few bugfixes.

Are there any plans to upgrade to protobufjs 7.x? (I'm working in a project that uses both ts-proto and protobufjs, so this is currently an upgrade-blocker for me 🙂)

stephenh:
@dbrgn I'd love to see an upgrade to protobuf 7, but personally am more in "stewardship mode" for ts-proto that actively making changes; would be great if you wanted to take this on, and I'd be happy to help out on the PR.

Can you tell how big of an upgrade it will be? Scanning the release notes, it doesn't seem too bad?

vriesk:
Upping this.

There have been recent security issues with `protobufjs`, fixed in 7.2.4 release.

ping-localhost:
Please update to protobufjs 7, the old versions contain a Prototype Pollution vulnerability. 

See https://github.com/advisories/GHSA-h755-8qp9-cq85 for more information

dbrgn:
Sorry for not following up on this, @stephenh. I did intend to prepare a pull request, but never got around to it.

With #867 being open: How thorough is the test suite? Is it sufficient that the tests pass with the new dependency, or are further verifications necessary?

stephenh:
Np @dbrgn ! The #867 is just for the `/protos` subdirectory; I think the primary dependabot for protobuf 7 PR is this one:

https://github.com/stephenh/ts-proto/pull/866

Which is failing at the `pbjs` part of the build, which seems kinda surprising:

```
Run ./integration/pbjs.sh
Usage Error: Couldn't find a script named "pbjs".
```

Although the pbjs output is provided by the protobuf.js project, so perhaps they changed something in how that is published/invoked.

threema-lenny:
I'm on it.

PR #874, opened 2023-07-11
Update protobufjs (and peer dependencies) to ^7

threema-lenny:
I scanned the dependencies we have to protobufjs files (mostly reader/writer) and there seems to have been no changes requiring any further action.

I've also scanned for peer dependencies to protobufjs 6 and updated those accordingly (best effort).

Note: We have a cyclic dependency between `ts-proto` and `ts-proto-descriptors` whose protobufjs dependencies affect one another, so now we have two different protobufjs versions in the `node_modules` folder. Not sure how you want to handle this, @stephenh.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "license": "ISC",
   "devDependencies": {
     "@grpc/grpc-js": "^1.8.14",
-    "@grpc/proto-loader": "^0.6.13",
+    "@grpc/proto-loader": "^0.7.7",
     "@improbable-eng/grpc-web": "^0.14.0",
     "@improbable-eng/grpc-web-node-http-transport": "^0.14.0",
     "@nestjs/common": "^9.4.2",
@@ -52,6 +52,7 @@
     "mongodb": "^4.3.0",
     "nice-grpc": "^2.1.4",
     "prettier": "^2.8.8",
+    "protobufjs-cli": "^1.1.1",
     "reflect-metadata": "^0.1.13",
     "rxjs": "^7.4.0",
     "semantic-release": "^21.0.3",
@@ -65,7 +66,7 @@
     "case-anything": "^2.1.10",
     "dataloader": "^1.4.0",
     "object-hash": "^3.0.0",
-    "protobufjs": "^6.11.3",
+    "protobufjs": "^7.2.4",
     "ts-poet": "^6.4.1",
     "ts-proto-descriptors": "1.9.0"
   },
--- a/protos/package.json
+++ b/protos/package.json
@@ -12,7 +12,6 @@
     "dist/"
   ],
   "dependencies": {
-    "long": "^4.0.0",
     "protobufjs": "^6.8.8"
   },
   "devDependencies": {
--- a/protos/yarn.lock
+++ b/protos/yarn.lock
@@ -192,7 +192,6 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "ts-proto-descriptors@workspace:."
   dependencies:
-    long: ^4.0.0
     protobufjs: ^6.8.8
     ts-proto: ^1.146.0
     typescript: ^4.7.4
--- a/yarn.lock
+++ b/yarn.lock
@@ -273,6 +273,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/parser@npm:^7.20.15":
+  version: 7.22.7
+  resolution: "@babel/parser@npm:7.22.7"
+  bin:
+    parser: ./bin/babel-parser.js
+  checksum: 02209ddbd445831ee8bf966fdf7c29d189ed4b14343a68eb2479d940e7e3846340d7cc6bd654a5f3d87d19dc84f49f50a58cf9363bee249dc5409ff3ba3dab54
+  languageName: node
+  linkType: hard
+
 "@babel/plugin-syntax-async-generators@npm:^7.8.4":
   version: 7.8.4
   resolution: "@babel/plugin-syntax-async-generators@npm:7.8.4"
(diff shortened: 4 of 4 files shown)


threema-lenny:
There is one failing test where I'd appreciate getting some help for because I'm unfamiliar with grpc and the intended behaviour. For what it's worth, downgrading `@grpc/proto-loader` to the previous version makes no difference.

stephenh:
This is great @threema-lenny !

This failing test:

```
FAIL integration/nestjs-simple/nestjs-simple-test.ts
  ● nestjs-simple-test nestjs › should findOneHero

    expect(received).toEqual(expected) // deep equality

    - Expected  - 14
    + Received  +  1

      Object {
        "birthDate": Object {
          "nanos": 2,
          "seconds": 1,
        },
    -   "externalData": Object {
    -     "fizz": 1,
    -     "foo": "bar",
    -     "nested": Object {
    -       "arr": Array [
    -         1,
    -         "foo",
    -         Array [
    -           "bar",
    -         ],
    -       ],
    -       "isFailing": false,
    -     },
    -   },
    +   "externalData": Object {},
```

Is probably due to our `google.protobuf.Struct` well known type handling--iirc we try to turn pb `Struct` into "just objects", but our conversion process for that is bespoke to ts-proto, and not the just "stock protobuf bytes to protobuf message" that `protobufjs` expects.

So far we've been able to intercept `protobufjs` with their wrappers API and tell it about our special struct handling with this line in `hero.ts`:

```
wrappers[".google.protobuf.Struct"] = { fromObject: Struct.wrap, toObject: Struct.unwrap } as any;
```

And `externalData` is typed as a `.google.protobuf.Struct` in the `hero.proto` file.

So, worst-case scenario is that they removed this wrappers API entirely in protobufjs 7, or best-case scenario is that the API and/or maybe even just the import/export changed (maybe due to a bundling/packaging change from 6 to 7?), but really not sure...

Maybe that's enough guidance for you to run with it from there?

Also fwiw/fyi that @boukeversteegh is our resident Struct/well-known-types effort, just as an FYI to him that our `wrappers` approach has some currently-unknown, maybe-easy/maybe-not issue in pbjs 7.

Thanks!

threema-lenny:
I think I'll need to pass on that task because I simply don't know anything about gRPC, NestJS and that feature in protobufjs. It would be a total poke in the dark. Sorry.

stephenh:
@threema-lenny Np! I poked around a little bit, thinking that maybe protobufjs broke the `wrappers` contract/feature during the upgrade, but looks like it was just caused by having two versions of `protobufjs` in `node_modules`, specifically the `@grpc/proto-loader` (which NestJS uses) had kept an older/shadowed version of protobuf, and so was using its global `wrapper` const instead of the `wrapper` const that `hero.ts` configured.

I pushed up the `yarn.lock` change and :crossed_fingers: we'll see if that's it.

stephenh (approved):
Awesome! Thanks @threema-lenny !

threema-danilo:
Thanks @stephenh and @threema-lenny!

dankeboy36:
Hello,

thank you for the work!

After installing `ts-proto@1.153.0`, I still have the dependency to `protobufjs@6.8.8` via `ts-proto-descriptors`. The `package.json` pulled from npm is different than in the repo.

https://github.com/stephenh/ts-proto/blob/ce064d06955730ffb86ea449942ac17aaf578fa0/protos/package.json#L2-L3

https://github.com/stephenh/ts-proto/blob/ce064d06955730ffb86ea449942ac17aaf578fa0/protos/package.json#L15

Script to reproduce:

```sh
% rm -rf ts-proto-test \
&& mkdir ts-proto-test \
&& cd ts-proto-test \
&& npm init -y \
&& npm i -S ts-proto@1.153.0 \
&& cat ./node_modules/ts-proto-descriptors/package.json | jq .dependencies
Wrote to /Users/akos.kitta/Desktop/dev/__trash/ts-proto-test/package.json:

{
  "name": "ts-proto-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [],
  "author": "",
  "license": "ISC"
}



added 25 packages, and audited 26 packages in 2s

1 package is looking for funding
  run `npm fund` for details

1 high severity vulnerability

To address all issues, run:
  npm audit fix

Run `npm audit` for details.
{
  "long": "^4.0.0",
  "protobufjs": "^6.8.8"
}
```

`npm audit` still shows the `protobufjs` vulnerability:

```sh
% npm audit
# npm audit report

protobufjs  6.10.0 - 7.2.3
Severity: high
protobufjs Prototype Pollution vulnerability - https://github.com/advisories/GHSA-h755-8qp9-cq85
fix available via `npm audit fix`
node_modules/ts-proto-descriptors/node_modules/protobufjs

1 high severity vulnerability

To address all issues, run:
  npm audit fix
```

What am I doing wrong? Thank you!

threema-lenny:
You're probably doing nothing wrong, I suspect it's the cyclic dependency I hinted at in #834. I'm assuming we need to bump `ts-proto-descriptors` to depend on `ts-proto@1.153.0`, then publish it, then bump `ts-proto` to depend on `ts-proto-descriptors@<next>` and then publish that again. Quite the chore. :sweat_smile:

stephenh:
Yep! I forgot about that; I think it should be fixed with this PR that I just merged:

https://github.com/stephenh/ts-proto/pull/876

stephenh:
Thanks for the bug report @dankeboy36 ; I just pushed out a fix, can you see if that works? Thanks!

dankeboy36:
It's working great now with `ts-proto@1.153.3`. Thank you!

PR #878, opened 2023-07-13
feat: Normalize `toJSON` output by omitting fields set to their default values

haines:
This PR implements the "possible future behavior" described in the README by omitting fields set to their default values from the `toJSON` output.

I haven't made this configurable to avoid adding yet another option. Let me know if you feel like this should not be the default.

### DIFF
--- a/integration/angular/simple-message.ts
+++ b/integration/angular/simple-message.ts
@@ -48,7 +48,9 @@ export const SimpleMessage = {
 
   toJSON(message: SimpleMessage): unknown {
     const obj: any = {};
-    message.numberField !== undefined && (obj.numberField = Math.round(message.numberField));
+    if (message.numberField !== 0) {
+      obj.numberField = Math.round(message.numberField);
+    }
     return obj;
   },
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -2071,9 +2071,11 @@ function generateToJson(
       }
     } else if (isRepeated(field)) {
       // Arrays might need their elements transformed
+      const transformElement = readSnippet("e");
+      const maybeMap = transformElement.toCodeString([]) !== "e" ? code`.map(e => ${transformElement})` : "";
       chunks.push(code`
         if (message.${fieldName}?.length) {
-          ${jsonProperty} = message.${fieldName}.map(e => ${readSnippet("e")});
+          ${jsonProperty} = message.${fieldName}${maybeMap};
         }
       `);
     } else if (isWithinOneOfThatShouldBeUnion(options, field)) {
--- a/src/main.ts
+++ b/src/main.ts
@@ -2049,8 +2049,8 @@ function generateToJson(
 
       if (ctx.options.useMapType) {
         chunks.push(code`
-          ${jsonProperty} = {};
-          if (message.${fieldName}) {
+          if (message.${fieldName}?.size) {
+            ${jsonProperty} = {};
             message.${fieldName}.forEach((v, k) => {
               ${jsonProperty}[${i}] = ${readSnippet("v")};
             });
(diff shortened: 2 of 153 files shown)


integration/angular/simple-message.ts
```diff
@@ -48,7 +48,9 @@ export const SimpleMessage = {
 
   toJSON(message: SimpleMessage): unknown {
     const obj: any = {};
-    message.numberField !== undefined && (obj.numberField = Math.round(message.numberField));
+    if (message.numberField !== 0) {
```
stephenh, line 51: Hi @haines ; just sanity checking, but you think we're good to drop the `!== undefined` check from before?

I know the `numberField` is typed as `number` anyway, so the check was being overly cautious, but I'm thinking we explicitly added it b/c of concerns some prior user/contributor had...

Dunno, personally I'm fine dropping it without a strong opinion / test case asserting why we should keep it, but just wanted to see if you had specific thoughts/strong/soft opinions on it. Wdyt?

stephenh (commented):
This looks great! I haven't had time to really read through the `main.ts` changes, but all of the codegen output changes seem reasonable, and the tests are passing, so all indications are good! :-)

Should be able to get through the `main.ts` review in the next day or so...

Thanks for the PR!

integration/angular/simple-message.ts
haines: Yeah, personally I think it makes sense to drop it.

As far as I can tell the only reason this check would be necessary is if there is a bug in the calling code that circumvents the type system. `fromJSON` correctly populates the default value, and we have `fromPartial` for manually constructing messages without having to specify every field.

I'm not convinced it is worthwhile to guard against bugs in the calling code. We don't have other runtime type checks (e.g. checking that the values that are supposed to be numbers aren't actually strings), so should this be any different?

There was a test case [dating back to 2019](https://github.com/stephenh/ts-proto/commit/666a96c789b2b2eeaa4c0db77b7b9d83644c8461#diff-27827f36c0c9cddb97f98e7a65dc3495360678cde9dc87e0964e3a87957cb4dc) that exercised the checks by doing `Simple.toJSON({} as Simple)`, but I really don't think this is a valid use case (the type cast is incorrect) so [I removed it](https://github.com/stephenh/ts-proto/pull/878#discussion_r1262744555).

`useOptionals` is there when you want to allow things to be `undefined` in general, otherwise it is straightforward to call `fromPartial` before calling `toJSON` if you somehow have an object without the default values populated.

src/main.ts
```diff
@@ -1976,9 +1976,7 @@ function generateToJson(
     const readSnippet = (from: string): Code => {
       if (isEnum(field)) {
         const toJson = getEnumMethod(ctx, field.typeName, "ToJSON");
-        return isWithinOneOf(field)
```
stephenh, line 1979: This clean up is great! Unfortunately it is rare when this `readSnippet`/etc code gets simpler, so really appreciate it.

stephenh (approved):
This looks great, thanks @haines !

michaelrosett:
@stephenh 

We just updated our version to v1.55.0 and picked up this change. Unfortunately it broke us and I'm trying to figure out if we need an option here (which I understand you want to avoid). Our unique use case is that we have a proto as our primary schema language and we generate (1) typescript code for runtime use and (2) a json schema that allow us to define yaml configurations based on the proto schema (I would have liked to use text protos but unfortunately I couldn't find any good tooling for them). We process multiple yaml configurations in typescript, merge them and generate a combined yaml configuration.

This has been working great until this change. Why? We have a boolean value that we define as required using proto annotations for our json schema generator. So now, we have a required boolean field in json schema and the default value being produced by typescript code is omitted. So when our json schema validator runs against our combined yaml files, it is finding the required boolean missing.

The quick fix is to avoid the required proto annotation on boolean fields. However, this is not a great user experience for our engineers writing yaml configurations because it doesn't force them to set these boolean values. Additionally, yaml auto-gen features in vscode, for example, leave out these very important boolean fields that often need to be set to true but are missed because it only include required fields.

If there's some way for me to configure my yaml output to include default values I would prefer that option. I'd even been good if the toJSON method had an options argument that tell it whether or not to include default values in the resulting JSON. What do you think? This allows you to avoid a protoc option and moves the logic to runtime.

stephenh:
Hi @michaelrosett ; you're right I'm hesitant to add yet-another-flag, but :shrug: ts-proto can't really escape it's "you can probably make it do whatever you want" approach at this point.

If you want to add a flag that turned:

```
    if (message.stringValue !== "") {
      obj.stringValue = message.stringValue;
    }
```

Back into `!== undefined`, on top of @haines 's PR (which did some clean up that should hopefully make this easier), that sgtm.

Something like `emitJsonDefaultValues` that defaults to false, but you could set it to `true` to restore the prior behavior.

PR #884, opened 2023-07-17
feat: added encode-only mode for JSONs

Frabat:


Frabat:
Will reopen with same changes

PR #886, opened 2023-07-18
feat(ts-proto-#859): added encode-only options to toJSON methods

Frabat:
Now we can use the toJSON and fromJSON methods separately.


haines:
I just found myself needing this exact feature 😄 

I think it needs to be documented here:

https://github.com/stephenh/ts-proto/blob/44e8e8aee16c4df245ae45ab6f4b660d3f219a8e/README.markdown?plain=1#L375-L377

src/options.ts
```diff
@@ -46,7 +46,7 @@ export type Options = {
   fileSuffix: string;
   importSuffix: string;
   outputEncodeMethods: true | false | "encode-only" | "decode-only" | "encode-no-creation";
-  outputJsonMethods: boolean;
+  outputJsonMethods: true | false | "encode-only" | "decode-only";
```
stephenh, line 49: @Frabat wdyt about calling these `"to-only"` and `"from-only"`, given that the methods that are output are `toJSON` / `fromJSON`.

stephenh (approved):
This is great @Frabat ! Thanks for the PR!

I can merge it now, but will wait just a bit to see if you'd like to do the readme update that @haines mentioned.

src/options.ts
Frabat: I'm completely ok with that! I wanted to keep the same options as the outputEncodeMethods field, but that's totally doable :D 

Will update the readme too.

stephenh (approved):
Looks great! Thanks!

Issue #897, opened 2023-08-05
stringEnums removeEnumPrefix not working together


cs-clarence:
The generated code when both stringEnums and removeEnumPrefix set to true sometimes conflicts.

For example, the generated code for the enum:
```ts
export enum Gender {
  UNSPECIFIED = "GENDER_UNSPECIFIED",
  MALE = "GENDER_MALE",
  FEMALE = "GENDER_FEMALE",
  OTHER = "GENDER_OTHER",
}
```

However, somewhere in the generated code, they reference GENDER_UNSPECIFIED key that doesn't exist, instead of unprefixed one.
```ts
function createBaseChangeGenderRequest(): ChangeGenderRequest {
  return { gender: Gender.GENDER_UNSPECIFIED, other_gender: undefined };
}

encode(message: ChangeGenderRequest, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
  if (message.gender !== Gender.GENDER_UNSPECIFIED) {
    writer.uint32(8).int32(genderToNumber(message.gender));
  }
  if (message.other_gender !== undefined) {
    writer.uint32(18).string(message.other_gender);
  }
  return writer;
},

```

stephenh:
Ah shoot. Fwiw the PR that introduced `removeEnumPrefix` was this one:

https://github.com/stephenh/ts-proto/pull/779/files

And that "not default" check is from around in here:

https://github.com/stephenh/ts-proto/blob/2639818ab3fa3e996e8f799970f886137094a254/src/main.ts#L1424

So `notDefaultCheck` should probably be using the remove-prefix-aware helper method from #779.

If you'd like to work on at PR @cs-clarence , that'd be great! Thanks!

stephenh:
Fixed in #902 , thanks @buzap !

PR #902, opened 2023-08-15
fix: fixed default value for enum fields when remove-enum-prefix and string-enums both are on

buzap:
Hi, 

This PR fixes the problem reported in issue #897. Please let me know if there is any code that needs to be changed.


stephenh:
Looks great, thanks @buzap !

Issue #789, opened 2023-03-06
removeEnumPrefix doesn't work with enum defined in messages


Frabat:
Hi, I was trying to generate some enums and I've noticed that when an enum is defined in a message, removeEnumPrefix doesn't work.

```
enum Foo {
    FOO_UNSPECIFIED = 0;
    FOO_BAR = 1;
    FOO_BAZ = 2;
}

```
```
message Nested {
    enum Foo {
        FOO_TEST = 0;
        FOO_TEST2 = 1;
    }
    enum Nested_Foo {
        NESTED_FOO_TEST=0;
        NESTED_FOO_TEST2=1;
    }
    enum Nested_Bar {
        NESTED_BAR_TEST=0;
        NESTED_BAR_TEST2=1;
    }
}
enum Bar {
    BAR_UNSPECIFIED = 0;
    BAZ = 1;
    QUX = 2;
}
```

outputs as 
```

export enum Foo {
  UNSPECIFIED = 0,
  BAR = 1,
  BAZ = 2,
}

export enum Bar {
  UNSPECIFIED = 0,
  BAZ = 1,
  QUX = 2,
}

export interface Nested {
}

export enum Nested_Foo {
  FOO_TEST = 0,
  FOO_TEST2 = 1,
}

export enum Nested_NestedFoo {
  NESTED_FOO_TEST = 0,
  NESTED_FOO_TEST2 = 1,
}


export enum Nested_NestedBar {
  NESTED_BAR_TEST = 0,
  NESTED_BAR_TEST2 = 1,
}

```

I think that the behaviour is correct for the Nested_Foo enum, whereas it should output 
```
export enum Nested_NestedFoo {
   TEST = 0;
   TEST2 = 1;
}

```
Is this an issue or did you plan the feature like this?

stephenh:
@Frabat ah no, I don't think the output you're seeing is intentional, probably just the first time someone is using `removeEnumPrefix` in nested enums.

If you'd like submit a PR that updates the behavior for nested enums, that'd be great!

Frabat:
@stephenh I'll give it a try :)

stephenh:
This was fixed by #903 , thanks @buzap !

PR #903, opened 2023-08-15
fix: remove-enum-prefix for nested enums

buzap:
Hi, 

This PR fixes issue #789 .

Previously when removing the enum value prefix, it used the `tsFullName`. This doesn't work for nested enums because for nested enums, `tsFullName` also includes the parent type name. However, the prefix we are removing only includes the enum type name (See https://buf.build/docs/lint/rules#enum_value_prefix).


stephenh:
Awesome! Thanks!

Issue #708, opened 2022-11-22
Always use Map for int64 keys


zhuscat:
```
message Foo {
  map<int64, int64> bar = 1;
}
```

is generated to

```ts
export interface Foo {
  bar: { [key: Long]: Long }
}
```

when `--ts_proto_opt=forceLong=long`

This is a type error:

```
An index signature parameter type must be 'string', 'number', 'symbol', or a template literal type.
```

stephenh:
Hey @zhuscat ; thanks for the report. That's a good find.

We already have some code that will use a JS map, which is enabled with:

```
--ts_proto_opt=useMapType=true
```

It would probably be a good idea to _always_ use the map type for longs, even if `useMapType=false`. 

If you'd like to submit a PR for that fix, that'd be great!

Otherwise just using `useMapType=true` should unblock you for now, I think.

stephenh:
This should be fixed by #905 in the next release, thanks @buzap !

PR #905, opened 2023-08-17
fix: always use Map for int64 keys (#708)

buzap:
Hi,

This PR fixes issue #708. 

As you suggested in the issue comment, I made the change to always use the Map type for a protobuf map with long keys.

Note that I had to downgrade the Prettier package from 3.0.0 to 2.8.8 (the latest version 2 release) because there is an issue preventing Jest snapshot updating: https://github.com/jestjs/jest/issues/14305.

I also added `"trailingComma": "all"` to the .prettierrc file to match the original formatting.

Please let me know if there is anything that needs to be changed.


stephenh:
This is great, thank you!

Issue #907, opened 2023-08-21
Option to log numbers when the value is 0


Frabat:
Hi! 
I'm facing an issue. I'm trying to serialize an object which has a property equal to 0.

Unluckily, when I call the .toJSON()/.encode() methods, that value gets skipped. 
This results in: 

`
        "event_type": {
            "seek_ended": {
            }
        }
`
instead of 

`
        "event_type": {
            "seek_ended": {
                "position": 0
            }
        }
`

Could we try to enable the latter instead of the former?

Thanks.

stephenh:
Hi @Frabat ; this is canonical protobuf behavior, to omit the default values when encoding to the wire.

We originally did always omit `0`, but then a contributor made the upgrade for us to follow the spec, and not put default values on the wire.

I could have swore we had an "emit default values" flag to restore the prior behavior, but I guess not.

The code to change would be around in here:

https://github.com/stephenh/ts-proto/blob/be5c22e4b7f8ee81d0ae7c50d433d768b8da66f5/src/main.ts#L1428

If you wanted to work on a PR. Thanks!

wgottschalk:
I'm running into the same issue where we want to display a json formatted version to an end user for editing and want to include the numbers. I can get a PR to add support for just to toJSON methods for now since I might not have time to handle all of the encoding changes

stephenh:
This should be fixed in #919 ; thanks @wgottschalk !

PR #919, opened 2023-09-11
feat: adds support for emitting default scalar values in json

wgottschalk:
Adds a new option for emitting default fields to support https://github.com/stephenh/ts-proto/issues/907

Note that I only added support for the `toJSON` method. I designed the option to be easily modified to accept an array of fields similar to `snakeToCamel`. I didn't support the encoders or extensions because I'm not sure how to properly test those methods. If there's any code samples, I'd be happy to take a look and implement that


stephenh:
Hey @wgottschalk ! Thanks for the PR! Apologies for the delay reviewing, have just been busy...

Fwiw I'm kind of dumb-founded that this so-obvious-in-retrospect slice point for "output default or not" already existed...it's been so long since I was deep in this code, but :shrug: , that's why we have smart & motivated contributors like yourself I guess! :tada: 

Looks like maybe there is a small options test failing, just b/c of the new option you added, I assume.

Also do you mind running format/prettier on `integration/emit-default-values-json/emit-default-values-test.ts`?

Otherwise it looks great, and I'll merge once the build is passing!

stephenh:
> I didn't support the encoders or extensions because I'm not sure
> how to properly test those methods.

I think the encoders are probably fine to leave as-is, b/c they output a binary format that a) no one sees, and b) b/c it's binary most implementations pretty strictly follow it, vs. the JSON encoding where it's got this "someone people want defaults/some don't" nuance. 

By extensions, what are you referring to? Sorry, just not following/remembering.

wgottschalk:
> Looks like maybe there is a small options test failing, just b/c of the new option you added, I assume.
> Also do you mind running format/prettier on integration/emit-default-values-json/emit-default-values-test.ts?

Just updated the snapshots and ran formatting on the directory.

wgottschalk:
> By extensions, what are you referring to? Sorry, just not following/remembering.

There's a function called generateExtension which has a call to notDefaultChecked

https://github.com/stephenh/ts-proto/blob/df0c59631d5dcb4f7c0413ac57d2045a7726b1f5/src/main.ts#L1519

https://github.com/stephenh/ts-proto/blob/df0c59631d5dcb4f7c0413ac57d2045a7726b1f5/src/main.ts#L1617-L1624

stephenh:
Looks great @wgottschalk , thanks!

PR #932, opened 2023-09-30
fix: Use 0n for BigInt(0).

stephenh:


stephenh:
Closing until we add a flag for pre-2020 runtimes.

Issue #927, opened 2023-09-29
Array message type generates invalid code on account of name collision.


jinfiesto:
I unfortunately am working with a schema where I have an Array message type. This causes issues with the generated code, as the Array message type that gets generated collides with the native Array type, see:

fromJSON(object: any): Whatever {
    return { array: Array.isArray(object?.array) ? object.array.map((e: any) => Number(e)) : [] };
  },

In this case, since I have an Array message type, that is unfortunately what is being referred to here in this code snippet. Did I miss some configuration in the docs which I can use to fix this?

Hilzu:
I ran into the same issue when working with [pg_query.proto](https://github.com/pganalyze/libpg_query/blob/15-latest/protobuf/pg_query.proto) from [libpg_query](https://github.com/pganalyze/libpg_query) but with `Boolean` and `String`. I have reproduced the issue here: https://github.com/Hilzu/ts-proto-with-pg-query#ts-proto-creates-invalid-code-for-pg_queryproto

jcready:
The simple fix here would be for ts-proto to prefix all of its global variable usage with `globalThis`. E.g.

```ts
globalThis.Array.isArray(object?.array) ? object.array.map((e: any) => globalThis.Number(e)) : []
```

stephenh:
Hey, thanks for the report; this should be fixed in #930 ... disclaimer I did not technically add a schema with `Date`, `String`, `Boolean`, and `Array` message types in it ... probably should as a regression test, but :shrug: , skipped it this time around.

But lmk if I missed any.

Another disclaimer is that the `globalThis` that ts-proto currently uses is a kind-of-ugly-but-bullet-proof fallback; we could probably stand use to a `globalThis=true` flag for modern runtimes that don't need the fallback, which could remove the polyfill. I might do that while I'm here I guess.

Hilzu:
@stephenh Adding a test with the message types might be a good idea as the fix in #930 doesn't cover all the cases. I'm still getting errors with the latest ts-proto v.1.159.1: https://github.com/Hilzu/ts-proto-with-pg-query#ts-proto-creates-invalid-code-for-pg_queryproto

```
pg_query.ts:13254:41 - error TS2349: This expression is not callable.
  Type '{ encode(message: String, writer?: Writer): Writer; decode(input: Uint8Array | Reader, length?: number | undefined): String; fromJSON(object: any): String; toJSON(message: String): unknown; }' has no call signatures.

[...]

PR #934, opened 2023-10-02
fix: Support using messages called String/Boolean/Number/Array

Hilzu:
Fixes https://github.com/stephenh/ts-proto/issues/927

* Extend global-this integration test to cover String, Boolean, Number and Array messages
* Fix remaining places that caused errors when generating code for these messages


stephenh:
Looks great, thank you!

PR #937, opened 2023-10-04
fix: toJSON methods don't respect useDate=false (#935)

### LINKED ISSUE #935: toJSON methods don't respect useDate=false
This leads to precision loss.
I'm using `useDate=false`, which affects generated objects, but doesn't affect `toJSON` method.

roboslone:
Fixes #935.


kirill-gadzhiev (approved):

stephenh (approved):
This is great @roboslone , thanks!

PR #938, opened 2023-10-04
feat: add bigint input validation

8de2fdb0:
Input validation for too large int64 and uint64 values when using
the compiler option forceLong=bigint.


stephenh:
Looks great @8de2fdb0 , thanks!

8de2fdb0:
Thanks for being so responsive, and `ts-proto` in generall.

I am a bit unsure if that was all, at least it catches all the BigInt types now if `forceLong=bigint`.

If I find any other occurrence of potential number overflows I will make a follow up MR.

PR #957, opened 2023-10-25
fix: return types and optional chaining in field masks when `useOptionals=all`

### LINKED ISSUE #956: Type error in field_mask.ts when outputServices=grpc-js and useOptionals=all
```
src/generated-grpc-js/google/protobuf/field_mask.ts(258,12): error TS18048: 'message.paths' is possibly 'undefined'.
src/generated-grpc-js/google/protobuf/field_mask.ts(277,5): error TS2322: Type 'string[] | undefined' is not assignable to type 'string[]'.
    Type 'undefined' is not assignable to type 'string[]'.
```

The interface for FieldMask gets generated as:
```ts
export interface FieldMask {
  /** The set of field mask paths. */
  paths?: string[] | undefined
}
```

And toJSON is missing optional chaining:
```ts
  toJSON(message: FieldMask): string {
    return message.paths.join(",")
  }
```

...and unwrap is returning a `string[] | undefined` where `string[]` is expected in the function signature:
```ts
  unwrap(message: FieldMask): string[] {
    return message.paths
  }
```

Looks like this requires a similar treatment as #949. I can give it a go unless I'm missing something?

timosaikkonen:
Fixes #956


stephenh:
This is great, thanks @timosaikkonen !

PR #960, opened 2023-10-31
feat: generate type namespaces for enums as literals

samualtnorman:
this pull request adds code for generating a namespace of type declarations for enums as literals that let you use them like regular enums:
old:
```ts
let foo: typeof Foo.Bar
```

new:
```ts
let foo: Foo.Bar
```

this means a library that was previously using `ts-proto` without the `enumsAsLiterals=true` can add it with less breakages.

---

`Foo.proto`
```proto
enum Foo {
	Bar = 0;
	Baz = 1;
	Qux = 2;
}
```

old generated `Foo.ts`
```ts
export const Foo = { Bar: 0, Baz: 1, Qux: 2, UNRECOGNIZED: -1 } as const;

export type Foo = typeof Foo[keyof typeof Foo];
```

new generated `Foo.ts`
```ts
export const Foo = { Bar: 0, Baz: 1, Qux: 2, UNRECOGNIZED: -1 } as const;

export type Foo = typeof Foo[keyof typeof Foo];

export namespace Foo {
  export type Bar = typeof Foo.Bar;
  export type Baz = typeof Foo.Baz;
  export type Qux = typeof Foo.Qux;
  export type UNRECOGNIZED = typeof Foo.UNRECOGNIZED;
}
```

---

I've not looked into adding any tests or anything for this change, but if this feature is something you're interested in, I wouldn't mind adding tests (and making any other changes you request)

### DIFF
--- a/integration/enums-as-literals-with-string-enums/enums-as-literals-with-string-enums.ts
+++ b/integration/enums-as-literals-with-string-enums/enums-as-literals-with-string-enums.ts
@@ -17,6 +17,14 @@ export const DividerData_DividerType = {
 
 export type DividerData_DividerType = typeof DividerData_DividerType[keyof typeof DividerData_DividerType];
 
+export namespace DividerData_DividerType {
+  export type DOUBLE = typeof DividerData_DividerType.DOUBLE;
+  export type SINGLE = typeof DividerData_DividerType.SINGLE;
+  export type DASHED = typeof DividerData_DividerType.DASHED;
+  export type DOTTED = typeof DividerData_DividerType.DOTTED;
+  export type UNRECOGNIZED = typeof DividerData_DividerType.UNRECOGNIZED;
+}
+
 export function dividerData_DividerTypeFromJSON(object: any): DividerData_DividerType {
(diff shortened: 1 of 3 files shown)


stephenh:
Hi @samualtnorman . Thanks for the PR!

What you're trying to do makes sense, i.e. emulate the previous `Foo.Bar` syntax...I don't love using the `namespace` keyword, because awhile ago it wasn't supported in...some bundler...

But this is just for the type namespace, so I think it makes sense.

If you could find an `integration-test/*` directly that is using `parameter.txt` with `enumsAsLiterals`, and just maybe update an existing test/file in their to show your output is getting generated as you like, and then also be a regression test, that'd be great.

Thanks!

samualtnorman:
what would the regression test be testing for?

stephenh:
I was thinking just something in a test file that did `someFunction(value: Foo.Bar)` to show the `Foo.Bar` syntax working, but I think the checked-in generated output should be a good enough regression test. Thanks!

samualtnorman:
thank you for accepting my change!

PR #961, opened 2023-11-03
feat: add before and after request methods to base service

lukealvoeiro:
### Description

Consumers of any generated service may want to abstract away functionality like additional error parsing or request/response logging into their Rpc implementation, as this lets them perform this shared logic across all requests and responses they have. This PR implements this change, adding two settings to ts_proto_opt:

- outputBeforeRequest: boolean (defaults to `false`)
- outputAfterResponse: boolean (defaults to `false`)

When `true`, these will allow you to define an Rpc with two additional, optional methods - `beforeRequest` and `afterResponse`

### Testing

- [x] Test conversion has been added to the integration folder
- [x] Jest tests have been set up to confirm `beforeRequest` and `afterResponse` are only called when they are set.

### Documentation

- [x] Have updated main `README.md` to include the two new settings


integration/before-after-request/before-after-request-test.ts
```diff
@@ -0,0 +1,49 @@
...
+    jest
+      .spyOn(FooServiceCreateResponse, "decode")
+      .mockReturnValue({ $type: "simple.FooServiceCreateResponse", ...exampleData });
+  });
+
+  it("performs function before request if specified", async () => {
+    const req = FooServiceCreateRequest.create(exampleData);
+    client = new FooServiceClientImpl({ ...rpc, beforeRequest: beforeRequest });
+    await client.Create(req);
+    expect(beforeRequest).toHaveBeenCalled();
```
stephenh, line 32: Can we assert against what was passed? I.e. would it be: `toHaveBeenCalledWith(req)`?

integration/before-after-request/before-after-request-test.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it("performs function after request if specified", async () => {
+    const req = FooServiceCreateRequest.create(exampleData);
+    client = new FooServiceClientImpl({ ...rpc, afterResponse: afterResponse });
+    await client.Create(req);
+    expect(afterResponse).toHaveBeenCalled();
```
stephenh, line 39: Same here; I wonder we don't have `res` in scope, but maybe just a simple subset of the keys, like `toHaveBeenCalledwith({ $type: ..., anotherField: ... })` would be good enough to know its working.

integration/before-after-request/simple.ts
```diff
@@ -0,0 +1,499 @@
...
+  constructor(rpc: Rpc, opts?: { service?: string }) {
+    this.service = opts?.service || FooServiceServiceName;
+    this.rpc = rpc;
+    this.Create = this.Create.bind(this);
+  }
+  Create(request: FooServiceCreateRequest): Promise<FooServiceCreateResponse> {
+    const data = FooServiceCreateRequest.encode(request).finish();
+    const requestMethod = messageTypeRegistry.get(request.$type);
+    if (this.rpc.beforeRequest && requestMethod) {
+      this.rpc.beforeRequest(requestMethod);
```
stephenh, line 465: I wasn't expecting to see the `messageTypeRegistry.get(request.$type)` lookup, b/c that just means the `beforeRequest` is going to be called the with metadata about the request, and not the actually `request` itself.

I.e. shouldn't this be:

```
this.rpc.beforeRequest(request);
```

And then if the caller wants to use `request.$type` to see what type of request it is, that's fine, but they also still have the request itself, to do any logging or manipulation of, as they'd like.

src/options.ts
```diff
@@ -240,6 +244,11 @@ export function optionsFromParameter(parameter: string | undefined): Options {
     options.exportCommonSymbols = false;
   }
 
+  if (options.outputBeforeRequest || options.outputAfterResponse) {
+    options.outputTypeRegistry = true;
```
stephenh, line 248: Related to my question about doing `beforeRequest(request)`, and then letting the caller do a `$type` lookup if the need to, that would also mean I think we could remove this part, but `outputBeforeRequest` / `outputBeforeResponse` nudging `outputTypeRegistry` on.

stephenh (commented):
Hey @lukealvoeiro , I had a few questions, but this is a great PR, thanks for submitting it!

integration/before-after-request/simple.ts
lukealvoeiro: Ah, yes. I think that's where I went wrong - I initially had `this.rpc.beforeRequest(request);` but got a bit confused between the `Request` function and `Request` interface, so when I was addressing the TS error I ended up moving towards the wrong functionality. Sending out an update with this fix shortly. 🙏🏼

src/generate-services.ts
```diff
@@ -351,6 +367,14 @@ export function generateRpcType(ctx: Context, hasStreamingMethods: boolean): Cod
   const maybeContextParam = options.context ? "ctx: Context," : "";
   const maybeAbortSignalParam = options.useAbortSignal ? "abortSignal?: AbortSignal," : "";
   const methods = [[code`request`, code`Uint8Array`, code`Promise<Uint8Array>`]];
+  const additionalMethods = [];
+  const is$TypePresent = options.outputTypeAnnotations || options.outputTypeRegistry;
+  if (options.outputAfterResponse) {
+    additionalMethods.push(code`afterResponse?(response: { ${is$TypePresent ? "$type: string" : ""} }): void;`);
```
lukealvoeiro, line 373: @stephenh Since the request / response type is truly dynamic, I was hoping to get your thoughts on this call signature. Options I can think of are the following:

```typescript
// OPTION 1:
beforeRequest: (request: {}): void; // or
beforeRequest: (request: { $type: string }): void;

// OPTION 2: using a index signature
beforeRequest: (request: {[property: string]: any)}: void;

// OPTION 3: using any or object 😢 
beforeRequest: (request: object): void;
```

stephenh: Hm, yeah, I dunno, isn't `request: object` kinda normal for these middleware-esque patterns?

I.e. if you want to listen to "any message going in/out", there really isn't a common/base interface for all messages.

To your point, unless the user has the type registry option turned on, then they'll at least have the `$type` key and can do some "if message.$type === ..." checks.

So maybe `request: object` by default, by `request: { $type: string }` if the type registry feature is enabled?

Fwiw I'm not super-close to the nuances of `object` vs. the index signature, so if you've done some WIP coding of these handlers, and the index signature makes the handler code more ergonomic, I'm good with that too!

lukealvoeiro: ended up going with:
```typescript
afterResponse?<T extends { [k in keyof T]: unknown }>(response: T): void;
```

stephenh:
Cool, lgtm, thanks @lukealvoeiro !

PR #965, opened 2023-11-17
feat: add error handler to rpc interface

lukealvoeiro:
### Description

Extending off of my previous [PR](https://github.com/stephenh/ts-proto/pull/961), I have added in additional 'middleware' to the default service `rpc` interface in the form of a method called `errorHandler`. This method catches any errors that may have occurred during the entire lifecycle of a RPC request. While ultimately you will still need to return some Error from the `handleError` method, consumers of the service may leverage this functionality to:
- return errors that are consistent with their API style guide
- log error messages

#### Other noteworthy additions to this PR:
- Added a `tryCatchBlock` helper to the `utils.ts` file to make it easier to create try-catch blocks in the future.
- Fixed a bug where setting either `outputBeforeRequest` or `outputAfterResponse` to `true` caused you to only be able to output the `default` service - and thus, not have access to the service `generic-definitions`.
- Added `service: string` and `method: string` as parameters to the `beforeRequest` and `afterResponse` methods in the `rpc` interface. These can be used to handle specific services or methods in a special way.
- Small change to the jest config in `package.json` to allow the VSCode-Jest debugger to attach correctly. Note: Happy to revert this if there is a reason why it's set.

### Testing

- [x] Added two new integration tests, to test `outputErrorHandler` by itself and the interaction with `outputAfterResponse`
- [x] Updated `before-after-request` to include `generic-definitions`
- [x] Updated documentation for `outputBeforeRequest`, `outputAfterResponse`, and `outputErrorHandler`


integration/handle-error-with-after-response/simple.ts
```diff
@@ -0,0 +1,188 @@
...
+            throw this.rpc.handleError(this.service, "GetBasic", error);
+          }
+          throw error;
+        }
+      });
+    } catch (error) {
+      if (error instanceof Error && this.rpc.handleError) {
+        throw this.rpc.handleError(this.service, "GetBasic", error);
+      }
+      throw error;
```
stephenh, line 163: @lukealvoeiro wdyt of instead of `throw error`, we make this `return Promise.reject(error)`?

My rationale is that, as-is, because this `GetBasic` method isn't using the `async` keyword, errors that are thrown won't actually turn into rejected promises, which is typically what callers except.

README.markdown
```diff
@@ -446,9 +446,11 @@ Generated code will be placed in the Gradle build directory.
 
 - With `--ts_proto_opt=outputServices=false`, or `=none`, ts-proto will output NO service definitions.
 
-- With `--ts_proto_opt=outputBeforeRequest=true`, ts-proto will add a function definition to the Rpc interface definition with the signature: `beforeRequest(request: <RequestType>)`. It will will also automatically set `outputTypeRegistry=true` and `outputServices=true`. Each of the Service's methods will call `beforeRequest` before performing it's request.
+- With `--ts_proto_opt=outputBeforeRequest=true`, ts-proto will add a function definition to the Rpc interface definition with the signature: `beforeRequest(service: string, message: string, request: <RequestType>)`. It will will also automatically set `outputServices=default`. Each of the Service's methods will call `beforeRequest` before performing it's request.
```
stephenh, line 449: @lukealvoeiro This is admittedly a little late, since we've already released these `outputBefore/After` flags, but wdyt of renaming all of these to:

* `rpcBeforeRequest=true`
* `rpcAfterResponse=true`
* `rpcErrorHandler=true`

The disclaimer/mea culpa is that I've not been great at having a super-consistent naming convention for flags, we just kind of make them up as we go, but given these are all related to the `Rpc` interface, that seems like a good prefix?

stephenh (commented):
Hi @lukealvoeiro , these changes look good/make sense, few questions but otherwise yep will merge after you respond back. Thanks!

integration/handle-error-with-after-response/simple.ts
lukealvoeiro: ah true, good suggestion - will make this change!

README.markdown
lukealvoeiro: No worries, those names are definitely better. Thanks for the quick 👀 on this PR :)

stephenh:
Looks great, thanks @lukealvoeiro !

PR #968, opened 2023-11-23
fix: Don't close client if we've already aborted

bouk:
grpc-web throw an error if you close the client after it's already been closed:

https://github.com/improbable-eng/grpc-web/blob/1d9bbb09a0990bdaff0e37499570dbc7d6e58ce8/client/grpc-web/src/client.ts#L327

This can happen with the abort signals if the observer is finished first and then the abort signal is triggered, causing an exception. Here I'm changing the code generator to remove the abort signal listener in the observer finish handler.

### DIFF
--- a/integration/grpc-web-abort-signal/example.ts
+++ b/integration/grpc-web-abort-signal/example.ts
@@ -1081,17 +1081,23 @@ export class GrpcWebImpl {
             }
           },
         });
-        observer.add(() => {
-          if (!abortSignal || !abortSignal.aborted) {
-            return client.close();
-          }
-        });
 
         if (abortSignal) {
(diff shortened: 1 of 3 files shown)


stephenh (approved):

stephenh:
Cool, thanks sense!

bouk:
@stephenh thx for the quick review! Would appreciate a release

stephenh:
@bouk ah shoot, those are supposed to happen automatically but looks like the commit message was missing the `fix: ` prefix (kinda annoying but github won't make the PR title overwrite the commit message when there is only a single commit).

I just force-pushed your commit with the `fix:` prefix added, so it should auto-release in the next few minutes.

PR #969, opened 2023-11-23
fix: revert useDate=false behaviour; add useJsonTimestamp option

### LINKED ISSUE #964: useDate=false doesn't match protobuf JSON spec
https://github.com/stephenh/ts-proto/issues/935

I'll reiterate here -

from your very own `README.md`:

```
fromJSON/toJSON use the [proto3 canonical JSON encoding format](https://developers.google.com/protocol-buffers/docs/proto3#json) (e.g. timestamps are ISO strings), unlike https://github.com/protobufjs/protobuf.js/issues/1304.
```

`toJSON()` is strictly meant to generate protojson matching json representations.

It's fine if someone wants to keep `useDate=false` for object representation it's fine (like we do), but mostly - a thing of its own.
Language / Desreialized representation is OK  as it's contained to the runtime code. Other services interoping with this service will never be aware of how the app chooses to represent the binary data. JSON repersentation is a cross-language concern, as everything that speaks protojson expects a very specific protojson representation

This fix **completely breaks the protobuf agenda of any language any platform compatibility, as protojsons can now be ts-proto compatible only**

I believe there was a reason why `useDate` didn't apply to json representations in the first place.

We have a gRPC based JVM/Python/Node JS ecosystem and we used `useDate=false`, but relied on the json repesentation to be protojson compatible to communicate across apps on certain scenarios. **This fix completey broke everything for us and we needed to downgrade the ts-proto version ATM**.

I think we should parameterize this option, or create another param `(useDateJson=false)` to allow for better compatiblity.

Please help

I'm willing to contribute a PR, if required.

Idan

roboslone:
Fixes #964


stephenh:
This looks great @roboslone ! Thank you!

PR #974, opened 2023-12-06
fix: do not mention :local scripts

Demon000:
They have been removed in 3219cdfc91976fa9149ff1f0962296f20968569a.


stephenh:
Nice! Thank you!

PR #980, opened 2023-12-26
fix: add serviceName to grpc-js client constructor type

le0nik:
Currently type exported by generated code does not match the return value of `makeConstructor` function from grpc-js

https://github.com/grpc/grpc-node/blob/c5d35fe22db9503fb7ec2f03370cfc1723efe6e8/packages/grpc-js/src/make-client.ts#L124

because it's missing  the `serviceName` field

https://github.com/grpc/grpc-node/blob/c5d35fe22db9503fb7ec2f03370cfc1723efe6e8/packages/grpc-js/src/make-client.ts#L94

This PR adds `serviceName` field to the generated type.


stephenh:
Hi @le0nik thanks for the PR!

PR #981, opened 2023-12-28
feat: add useDate=string-nano option

le0nik:
This PR introduces a new option, `useDate=string-nano`. This feature allows the generation of ISO strings with nanosecond precision, such as `1970-01-01T14:27:59.987654321Z`. The conversion is handled by the [nano-date](https://www.npmjs.com/package/nano-date) library.


stephenh:
This looks great! Thanks @le0nik !

PR #982, opened 2023-12-30
fix: Allow other services with nestJs.

### LINKED ISSUE #972: Allow NestJS to leverage Generic definitions
More a question than anything.  Is it possible to have `nestJs=true` not ignore the `outputServices` parameter?  Or, is there a specific option that will give me the portion of `genericDefinitions` that has the "definition" which looks like 
```typescript
export type SomeServieDefinition = typeof SomeServiceDefintion;

export const SomeServicesDefintition = {
  name: 'SomeService',
  fullName: '....',
  methods: {
    ....
  }
}
``` 

We have a utility that we wrote that takes the output encoders/decoders and uses the definition above to adjust the output definitions from `proto-loader` so that Nest is able to leverage the encoders and decoders from `ts-proto`.  I've tinkered with a lot of the options and I wasn't able get my desired result.  I was able to determine that the output I want comes from the `genericDefintions` option, however.

stephenh:
Fixes #972

### DIFF
--- a/src/options.ts
+++ b/src/options.ts
@@ -1,5 +1,3 @@
-import { parse } from "path";
-import { Code } from "ts-poet";
 import { ToStringOpts } from "ts-poet/build/Code";
 
 export enum LongOption {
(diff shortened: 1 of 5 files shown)


src/options.ts
```diff
@@ -202,15 +200,19 @@ export function optionsFromParameter(parameter: string | undefined): Options {
...
-
-  if (options.outputServices.length == 0) {
+  // Assume the user wants the default service output, unless they're using nestJs, which has
+  // its own controllers output (although nestjs users can ask for other services too).
+  if (options.outputServices.length == 0 && !options.nestJs) {
     options.outputServices = [ServiceOption.DEFAULT];
   }
+  // If using nestJs + other services, add the encode methods back
+  if (options.nestJs && options.outputServices.length > 0) {
+    options.outputEncodeMethods = true;
```
Pikachews, line 214: This ends up running even when `outputServices=false`.

Pikachews:
Also the README should be updated: https://github.com/stephenh/ts-proto/blob/c0068bac164aabe15088e2fa220b5ba56bdfbbb7/README.markdown?plain=1#L425

PR #983, opened 2023-12-31
fix: error handling on non-`Error` type errors

lukealvoeiro:
### Description

Small issue noticed with the behavior of the default service implementation `handleError` method. If the throw error was an instance of Error, then the handleError function is called correctly. However, if it was not, we had a check that prevented `handleError` from called. This isn't desired behavior as errors can be strings. 

### Testing performed

Tested in a consuming application and this issue is resolved. Unit tests have been updated and are passing.


stephenh:
Makes sense, thanks @lukealvoeiro !

PR #986, opened 2024-01-17
fix: support of importSuffix for index files

### LINKED ISSUE #845: outputIndex does not respect importSuffix
when --importSuffix=.js the index file generated by outputIndex doesn't not add .js suffix to imports, which leads to TypeScript error when targeting node>16

ts-node version: 1.148.2

arnidan:
Closes #845


stephenh:
This is great @arnidan , thank you!

arnidan:
@stephenh thank you for fast response! 

Could you also make a release? 
Seems like @semantic-release-bot won't help this time

https://github.com/stephenh/ts-proto/actions/runs/7561024481/job/20588434774#step:6:53

stephenh:
Ah yeah, sorry, the PR check is supposed to catch that--for now I just amended your commit to add `fix:` as a prefix, and it should get released soon!

PR #989, opened 2024-01-22
feat: Allow comments to be omitted

haines:
We ship generated code as part of a package, but it's an internal implementation detail so the comments that get copied from the proto files are just wasted bytes.

This adds a `comments` option which defaults to `true` (maintaining the current behaviour), but may be set to `false` to omit comments from the generated code.


stephenh (approved):
Looks great @haines , thank you!

Issue #908, opened 2023-08-23
Error TS2306 (not a module) for empty files


jk-apple:
**NOT A CONTRIBUTION**

With `esModuleInterop=true` and `outputIndex=true` ts-proto sometimes produces empty files which are then exported from an index file which leads to error TS2306 (not a module). 

When running `protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto --ts_proto_opt=esModuleInterop=true --ts_proto_opt=outputIndex=true --ts_proto_out=out/ empty.proto` on
```proto
// empty.proto
syntax = "proto2";
package empty;
```
we get
```
$ head out/{empty.ts,index.empty.ts}
==> out/empty.ts <==
/* eslint-disable */

==> out/index.empty.ts <==
/* eslint-disable */

export * from "./empty";
```
See https://github.com/microsoft/TypeScript/issues/46343 for a discussion of why this causes a compiler error when importing ``index.empty.ts``.

Empty protobuf files like shown are just one example of how to trigger that error.
More commonly, this also happens with protobuf files that purely consist of ``extend`` declarations.

stephenh:
Ah yeah, that makes sense...thanks for the report @jk-apple . If you'd like to submit a PR to fix this, that'd be great. Thanks!

PR #992, opened 2024-01-25
fix: generate modules for empty files with esModuleInterop

jk-apple:
Fixes https://github.com/stephenh/ts-proto/issues/908.

### DIFF
--- a/integration/extension-import/extension-import-test.ts
+++ b/integration/extension-import/extension-import-test.ts
@@ -1,0 +1,13 @@
+import * as path from "node:path";
+import * as fs from "node:fs";
+import * as ts from "typescript";
+
+describe('extension-only-files', () => {
+  it("generate as external module", () => {
+    const generatedPath = path.join(__dirname, "extension.ts");
+    const generatedCode = fs.readFileSync(generatedPath, "utf8");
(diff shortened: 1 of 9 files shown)


integration/extension-import/extension-import-test.ts
```diff
@@ -0,0 +1,13 @@
+import * as path from "node:path";
+import * as fs from "node:fs";
+import * as ts from "typescript";
+
+describe('extension-only-files', () => {
+  it("generate as external module", () => {
+    const generatedPath = path.join(__dirname, "extension.ts");
+    const generatedCode = fs.readFileSync(generatedPath, "utf8");
+    const source = ts.createSourceFile(generatedPath, generatedCode, ts.ScriptTarget.ES2018);
```
stephenh, line 9: I appreciate the test @jk-apple , but I think in this case our "lazy" approach of just checking in the `extension.ts` file will serve as a good-enough regression + probably resolve the CI error that's failing to find the file to assert against.

If you could push that up, that'd be great; appreciate the fix!

jk-apple: Thanks for the help, pushed it.

stephenh:
Looks great, thanks!

PR #996, opened 2024-02-02
fix: ensure default service streaming methods compile when middleware methods are enabled

lukealvoeiro:
### Description

When we codegen default service implementations with the `--rpcBeforeRequest`, `--rpcAfterResponse` or `--rpcErrorHandler` options , we weren't correctly handling the streaming client side and server side methods. This PR fixes that issue.

### Testing

- [x] Ensure we have unit tests to cover this and they are passing
- [x] Create an implementation folder to cover this case


stephenh:
Looks great! Thanks!

Issue #967, opened 2023-11-21
env=browser treated same as env=both. Causes globalThis.Buffer typescript error


davidfiala:
When outputting with `env=browser` you get the same output as `env=both`.

This is generally OK, except when you have strict typescript checks on. You end up in a situation where the generated code includes a line like: `if (globalThis.Buffer)` to see if Buffer exists. This makes sense for `env=both`, but will cause typescript errors on a browser environment because globalThis type does not include `Buffer`.

As a workaround, we can enable: both of: `env=browser,globalThisPolyfill=true`.

However, it seems to me that a cleaner approach might be to have the `env` enum disable checks for `globalThis.Buffer` altogether. In searching through today's ts-proto code, I cannot see any cases where the `env` is checked against anything except `node`. Thereby accidentally making `both` equivalent to `browser`, which is probably not the original intent.

Some proposals:

1) **preferred**: `env=browser` will stop checking for globalThis.Buffer
2) If `env=browser` is detected, then we auto-enable the polyfill.

jonaskello:
For me ts-proto generates code like this:
```ts
if (globalThis.Buffer) 
```

Which leads to typescript error:

```
error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.
```

I want to support both node and browser with the same code. What would be the correct setting?

stephenh:
> I want to support both node and browser with the same code. What would be the correct setting?

@jonaskello I guess for `env=both`, we could make the snippet be `if ((globalThis as any).Buffer)` and that will remove the TS warning.

If you'd like to submit a PR to do that, that'd be great!

I'm going to close this issue since #999 fixes the behavior with env=browser.

jonaskello:
Got it. If I can find the time I'll make a PR for that.

I was able to work around it for now. We have all the generated files in a separate package with it's own tsconfig so I could disable the strict any setting for those files without affecting the rest of the app.

stephenh:
@jonaskello it was an easy fix, so I went ahead and pushed out a change to add the `as any` to the globalThis.Buffer check.

jonaskello:
@stephenh Thanks for the quick fix. I tried `1.167.7` and it works for the `base64FromBytes` function but not for the `bytesFromBase64` function. So I created  PR #1005 with the same fix for that function.

PR #999, opened 2024-02-13
fix: don't reference globalThis.Buffer when env=browser (#967)

cmd-johnson:
I stumbled over #967 today. The code generated by ts-proto includes references to `globalThis.Buffer`, even though the `env` option is set to `browser`.

While the transpiled code still works in the browser, TypeScript (when using strict mode) complains that `error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.`.

The fix is relatively simple: update the `makeByteUtils` function in `main.ts` and only emit the implementations matching chosen environment:

- for `env=node`, the emitted `bytesFromBase64` and `base64FromBytes` functions now exclusively use `globalThis.Buffer` to de-/encode to and from base64
- for `env=browser`, `globalThis.btoa` / `globalThis.atob` are used
- for `env=both`, the emitted functions work as before: they check for `globalThis.Buffer` and use either the node or browser implementations accordingly


stephenh:
This is great @cmd-johnson , thank you!

PR #1000, opened 2024-02-14
fix: import fails when folder name overlaps with file name

lukealvoeiro:
### Description

This PR shows the error that occurs when the a proto file imports a file named the same thing as its containing folder. 

#### Example of the old codegen behavior

Example:
```
<root>/
├─ ui/
│  ├─ simple.proto (imports ui.proto)
├─ ui.proto
```
This results in generated Typescript such as:
```ts
import {
  Simple as Simple2,
  SimpleEnum as SimpleEnum1,
  simpleEnumFromJSON as simpleEnumFromJSON3,
  simpleEnumToJSON as simpleEnumToJSON4,
} from "./";
```
This is not valid Typescript.

If you use the `fileSuffix` or `importSuffix` options, this issue disappears.

#### Fix

This [PR](https://github.com/stephenh/ts-poet/pull/55) in `ts-poet` fixes the issue


lukealvoeiro:
@stephenh Do you have any ideas on what may be going on here? Spent a few hours on it but didn't manage to track down the issue.

stephenh:
Hi @lukealvoeiro ; great job diagnosing & fixing this! Lmk when you've got the ts-poet bump incorporated here, and we can merge the PR, with the test case to keep it from regressing. Thank you!

stephenh:
This is great! Thanks for the test!

PR #1005, opened 2024-02-18
fix: Use as any on globalThis.Buffer check for bytesFromBase64

jonaskello:
This is a follow up PR for #1004 which also add `as any` for the `bytesFromBase64` function to make it work fully with strict tsconfig setting. See #967.


jonaskello:
I'm not sure about what happend with the `integration/extension-import/*` files. They seem to not exist before I ran `yarn bin2ts`. So now they are added as new files with this PR but they are not really in scope for this PR. Could it be that they were missing or did I make a mistake somewhere?

jonaskello:
@stephenh Perhaps I should remove the generated `integration/extension-import/*` files from this PR as they seem to make the build fail?

jonaskello:
I went ahead and removed the extra files that was generated by `yarn build:test` as they are not related to this PR.  You can see which ones it was in [this commit](https://github.com/stephenh/ts-proto/pull/1005/commits/b8261faf133bd827b44b1a97489ff1b686dd6569)

stephenh:
Ah, I was the passing checks, so merged your PR without seeing the comments about the extra files. :-)

But, yeah, looks like a prior PR must not have ran the right build incantation; I'll do a chore PR to get those added. Thanks for calling it out!

Issue #993, opened 2024-01-29
Should deprecated fields be optional? 


husseinhere:
Given the message:

```
message Amount {
  Currency currency = 1;
  int64    old_value = 2 [deprecated = true];
  double   value = 3;
}
```

ts-proto generates:
```
export interface Amount {
    currency: Currency;
    /** @deprecated */
    oldValue: number;
    value: number;
}
```

But wouldn't it be a lot nicer if `oldValue` was defined as  `oldValue?: number` ?

stephenh:
That seems reasonable; if you want to try adding that check here, we could see what happens:

https://github.com/stephenh/ts-proto/blob/main/src/types.ts#L356

stephenh:
Closing this out thanks to #1010, thank you @tufandevrim !

PR #1010, opened 2024-02-28
feat: support deprecatedOnly option to make deprecated fields optional

tufandevrim:
Add `deprecatedOnly` option for `useOptionals` option to make only deprecated fields optional. This PR aims to address the issue https://github.com/stephenh/ts-proto/issues/993


tufandevrim:
@stephenh thoughts on this PR appreciated. I figured `deprecatedOnly` in `useOptionals` config is practical solution but we could also have this as separate option. 

Right now in our codebase, we don't use `useOptional` config which forces developers to set the fields if it is marked as "required" (without optional). Even though we rarely depreciate required fields in our schemas, it is very annoying that we must still set deprecated required fields!

integration/use-optionals-deprecated-only/test.ts
```diff
@@ -0,0 +1,640 @@
...
+    | undefined;
+  /** @deprecated */
+  description?:
+    | string
+    | undefined;
+  /** @deprecated */
+  data?: Uint8Array | undefined;
+  repId: number[];
+  /** @deprecated */
+  repState: StateEnum[];
```
paralin, line 64: Should this be `repState?` since it is deprecated? Maybe I am not understanding the purpose of the flag.

stephenh: It looks like there is a check to say that repeated fields should not be marked as optional 
... @tufandevrim was there a reason to not include this for repeateds?

stephenh:
Hi @tufandevrim this seems pretty reasonable, see the one question about repeateds, but otherwise this lgtm!

integration/use-optionals-deprecated-only/test.ts
tufandevrim: @paralin @stephenh so sorry for the delay. My original plan was keeping repeated fields out of `deprecatedOnly` config but it really makes sense to make the repeated fields also optional if they are deprecated. I just updated the code, please review it when you get a chance. Thanks again for the review and feedback.

tufandevrim:
> Hi @tufandevrim this seems pretty reasonable, see the one question about repeateds, but otherwise this lgtm!

@stephenh yeah that was great feedback and made sense. Just updated the code to make deprecated repeated fields optional. Thanks again

stephenh:
@tufandevrim looks like a minor prettier error in `main.ts`; can you fix it up? Thanks!

tufandevrim:
> @tufandevrim looks like a minor prettier error in main.ts; can you fix it up? Thanks!

Oops! fixed now!

stephenh:
Sweet, thanks @tufandevrim !

PR #1013, opened 2024-03-08
feat: allow `$type` to be optional

lukealvoeiro:
### Description

Add another possible value to the `outputTypeAnnotations` option: `optional`. This makes type definitions on interfaces optional, which may be useful if you want to use the `$type` field for runtime type checking on responses from a server but don't want to have to set the $type on each request message you create.

### Testing

Ensured this causes no regressions for existing integration tests. Also added new integration tests to prevent future regressions.


stephenh (approved):
This makes sense. I worry a little bit that it shows ts-proto should have a clearer distinction of "read types" vs. "creation types".

I.e. my assertion/understanding would be that:

* Anyone _reading_ / accepting a message type would generally prefer `$type` to not be optional, and always be required, b/c it simplifies their code, without having to check "if type set/not set".

* Anyone _creating_ a message type would probably prefer to leave `$type` unset, and like a `SomeMessage.create({ ... })` fill in the `$type`for them, similar to how `create` can set field defaults.

In theory, ts-proto does have "read types" (just the `interface`) and "write types" (the `Partial`s that are accepted by `SomeMessage.create`, but I think in practice the allure / primary feature of ts-proto is precisely that message types are "just interfaces" (not the usual classes that protobuf frameworks usually use), and the "just interfaces" approach is just kinda fundamentally at odds with the "read type vs. write type" nuance.

So, dunno, just some "fwiw" background on why I'd preferred to keep `$type` required so far, but it's a really simple change, and so whatever you find most ergonomic for your codebase sgtm!

Thanks!

(I'll approve & merge this PR as a formality, but if you're interested feel free to start approving/self-merging your own PRs (once they're passing tests / you feel comfortable with the changes / etc); it'd be great to have another maintainer helping out with the project!)

Issue #1014, opened 2024-03-08
Provide Options type for execution in TS


dasco144:
I've created a TypeScript node script, that will execute `protoc` with `ts-proto`

I feel like it would be a nice idea to provide the Options type (provided [here](https://github.com/stephenh/ts-proto/blob/main/src/options.ts#L41)) in the npm package so that we can ensure that the options passed can have an extra layer of type safety/checking.

stephenh:
Hi @dasco144 ; ah, that makes sense. I think nearly all users end up invoking ts-proto via a protoc command line call, so historically I've not really worried about this.

But if this is as simple as `export`-ing the `Options` type, so you could do:

```
import { Options } from ts-proto;
```

That sounds good to me! If you could submit a PR for it, that'd be great, thanks!

PR #1027, opened 2024-04-12
feat: export options types

dasco144:
Closes https://github.com/stephenh/ts-proto/issues/1014


stephenh:
Great PR @dasco144 , thank you!

PR #1030, opened 2024-04-16
feat: Add js type support

### LINKED ISSUE #958: Support for FieldOptions.JSType
Reference: https://protobuf.dev/reference/java/api-docs/com/google/protobuf/DescriptorProtos.FieldOptions.JSType

Hi!

In my proto file I have:
```
int64 foobar = 3 [jstype = JS_STRING];
```

In my generated ts file I get:
```
foobar: number;
```

Using the arguments:
```
--ts_proto_opt=onlyTypes=true
```

It should be this?
```
foobar: string;
```

Is there a configuration setting I am missing?

dasco144:
Closes #958


stephenh:
Hi @dasco144 , thanks for the PR! Just wanted to say it'll probably be ~few days/this weekend before I get a chance to look at it.

dasco144:
> Hi @dasco144 , thanks for the PR! Just wanted to say it'll probably be ~few days/this weekend before I get a chance to look at it. 

Thanks! Let me know if I need to change anything 💪

src/types.ts
```diff
@@ -184,10 +190,26 @@ export function packedType(type: FieldDescriptorProto_Type): number | undefined
...
+): FieldDescriptorProto_Type | undefined {
+  if (!options.useJsTypeOverride || field.options?.jstype === undefined) {
+    return;
+  }
+
+  switch (field.options.jstype) {
+    case FieldOptions_JSType.JS_STRING:
+      return FieldDescriptorProto_Type.TYPE_STRING;
+    case FieldOptions_JSType.JS_NUMBER:
+      return FieldDescriptorProto_Type.TYPE_INT64;
```
stephenh, line 205: Do we need to handle the JS_NORMAL value here?

https://googleapis.dev/nodejs/nodejs-functions/latest/google.protobuf.FieldOptions.html

If not, maybe put an in-source comment about why not? Like maybe if `JS_NORMAL` is used, the `jstype` ends up being `undefined` because `JS_NORMAL` is an 0th value enum (just guessing!)...

src/main.ts
```diff
@@ -2357,7 +2392,11 @@ function generateFromPartial(ctx: Context, fullName: string, messageDesc: Descri
     const objectProperty = getPropertyAccessor("object", fieldName);
 
     const readSnippet = (from: string): Code => {
-      if ((isLong(field) || isLongValueType(field)) && options.forceLong === LongOption.LONG) {
+      if (
+        (isLong(field) || isLongValueType(field)) &&
+        options.forceLong === LongOption.LONG &&
+        !isJsTypeFieldOption(options, field)
```
stephenh, line 2398: Just curious, but we don't have any other `isJsTypeFieldOption` checks in here, but I imagine maybe that's because the types will line-up (for anything not this forceLong case?) and so we don't need any special handling?

stephenh (approved):
Hi @dasco144 , this is great! The test coverage is especially amazing, and appreciated! Just a few questions, mostly for my own understanding, but otherwise lgtm!

src/main.ts
dasco144: Yeah, exactly that. These cases will fall through to the `isPrimitive` check below, and just output reading from the variable as is (without any special method calls or handling)

src/types.ts
dasco144: From my understanding, the `JS_NORMAL` case would be equivalent to not having the `JSType` set on the field at all, so we want the behaviour to match the same case as if it were undefined. That's why I made it fallthrough and default to not returning an explicit value.
I'll add a quick comment explaining this 👍
I could also maybe move it into the guard clause above

stephenh:
Sweet, this is amazing @dasco144 , thank you!

dasco144:
I realised I did not update the `[README.markdown](https://github.com/stephenh/ts-proto/blob/main/README.markdown)` explaining this setting. I'll create a new PR to add this

Issue #531, opened 2022-03-09
Add option to allow for a generated code comment


adayNU:
We'd like for GH / Reviewable to understand these files are generated. Would you be open to adding a flag like: `--ts_proto_opt=header=true`

Which would prepend something like `// Code generated by protoc-gen-ts. DO NOT EDIT.` to the first line of every file.

If this seems reasonable, I'd be happy to implement.

stephenh:
@adayNU yep, that sounds like a good idea; a PR would be great!

boukeversteegh:
It could even be there by default. It's a comment so it will be stripped in most cases by webpack anyway.

stephenh:
Yep, good point / +1 to a hard-coded value. We could someday have like a `prefixFile=./my-custom-prefix.txt` opt to override it, but I assume for now we could drop basically any known string into `prefixDisableLinter` and be good enough.

dasco144:
I think this might be a nice feature to add. We could include additional information, like how the `protoc-gen-go` plugin does

```ts
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.29.0
// 	protoc        v4.22.2
// source: foo.proto
```

stephenh:
Better late than never :-) thank you @dasco144 ! :tada:

PR #1037, opened 2024-04-30
feat: add generated code comments

dasco144:
Closes https://github.com/stephenh/ts-proto/issues/531


dasco144:
Looks like the build output gets changed due to the inc;lusion of the `package.json` as a reference.

image

I assume we'd want to retain the flat structure, so let me see what I can find to fix it.

src/utils.ts
```diff
@@ -326,3 +319,20 @@ export function withAndMaybeCheckIsNotNull(options: Pick<Options, "useNullAsOpti
...
+export async function getVersions(request: CodeGeneratorRequest) {
+  let protocVersion = "unknown";
+  if (request.compilerVersion) {
+    const { major, minor, patch } = request.compilerVersion;
+    protocVersion = `v${major}.${minor}.${patch}`;
+  }
+
+  const path: string = "../package.json";
+  const packageJson = await import(path);
+  const tsProtoVersion = `v${packageJson?.version ?? "unknown"}`;
```
dasco144, line 332: This might be a bit of a hack. When providing the path to the dynamic import directly, `tsc` will include it into the compilation, which leads to the build output having the code in the `src` sub folder. 
By moving it out, and typing it as a string, it can no longer infer that path, and then won't include it in the output.

The path is still valid on the published npm package, from running yarn pack and checking the output.

stephenh: Ah cool, good find!

I'm honestly a little surprised the `await import` just works; do we have to worry about whether ts-proto is invoked in a CJS/ESM environment? I guess this passed the build on node 16, so it must be working.

If users report issues, we could also look at doing a post-tsc step that uses `sed`/something to replace a hard-coded `const tsProtoVersion = "placeholder"` type line with the just-being-published version, in the `.js` file, before upload to npm.

But I'm good with using this and seeing how it goes!

stephenh (approved):

PR #1039, opened 2024-05-09
fix: `outputSchema=true` when `onlyTypes=true`

lukealvoeiro:
This PR aims to fix a conflict that is created between two options that results in code that fails to compile. Please note that this is not an attempt to fix the underlying problem, which I have explained in a TODO comment in the code for posterity. I also created new tests to capture any regressions.


stephenh:
Hi @lukealvoeiro ! Definitely appreciate the todo/test coverage to at least document behavior.

Looks like the build is failing, but is it fall just version bumps of protoc? Lol, that would be unfortunate :thinking: . Wonder how we can handle that cleanly... Might have to make that output optional, if even just for our test suite?

lukealvoeiro:
@stephenh This should be all good to merge!

PR #1040, opened 2024-05-13
feat: optionally output versions used to generate files

lukealvoeiro:
### Description

- Add the `annotateFilesWithVersion` option so that changes are correctly captured in CI (eliminating noise on each file)
- Added this option to the README


stephenh:
This is great, thanks @lukealvoeiro !

PR #1046, opened 2024-05-24
fix: camelToSnake to respect uppercase words, such as "GetAPIValue" -> "GET_API_VALUE"

BrandonArmand:
## Problem
`camelToSnake()` is converting the example string "TestAPI" -> "TEST_AP_I" when converting a 3rd party integration proto files that I am attempting to generate with the Nest option. I believe that the conversion should respect the "API" part of the string and not split it into two parts like so. 

## Fix
I swapped the logic to use the already imported "case-anything" library's `snakeCase()` function as `snakeCaseAnything()` and added a couple of simple tests to ensure the logic.

I would love for this fix to go through as it is a pretty inconvenient problem at the moment (:

Thanks!

### DIFF
--- a/src/case.ts
+++ b/src/case.ts
@@ -1,4 +1,4 @@
-import { camelCase as camelCaseAnything } from "case-anything";
+import { camelCase as camelCaseAnything, snakeCase as snakeCaseAnything } from "case-anything";
 
 import { Options } from "./options";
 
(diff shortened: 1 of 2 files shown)


stephenh:
Hi @BrandonArmand , thanks for the PR! Can you remove the `package-lock.json` and `yarn.lock` changes? Thanks!

stephenh:
Fwiw my guess is that the build failures are because a dependency like ts-poet or prettier accidentally got unlocked/bumped to latest with the package-lock/yarn.lock changes.

BrandonArmand:
Removed `package.lock` and reverted `yarn.lock`, and squashed the commits to cleanup the git logs regarding those changes.

stephenh:
Looks great! Thanks!

PR #1047, opened 2024-05-29
feat: `no-file-descriptor` setting for outputSchema option

lukealvoeiro:
### Description

When the outputSchema setting is enabled, it generated an object containing a `fileDescriptor` entry. Depending on the proto file, the associated value can be very large. This PR allows consumers of `ts-proto` to reap the benefits of the output schema, while minimizing the overall size of the object by removing the `fileDescriptor` entry.

### Testing

- [x] Created a new integration directory with a unit test to capture regressions
- [x] Existing unit test cases pass


lukealvoeiro:
@stephenh I've noticed these diffs being generated even on main. Can you confirm whether you experience this too? Or... maybe its just an issue when I run this locally.

stephenh:
@lukealvoeiro yeah, I had noticed it failing on `main` ... finally was able to get it fixed in #1050 . I think if you rebase this PR, things should work! :crossed_fingers:  Thank you!

PR #1051, opened 2024-05-31
feat: add option `noDefaultsForOptionals`

robmonie:
Adds a new option - `noDefaultsForOptionals` that allows for compatibility with [Square/Block's Wire Protobuf implementation](https://square.github.io/wire/) with respect to defaulting of optional fields.

Note: As per the documentation added to the readme, this option provides non-standard protobuf behaviour so it would be understandable if this isn't something you'd like in the library.  I'm hoping with the appropriate warning in the docs that you'll consider this as as option.


src/main.ts
```diff
@@ -1249,10 +1248,14 @@ function generateDecode(ctx: Context, fullName: string, messageDesc: DescriptorP
               ${messageProperty} = ${generateMapType ? "new Map()" : "{}"};
             }`
           : "";
+
+        const ifValueCheck = `${varName}.value !== undefined ${withAndMaybeCheckIsNotNull(options, `${varName}.value`)}`;
+        const maybeIfKeyCheck = `${options.noDefaultsForOptionals ? ` && ${varName}.key !== undefined ${withAndMaybeCheckIsNotNull(options, `${varName}.key`)}`: ''}`
```
robmonie, line 1253: This change to `map` exists solely to satisfy the types. Now that both key and value are potentially undefined, assignment against an undefined key isn't valid so a check against both value and key is required.  Is there possibly a better way to think about this?

integration/extension-import/base.ts
```diff
@@ -0,0 +1,82 @@
+// Code generated by protoc-gen-ts_proto. DO NOT EDIT.
```
robmonie, line 1: This and other related files in this folder we generated from the existing `base.proto` as part of performing a build for tests.

stephenh:
Hi @robmonie ! Sorry about that `extension-import` thing; I just fixed this up on `main`, can you try rebasing the PR and see if that works? Thanks!

stephenh:
I looked into this a bit more, and huh, I had not heard of Wire/didn't know about their "primitives are always boxed" behavior. That is certainly different! Although ngl I think I prefer it; protobuf's optional behavior has been such a bug bear over the years.

Normally I'd want to keep this out of ts-proto, but the change is so relatively tiny, to our codebase, that :shrug: I guess it seems fine to me.

@robmonie just curious, but does Square actually have public-facing APIs that use the Wire protocol/behavior, or have you just seen/used Wire's behavior, and prefer it, and so would like to use it on your own/non-Square projects? Just curious!

robmonie:
Thanks for looking at this @stephenh. I'm not sure whether there are any externally facing apis that use Wire, however pretty much all the internal Apis, Events etc are based on protos written in Kotlin which use Wire as an internal standard. This makes interfacing with unsupported languages difficult.  

The biggest issue comes from the default proto behaviour of not encoding default values such as `false` or `0`, so when the client is based on ts-proto and the service is based on Wire, null values are not defaulted and as a consequence are interpreted as null, essentially making interoperability impossible unless it's accounted for on the Kotlin side. Unfortunately we interface with many preexisting apis that can't or shouldn't be changed to account for us as a client.

I'll do a rebase shortly and if there are any improvements you think I can make, let me know.

stephenh:
Looks great @robmonie , thanks!

robmonie:
> Looks great @robmonie , thanks!

Thanks @stephenh. Much appreciated :-)

PR #1057, opened 2024-06-06
fix: Add check for lower bound with forceLong=number

timostamm:
With default plugin options, or `useJsTypeOverride` for a field with option `jstype = JS_NUMBER`, 64-bit integral values are converted to a JavaScript Number when decoding from binary. When a value is too large for Number, an error is thrown.

But signed integers can also be too _small_ for Number. This PR adds a check for the lower bound.

### DIFF
--- a/integration/wrappers-regression/google/protobuf/timestamp.ts
+++ b/integration/wrappers-regression/google/protobuf/timestamp.ts
@@ -140,7 +140,7 @@ export const Timestamp = {
             break;
           }
 
-          message.seconds = longToNumber(reader.int64());
+          message.seconds = longToNumber(reader.int64() as Long);
           continue;
         case 2:
           if (tag !== 16) {
--- a/integration/wrappers-regression/google/protobuf/timestamp.ts
+++ b/integration/wrappers-regression/google/protobuf/timestamp.ts
@@ -140,7 +140,7 @@ export const Timestamp = {
             break;
           }
 
-          message.seconds = longToNumber(reader.int64() as Long);
+          message.seconds = longToNumber(reader.int64());
           continue;
         case 2:
           if (tag !== 16) {
--- a/integration/wrappers-regression/google/protobuf/timestamp.ts
+++ b/integration/wrappers-regression/google/protobuf/timestamp.ts
@@ -3,6 +3,7 @@
 
 /* eslint-disable */
 import * as _m0 from "protobufjs/minimal";
+import Long = require("long");
 
 export const protobufPackage = "google.protobuf";
 
@@ -199,8 +200,8 @@ type KeysOfUnion<T> = T extends T ? keyof T : never;
 export type Exact<P, I extends P> = P extends Builtin ? P
   : P & { [K in keyof P]: Exact<P[K], I[K]> } & { [K in Exclude<keyof I, KeysOfUnion<P>>]: never };
 
-function longToNumber(int64: { toString(): string }): number {
-  const num = globalThis.Number(int64.toString());
+function longToNumber(long: Long): number {
+  const num = long.toNumber();
   if (num > globalThis.Number.MAX_SAFE_INTEGER) {
     throw new globalThis.Error("Value is larger than Number.MAX_SAFE_INTEGER");
   }
@@ -208,6 +209,11 @@ function longToNumber(int64: { toString(): string }): number {
     throw new globalThis.Error("Value is smaller than Number.MIN_SAFE_INTEGER");
   }
   return num;
+}
+
+if (_m0.util.Long !== Long) {
+  _m0.util.Long = Long as any;
+  _m0.configure();
 }
 
 function isSet(value: any): boolean {
--- a/integration/wrappers-regression/google/protobuf/timestamp.ts
+++ b/integration/wrappers-regression/google/protobuf/timestamp.ts
@@ -201,14 +201,13 @@ export type Exact<P, I extends P> = P extends Builtin ? P
   : P & { [K in keyof P]: Exact<P[K], I[K]> } & { [K in Exclude<keyof I, KeysOfUnion<P>>]: never };
 
 function longToNumber(long: Long): number {
-  const num = long.toNumber();
-  if (num > globalThis.Number.MAX_SAFE_INTEGER) {
+  if (long.gt(globalThis.Number.MAX_SAFE_INTEGER)) {
     throw new globalThis.Error("Value is larger than Number.MAX_SAFE_INTEGER");
   }
-  if (num < globalThis.Number.MIN_SAFE_INTEGER) {
+  if (long.lt(globalThis.Number.MIN_SAFE_INTEGER)) {
     throw new globalThis.Error("Value is smaller than Number.MIN_SAFE_INTEGER");
   }
-  return num;
+  return long.toNumber();
 }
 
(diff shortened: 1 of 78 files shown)


src/main.ts
```diff
@@ -510,29 +509,33 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
   const longToString = conditionalOutput(
     "longToString",
     code`
-      function longToString(long: ${Long}) {
-        return long.toString();
+      function longToString(int64: { toString(): string }) {
+        return int64.toString();
```
timostamm, line 513: Long, bigint, number, and string all share the toString() method.

src/main.ts
```diff
@@ -510,29 +509,33 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
...
     `,
   );
 
   const longToBigint = conditionalOutput(
     "longToBigint",
     code`
-      function longToBigint(long: ${Long}) {
-        return BigInt(long.toString());
+      function longToBigint(int64: { toString(): string }) {
+        return typeof int64 == "bigint" ? int64 : ${bytes.globalThis}.BigInt(int64.toString());
```
timostamm, line 522: Same as above.

src/main.ts
```diff
@@ -510,29 +509,33 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
...
+      function longToNumber(int64: { toString(): string }): number {
+        const num = ${bytes.globalThis}.Number(int64.toString());
+        if (num > ${bytes.globalThis}.Number.MAX_SAFE_INTEGER) {
           throw new ${bytes.globalThis}.Error("Value is larger than Number.MAX_SAFE_INTEGER")
         }
-        return long.toNumber();
+        if (num < ${bytes.globalThis}.Number.MIN_SAFE_INTEGER) {
+          throw new ${bytes.globalThis}.Error("Value is smaller than Number.MIN_SAFE_INTEGER")
+        }
+        return num;
```
timostamm, line 538: long.gt is actually not necessary to perform the check. I tested it with the script below, comparing the previous version against this version with values surrounding MAX_SAFE_INTEGER, and select values up to the max value for signed 64 integers.

<details><summary>test.ts</summary>

```ts
import Long from "long";

// The current implementation in ts-proto, using long.gt() to check bounds
function currentLongToNumber(input: string): number {
  const long = Long.fromString(input);
  if (long.gt(globalThis.Number.MAX_SAFE_INTEGER)) {
    throw new globalThis.Error("Value is larger than Number.MAX_SAFE_INTEGER");
  }
  return long.toNumber();
}

// A more simple implementation that uses numbers only to check bounds
function betterLongToNumber(input: string): number {
  const num = Number(input);
  if (num > Number.MAX_SAFE_INTEGER) {
    throw new globalThis.Error("Value is larger than Number.MAX_SAFE_INTEGER");
  }
  return num;
}

const MAX_SIGNED_INT64 = 9_223_372_036_854_775_807n;
const MAX_SAFE_INTEGER = BigInt(Number.MAX_SAFE_INTEGER);
const tests: bigint[] = [];

// Add tests around Number.MAX_SAFE_INTEGER
for (let i = -100; i < 100; i++) {
   tests.push(BigInt(i) + MAX_SAFE_INTEGER);
}

// Add tests up to MAX_SIGNED_INT64
for (let i = MAX_SAFE_INTEGER + 100n; i < MAX_SIGNED_INT64;) {
  tests.push(i);
  i = i + (MAX_SIGNED_INT64 - MAX_SAFE_INTEGER) / 1_000_000n;
}
tests.push(MAX_SIGNED_INT64);


// Make sure both functions behave the same for each test
for (const test of tests) {
  let a: unknown;
  try {
    a = currentLongToNumber(test.toString());
  } catch (e) {
    a = e;
  }
  let b: unknown;
  try {
    b = betterLongToNumber(test.toString());
  } catch (e) {
    b = e;
  }
  const ok = (a instanceof Error && b instanceof Error)
    ? a.message === b.message
    : a === b;
  if (!ok) {
    throw new Error(`case ${test}: ${String(a)} !== ${String(b)}`);
  }
}

console.log(tests.length, "tests passed");
```
</details>

src/main.ts
```diff
@@ -1117,20 +1120,20 @@ function getDecodeReadSnippet(ctx: Context, field: FieldDescriptorProto) {
...
           case FieldOptions_JSType.JS_STRING:
-            readSnippet = code`${utils.longToString}(${readSnippet} as Long)`;
+            readSnippet = code`${utils.longToString}(${readSnippet})`;
             break;
         }
       } else if (options.forceLong === LongOption.LONG) {
         readSnippet = code`${readSnippet} as Long`;
       } else if (options.forceLong === LongOption.STRING) {
-        readSnippet = code`${utils.longToString}(${readSnippet} as Long)`;
+        readSnippet = code`${utils.longToString}(${readSnippet})`;
```
timostamm, line 1132: No need to cast anymore, except for forceLong=long.

integration/wrappers-regression/google/protobuf/timestamp.ts
```diff
@@ -200,16 +199,15 @@ type KeysOfUnion<T> = T extends T ? keyof T : never;
...
+  const num = globalThis.Number(int64.toString());
+  if (num > globalThis.Number.MAX_SAFE_INTEGER) {
     throw new globalThis.Error("Value is larger than Number.MAX_SAFE_INTEGER");
   }
-  return long.toNumber();
-}
-
-if (_m0.util.Long !== Long) {
-  _m0.util.Long = Long as any;
-  _m0.configure();
```
stephenh, line 212: @timostamm might be a moot point in the long run, but I think this "hook up Long to protobufjs and call configure" is required, and had been using a ts-poet-ism to get like auto-inserted into the output on the conditional use of `Long` in the output...

So the otherwise-lgtm change of using the `{ toString(): string }` meant ts-poet doesn't know to implicitly add this anymore (which is admittedly kind of cute/too-cute on how its gets added).

Am I missing something, where this isn't actually required? The intent was just to have it "just work" for users, and not need to do this themselves. If we should keep it, we can probably figure out another way to get it conditionally into the output...

src/main.ts
stephenh: Nice, I like the approach!

integration/wrappers-regression/google/protobuf/timestamp.ts
timostamm: Let's add it back 👍  I'll take a look tomorrow.

timostamm: Uh, I remember what's going on with protobuf.js and Long.js, and I wish I didn't.

`protobufjs` automagically loads `long` from node_modules, and uses it in `Reader`. If the magic fails (easily happens with bundlers), it falls back to an alternative and returns Number instead of Long.

We could hook the ConditionalOutput to `Reader.int64` instead. Then we're free to refactor longToNumber.

timostamm:
I don't think it's worth going through the the long conversion refactor, see https://github.com/stephenh/ts-proto/pull/1057#discussion_r1631113981.

Updated the PR to just add the check for lower bounds.

stephenh (approved):

PR #1058, opened 2024-06-10
feat!: Replace protobuf.js

timostamm:
This PR replaces [protobuf.js](https://www.npmjs.com/package/protobufjs) with [@bufbuild/protobuf](https://www.npmjs.com/package/@bufbuild/protobuf) in generated code. The libraries are used for low-level wire encoding. Since they have a similar API, the changes to main.ts are relatively simple.

The main advantage is that Long.js is no longer a dependency, unless you explicitly _want_ to use Long.js (plugin option `forceLong=long`).

The signatures of `encode` and `decode` change - they no longer use `Reader` and `Writer` from protobuf.js, but the counterparts from `@bufbuild/protobuf`. This will be a breaking change for users that manually pass in readers, or rely on the returned writer to be a specific type.

Follow-ups:
- ~Replace `Reader` in src/schema.ts~ https://github.com/stephenh/ts-proto/pull/1058/commits/0d763e98ad96abded801a45d1b53370564f9950d
- ~Re-generate code for `ts-proto-descriptors`~ https://github.com/stephenh/ts-proto/pull/1058/commits/3ae88891b08d21597f085bdaabbf1404e9e44c9e
- ~Move `protobufjs` from dependencies to devDependencies~ https://github.com/stephenh/ts-proto/pull/1058/commits/7ef0c73543706c069a9e189ac02511316495afda


src/main.ts
```diff
@@ -505,40 +488,23 @@ function makeLongUtils(options: Options, bytes: ReturnType<typeof makeByteUtils>
...
-    `,
-  );
-
   const longToNumber = conditionalOutput(
     "longToNumber",
     code`
-      function longToNumber(long: ${Long}): number {
-        if (long.gt(${bytes.globalThis}.Number.MAX_SAFE_INTEGER)) {
+      function longToNumber(int64: { toString(): string }): number {
+        const num = ${bytes.globalThis}.Number(int64.toString());
```
timostamm, line 495: I've verified that this check is reliable in https://github.com/stephenh/ts-proto/pull/1057#discussion_r1630141662.

stephenh (approved):

timostamm:
@stephenh, I took care of the follow-ups. 

This is using an alpha of @bufbuild/protobuf v2. I promise we won't break it if you want to publish as is. A stable v2 is about a month out in case you prefer to wait.

`ts-proto-descriptors` should be published first - the `ts-proto` package.json has a dep on it.

stephenh:
Amazing, thank you @timostamm !

timostamm:
A beta of `@bufbuild/protobuf` is available, and it won't be long until it's stable. I'll update this PR once the stable v2 is published.

stephenh:
Awesome! Thanks for the update @timostamm !

oliveryasuna:
@stephenh Is this really a good choice to move away from the official implementation? At least for me, I moved away from protoc-gen-ts because it uses the outdated [google-protobuf](https://www.npmjs.com/package/google-protobuf).

corwinsheahan-wf:
@timostamm @stephenh It looks like `@bufbuild/protobuf` v2 stable has been released. Any chance we can now get this PR merged/released?

stephenh:
This is amazing, as usual, @timostamm , thank you for the bump to the latest!

I merged this to `main`, with a semantic-release change to hopefully push it out as an alpha release, and we can see if that works! :crossed_fingers:

stephenh:
Just wanted to say thanks again @timostamm ! Really appreciate the help on this PR, and looking forward to @bufbuild/protobuf being a much better runtime for us! :pray:

timostamm:
Sorry, I was out for a couple of days. This was fun 😃 Congratulations on the v2 release!

PR #1065, opened 2024-06-27
docs: Describe `oneof=unions-value` output as Algebraic Data Type

probeiuscorp:
The discriminated union output of `oneof=unions` and `oneof=unions-value` is incorrectly described as an Abstract Data Type. Describe as Algebraic Data Type instead.


stephenh:
Ah nice catch, thank you!

Issue #1084, opened 2024-08-09
Feature Request: add options to limit generating encode and decode methods to only specific message types


akrennmair:
I'm facing the specific problem that a significant amount of the generated script size of a project mine consists of scripts generated by ts-proto. I was looking into limiting what is being generated, and found the `outputEncodeMethods` option to be too limited: generally speaking, I have two top-level types, and for one (the request type) I only want to generate an `encode` method, while for the other (the response type), I only want to generate a `decode` method.

I would therefore like to have options to limit the generation of `encode` resp. `decode` methods to only specific types. A similar suggestion has also been made as a comment to #773: https://github.com/stephenh/ts-proto/issues/773#issuecomment-1422208778

I also have a branch in which I implemented a possible solution that simply regular expressions. I'm not sure though whether this is sufficient as a general solution. I will submit a PR for it.

stephenh:
Hi @akrennmair ; thanks for the PR! I guess that makes sense...

Out of curiosity, for your use case, is each type only primitives? Or does it have other messages?

I.e. I could see something like:

* `message Person` -> you want to encode `Person`, but
* `message Person` has a `address = Address` field and a `pets = repeated Pet` field

Such that to "encode Person", you actually end up ending `encode` methods for ~4-5 different `message`s, which seems like it would make keeping the regex updated kind of tedious? Or do you think that wouldn't be a problem?

stephenh:
Hi @stefan-koshiw , thanks for the ping; #1085 does lgtm. For some reason I don't see the "Approve & Run" button that I usually click for PRs to have the GitHub actions run -- @akrennmair can you try rebasing/updating your PR, and see if that fixes things? Once I can get the CI build passing, it lgtm and we can merge it. Thanks!

stephenh:
Fixed in #1085 , thanks @akrennmair !

PR #1085, opened 2024-08-09
feat: Add options to limit generation of encode and decode methods to only specific message types

akrennmair:
I created this PR as a possible to solution to the feature request submitted in #1084.

I added integration tests for test coverage, and updated the documentation to include the two new options.


stephenh:
Hi @akrennmair , thanks for updating the PR! Looks like there is a build error with `MessageFns`...maybe we need two/separate input/output `MessageFns` types?

akrennmair:
@stephenh I didn't want to touch the `MessageFns` type as I didn't know what it would break. What I did instead was to generate a tiny dummy encode or decode method that throws an error if the presence of the method is required.

stephenh:
Sounds great, thanks @akrennmair !

Issue #1086, opened 2024-08-12
Bug in toJSON Function with `removeEnumPrefix=true` and `unrecognizedEnumValue=0` Option Combination


kanziw:
I encountered a bug in the `toJSON` function generated by ts-proto when using both the `removeEnumPrefix=true` and `unrecognizedEnumValue=0` options. The issue occurs with the following proto definition:

```proto
enum Quality {
  QUALITY_UNSPECIFIED = 0;
  QUALITY_STANDARD = 1;
  QUALITY_HD = 2;
}
```

When these options are applied, the generated `qualityToJSON` function looks like this:

```ts
export function qualityToJSON(object: Quality): string {
  switch (object) {
    case Quality.UNSPECIFIED:
      return "QUALITY_UNSPECIFIED";
    case Quality.STANDARD:
      return "QUALITY_STANDARD";
    case Quality.HD:
      return "QUALITY_HD";
    default:
      return "UNSPECIFIED";
  }
}
```

The problem lies in the `default` case, where the function returns `"UNSPECIFIED"` instead of `"QUALITY_UNSPECIFIED"`. This leads to incorrect JSON serialization for unrecognized enum values. The `default` case should return `"QUALITY_UNSPECIFIED"`.

stephenh:
Hi @kanziw , you're right, that sounds like a bug! 

If you could add a test case, maybe to this test:

https://github.com/stephenh/ts-proto/tree/main/integration/remove-enum-prefix

The bug is *probably* around in here:

https://github.com/stephenh/ts-proto/blob/main/src/enums.ts#L118

If you could submit a PR, that'd be great! Thanks!

PR #1089, opened 2024-08-15
fix: toJSON Function with `removeEnumPrefix=true` and `unrecognizedEnumValue=0` Options

kanziw:
close: #1086 

This PR addresses the issue described in #1086, where the `toJSON` function generated by ts-proto incorrectly serializes enum values when both the `removeEnumPrefix=true` and `unrecognizedEnumValue=0` options are used.

The expected behavior is that, even with `removeEnumPrefix=true`, the `toJSON` function should encode the enum values using their original string values, including the prefix. However, the bug causes the function to encode the enum values without the prefix, leading to incorrect JSON serialization.

To fix this, the PR updates the `toJSON` function to correctly use the original enum name, including the prefix, when encoding. This is achieved by storing the `originalName` alongside the `UnrecognizedEnum` value and using it during serialization.

This change ensures that the enum values are correctly serialized with their full names, aligning the behavior with the expected output when these options are enabled.

If this code format doesn't align with the maintainer's preferences, please feel free to modify it as needed.


kanziw:
When I first tried to contribute to this project, it took me quite some time to figure out how to write and run test cases. I would like to suggest creating a [CONTRIBUTING](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) document to share with the community, so that others can more easily contribute to the project.

Based on my experience, the document could include the following steps:
1. Create a directory under `integration` with an appropriate `TEST_NAME`.
2. Write the proto file for your test in `$TEST_NAME.proto`.
3. Specify the necessary ts-proto options in `parameters.txt`.
4. Run `yarn proto2ts $TEST_NAME` to generate the stubs.
5. Using the generated stubs, write your test cases in `$TEST_NAME-test.ts`.
6. Continue writing your code until the test passes by running `yarn test $TEST_NAME-test.ts`.

stephenh:
This is a great fix @kanziw , thank you!

Definitely hear you on the CONTRIBUTING suggestion; this `Development` section of the readme is supposed to help with that:

https://github.com/stephenh/ts-proto?tab=readme-ov-file#development

But the readme is quite long.

If you'd like to pull that `#development` section out into a CONTRIBUTING.md, and clean up it based on your fresh eyes, that would be an amazing PR as well! Thank you!

Issue #495, opened 2022-01-28
Not clear how to use the generated client implementations


torkelrogstad:
Thanks for creating what looks like a great, idomatic TS code generator. 

However, I'm having some issues getting started. I've generated a TS file from my Proto file, with (among other things), this: 

```ts
export interface HelloWorldService {
  Ping(request: PingRequest): Promise<PingResponse>;
}

export class HelloWorldServiceClientImpl implements HelloWorldService {
  private readonly rpc: Rpc;
  constructor(rpc: Rpc) {
    this.rpc = rpc;
    this.Ping = this.Ping.bind(this);
  }
  Ping(request: PingRequest): Promise<PingResponse> {
    const data = PingRequest.encode(request).finish();
    const promise = this.rpc.request(
      "helloworld.v1.HelloWorldService",
      "Ping",
      data
    );
    return promise.then((data) => PingResponse.decode(new _m0.Reader(data)));
  }
}
```

What am I supposed to stick into `new HelloWorldServiceClientImpl`? The goal here is to create a Node.js client.

torkelrogstad:
Got this working: 

```ts

const conn = new grpc.Client(
  "localhost:8765",
  grpc.credentials.createInsecure()
);

type RpcImpl = (
  service: string,
  method: string,
  data: Uint8Array
) => Promise<Uint8Array>;

const sendRequest: RpcImpl = (service, method, data) => {
  // Conventionally in gRPC, the request path looks like
  //   "package.names.ServiceName/MethodName",
  // we therefore construct such a string
  const path = `${service}/${method}`;

  return new Promise((resolve, reject) => {
    // makeUnaryRequest transmits the result (and error) with a callback
    // transform this into a promise!
    const resultCallback: UnaryCallback<any> = (err, res) => {
      if (err) {
        return reject(err);
      }
      resolve(res);
    };

    function passThrough(argument: any) {
      return argument;
    }

    // Using passThrough as the serialize and deserialize functions
    conn.makeUnaryRequest(path, passThrough, passThrough, data, resultCallback);
  });
};
```

This was not obvious at all. Am I missing something from the documentation, or is there some generated code somewhere I haven't seen?

torkelrogstad:
I see that if I set `outputServices=generic-definitions` as an option, I get generated stubs that seems to do everything I need for me. However, I then have to resort to handling responses with callbacks. Is there any reason that the default behavior is to not generate a complete client?

stephenh:
[...]

PR #1094, opened 2024-08-18
Added generated client implementations

jayakrishnancn:
Not clear how to use the generated client implementations

thanks to @torkelrogstad 
ref: https://github.com/stephenh/ts-proto/issues/495


stephenh:
This is great, @jayakrishnancn , thank you! Happy to merge any future doc updates/clarifications that you'd like to make as well! :-)

Issue #1080, opened 2024-07-30
Adding Explicit Type Definitions to Factory Method Objects


sam-step:
Hi, I recently ran into the "inferred type of this node exceeds the maximum length" error referenced in https://github.com/stephenh/ts-proto/issues/449

```
TS7056: The inferred type of this node exceeds the maximum length the compiler will serialize. An explicit type annotation is needed.
```

It can be easily reproduced by running ts-proto on [google/protobuf/descriptor.proto](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto).

I understand turning off exact types works around the problem. However, the exact types feature is pretty nice so I'm looking to see if there are alternative solutions.

Following the compiler's suggestion to add explicit annotations seems to address the problem. E.g.
```diff
-export const FileDescriptorProto = {
+export const FileDescriptorProto: {
+    encode(message: FileDescriptorProto, writer?: _m0.Writer): _m0.Writer;
+    decode(input: (_m0.Reader | Uint8Array), length?: number): FileDescriptorProto;
+    fromJSON(object: any): FileDescriptorProto;
+    toJSON(message: FileDescriptorProto): unknown;
+    create<I extends Exact<DeepPartial<FileDescriptorProto>, I>>(base?: I): FileDescriptorProto;
+    fromPartial<I extends Exact<DeepPartial<FileDescriptorProto>, I>>(object: I): FileDescriptorProto;
+} = {
  encode(message: FileDescriptorProto, writer: _m0.Writer = _m0.Writer.create()): _m0.Writer {
```

Is adding support for these explicit annotations something you'd consider (it certainly produces more verbose code...)? If so, happy to take a stab at contributing a PR for it.

stephenh:
Hi @sam-step ! Apologies for the late reply, but the explicit annotations sounds worth trying!

I was thinking it will blow up the `*.ts` file output size, *but* it's just a type, so would get minimized out in terms of actual `*.js` code.

Do you think doing an interface like `const FileDescriptProto: SomeGenericType<FileDescriptorProto>` would also work?

Even if `SomeGenericType` itself was included in the generated output (instead of being imported from an npm library), seems like that would also cut down on the output size quite a bit.

If you're still interested, would be great to have a PR! Thanks!

sam-step:
> Do you think doing an interface like const FileDescriptProto: SomeGenericType<FileDescriptorProto> would also work?

[...]

PR #1104, opened 2024-09-05
feat: Add interface for static message methods

sam-step:
Fixes https://github.com/stephenh/ts-proto/issues/1080

Adds an interface for the static message methods to avoid tsc errors on large messages. 

Since the set of static methods created may vary from message-to-message, multiple interfaces were created and are unioned as needed:
- `MessageFns`. The common encode/decode methods. All the message objects implement this.
- `ExtensionFns`. The get/set extension methods. Added if extensions are enabled and the message has extensions.
- `ExtensionHolder`. The individual extension properties.
- `[Struct|AnyValue|ListValue|FieldMask]WrapperFns`. The wrap/unwrap methods for the well-known types.

### DIFF
--- a/integration/affixes/affixes.ts
+++ b/integration/affixes/affixes.ts
@@ -55,7 +55,7 @@ function createBasePrefixAwesomeMessageSuffix(): PrefixAwesomeMessageSuffix {
   return {};
 }
 
-export const PrefixAwesomeMessageSuffix = {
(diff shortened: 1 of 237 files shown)


stephenh (approved):
Great use of the conditional output feature @sam-step ! Looks great! Thank you!

PR #1108, opened 2024-09-20
chore: Fix typos

NathanBaulch:
Just thought I'd contribute some typo fixes that I stumbled on. Nothing controversial (hopefully).

Use the following command to get a quick and dirty summary of the specific corrections made:
```shell
git diff HEAD^! --word-diff-regex='\w+' -U0 \
  | grep -E '\[\-.*\-\]\{\+.*\+\}' \
  | sed -r 's/.*\[\-(.*)\-\]\{\+(.*)\+\}.*/\1 \2/' \
  | sort | uniq -c | sort -n
```

FWIW, the top typos are:
* nsted (20)
* emtpy (13)
* detatch (6)
* unparseable (4)
* specficied (2)
* straming (2)
* inteegration (1)
* mutliple (1)
* modifer (1)

### DIFF
--- a/integration/meta-typings-as-const/google/protobuf/descriptor.ts
+++ b/integration/meta-typings-as-const/google/protobuf/descriptor.ts
@@ -167,7 +167,7 @@ export interface FieldDescriptorProto {
    * If true, this is a proto3 "optional". When a proto3 field is optional, it
    * tracks presence regardless of field type.
    *
-   * When proto3_optional is true, this field must be belong to a oneof to
+   * When proto3_optional is true, this field must belong to a oneof to
    * signal to old proto3 clients that presence is tracked for this field. This
    * oneof is known as a "synthetic" oneof, and this field must be its sole
    * member (each proto3 optional field gets its own synthetic oneof). Synthetic
@@ -429,7 +429,7 @@ export interface FileOptions {
   /**
    * By default Swift generators will take the proto package and CamelCase it
    * replacing '.' with underscore and use that to prefix the types/symbols
-   * defined. When this options is provided, they will use this value instead
+   * defined. When this option is provided, they will use this value instead
    * to prefix the types/symbols defined.
    */
   swiftPrefix?:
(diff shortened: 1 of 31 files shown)


integration/meta-typings-as-const/google/protobuf/descriptor.ts
```diff
@@ -167,7 +167,7 @@ export interface FieldDescriptorProto {
    * If true, this is a proto3 "optional". When a proto3 field is optional, it
    * tracks presence regardless of field type.
    *
-   * When proto3_optional is true, this field must be belong to a oneof to
```
stephenh, line 170: @NathanBaulch these `google/*/*.ts` files/comments come from comments within the google `*.proto` files that we can't change, so I think we need to revert any `descriptor.ts` changes.

protos/google/protobuf/compiler/plugin.proto
```diff
@@ -103,7 +103,7 @@ message CodeGeneratorResponse {
   // This should be used to indicate errors in .proto files which prevent the
   // code generator from generating correct code.  Errors which indicate a
   // problem in protoc itself -- such as the input CodeGeneratorRequest being
-  // unparseable -- should be reported by writing a message to stderr and
+  // unparsable -- should be reported by writing a message to stderr and
```
stephenh, line 106: These comments are also probably coming from a google `*.proto` file

stephenh (approved):
Hi @NathanBaulch , thanks for the PR! The typo fixes are appreciated. Just one ask to revert some of the comments in code generated from `*.proto` files that we don't control/can't fix the source typo in, but otherwise this is great.

NathanBaulch:
Restored all typos in `**/google/protobuf/**/*.*` files, as requested.

integration/meta-typings-as-const/google/protobuf/descriptor.ts
NathanBaulch: Done!

protos/google/protobuf/compiler/plugin.proto
NathanBaulch: Also done!

stephenh:
Looks great, thanks @NathanBaulch !

PR #1113, opened 2024-09-29
fix: Compilation error for nested repeated fields with `useOptionals=all`

### LINKED ISSUE #1112: Compilation error for nested repeated fields with `useOptionals=all`
The TS code generated for the following protobuf code with the `useOptionals=all` parameter does not compile.

```proto
syntax = "proto3";

message NestedList { repeated google.protobuf.StringValue a_string = 1; }

message Example { NestedList list = 1; }
```

The problem lies in the `decode` function not taking into account that the `value` after decoding could also be `undefined`. Therefore the following generated line:

```ts
message.aString!.push(StringValue.decode(reader, reader.uint32()).value);
```

results in the following TS compilation error:

```
Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.ts(2345)
```

vecerek:
Fixes #1112.

The fix is to push the decoded element to the array only if it's not an `undefined` value.


integration/nested-repeated-values-optional-all/nested-repeated-values-optional-all.ts
```diff
@@ -0,0 +1,165 @@
...
+    while (reader.pos < end) {
+      const tag = reader.uint32();
+      switch (tag >>> 3) {
+        case 1:
+          if (tag !== 10) {
+            break;
+          }
+
+          const el = StringValue.decode(reader, reader.uint32()).value;
+          if (el) {
```
stephenh, line 45: @vecerek if `el` was an empty string, we probably want to still `.push` it? I.e. should we just for undefined instead of truthiness?

stephenh (approved):
One question about empty strings but otherwise looks great, thanks for the PR @vecerek !

integration/nested-repeated-values-optional-all/nested-repeated-values-optional-all.ts
vecerek: Yes, you're correct! I'll make that change and I've just noticed I didn't push the re-generated TS files and I noticed another problem related to multiple `case` statements:

```
Cannot redeclare block-scoped variable 'el'.ts(2451)
```

I'll have to wrap the cases in a block to address this issue.

vecerek: @stephenh I've just force-pushed the rest of the changes. Thank you for the swift review!

stephenh (approved):
Looks great, thanks!

Issue #1109, opened 2024-09-20
Feature request: Deduplicate helpers and helper types across multiple generated files


absidue:
Currently when multiple proto files are passed in, the helper functions such as `isSet`, `longToNumber`, `bytesFromBase64` and `base64FromBytes` and the helper types `Builtin`, `DeepPartial` and `KeysOfUnion` are written into every single file. To reduce that duplication it would be great if those helpers and helper types were written into a separate file once and then imported where they are needed, that way they would only appear once in the output.

If that extra file and imports are undesirable for Node.js use, it could be hidden behind the `--ts_proto_opt=env=browser` option, as people targetting web browsers usually want smaller code sizes and use bundlers which are able to handle the extra imports.

Example:
- generated
  - \_\_ts_proto_helpers__.ts
  - example_request.ts -> imports from `__ts_proto_helpers__.ts`
  - example_response.ts -> imports from `__ts_proto_helpers__.ts`

stephenh:
Hi @absidue ; I agree this would be a good improvement.

You mention making it conditional based on `env=browser`; not against that, but honestly I wouldn't mind just trying it, as the default output with no conditionals/no flags, and see if that works for everyone.

It would be amazing if you could work on a PR that does this; lmk if you have any questions/want any pointers about where to start. Thanks!

absidue:
I have not looked into the codebase much yet, but here is my proposal on how something like this could potentially be implemented. Please let me know what you think about it and if you have answers to some of the questions below.

I would suggest generating and outputting the helper file right at the end, that way we only need to do it once. To do that we would need some way of keeping state across the different files that are processed so that we can keep a tally of what we actually need to output into the helpers file. Without knowing the codebase, in my head it makes the most sense to do that in the place that currently writes the helpers to the file, so instead of writing the helper or the type, it would record that it needs it in the "global" state and then generate an import instead.

Potential difficulties and open questions:
1. To keep the code simpler, I would suggest outputting the separate helper file all the time, even if the user only passes in a single proto file, that will also make the behaviour more predictable from the perspective of the users.
2. We would need to make sure that the helper file doesn't conflict with any of the other generated files, this could either be done by preventing users from naming their files the same as the helper file or by ensuring we generate a unique name if there is a conflict, the latter would make it more difficult for the people using the generated files, as the file name would be unpredictable.
3. As the output can be nested into various different folders we need to be able to generate correct relative imports even if the proto file output is nested in various directories.
[...]

PR #1115, opened 2024-09-30
Export generated `Rpc` interfaces

### LINKED ISSUE #1107: `interface Rpc` not exported in service definition
When I generate a service the `Rpc` interface is not exported, should it be?

```typescript
interface Rpc {
  request(service: string, method: string, data: Uint8Array): Promise<Uint8Array>;
}
```

I'm trying to implement it as a custom rpc to pass to my `ServiceClientImpl` that is generated, but since it's not exported I have to edit the file to add `export`.

marcaddeo:
This updates the generation of the `Rpc` interface to be publicly exported so consumers can define custom `Rpc` implementations.

Closes #1107

### DIFF
--- a/src/generate-grpc-web.ts
+++ b/src/generate-grpc-web.ts
@@ -173,7 +173,7 @@ function generateGrpcWebRpcType(ctx: Context, returnObservable: boolean, hasStre
   const { options } = ctx;
   const { useAbortSignal } = options;
 
-  chunks.push(code`interface Rpc {`);
+  chunks.push(code`export interface Rpc {`);
 
   const wrapper = returnObservable ? observableType(ctx) : "Promise";
   chunks.push(code`
--- a/src/generate-services.ts
+++ b/src/generate-services.ts
@@ -456,7 +456,7 @@ export function generateRpcType(ctx: Context, hasStreamingMethods: boolean): Cod
     ]);
   }
   const chunks: Code[] = [];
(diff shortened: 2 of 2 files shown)


src/generate-grpc-web.ts
```diff
@@ -173,7 +173,7 @@ function generateGrpcWebRpcType(ctx: Context, returnObservable: boolean, hasStre
   const { options } = ctx;
   const { useAbortSignal } = options;
 
-  chunks.push(code`interface Rpc {`);
+  chunks.push(code`export interface Rpc {`);
```
stephenh, line 176: Hi @marcaddeo , thanks for the PR!

Can you update this to conditionally add the `export`, i.e. something like:

```ts
  const maybeExport = options.exportCommonSymbols ? "export" : "";
```

Because we have some users where, when working with multiple `*.proto` files, these common exports can import conflicts, so they can opt to turn off the export. Thanks!

marcaddeo: Ah dang, that's true. I'm one of those people who opt out of the common exports too... So now this isn't going to help me at all 😅 

Will update though.

stephenh: Oh no! Sorry! :-)

Fwiw #1109 is an idea to make a shared `./ts-proto-types.ts` type file that have all these common types, and be output all the time, in just one spot, in theory allowing us to kill the common exports problem.

If you're up for tackling a PR for that, that'd be amazing! :pray:

PR #1119, opened 2024-10-05
fix: Don't fail on Function message names

### LINKED ISSUE #1110: Add support for reserved message names.
```
// Function notification settings
message Function {
    string namespace = 1;
    string name = 2;
}
```

Generates a file that typescript is unable to process (for good reason)
```
 error TS2345: Argument of type 'Function | {}' is not assignable to parameter of type 'Function'.
  Type '{}' is missing the following properties from type 'Function': namespace, name
```

I know i can use `--ts_proto_opt=typePrefix=MyPrefix` or `--ts_proto_opt=typeSuffix=MySuffix` to add extra to each type, but that would do it for everything, not just reserved words, which would make migrated to generated output a little rougher.

Is there anything I can do to handle the reserved ones? I am up for contributing.

```
message Object {
  string key = 1;
}
```

Is another example.

halkeye:
I tried other resolved words from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects#fundamental_objects and nothing seems to break.

Fixes #1110


stephenh:
Looks great @halkeye , appreciate testing for the extra reserved words!

Issue #1121, opened 2024-10-12
Nested message named "Function" now ends up as *_FunctionMessage in generated ts.


jdillon:
Caused by https://github.com/stephenh/ts-proto/commit/da048a1e78cdc1baa228700c40a944652998d2a1#diff-d34bf5280590ab473d1ad488821d664a0746c3459a2d1f3c2d7ea6037be2b042R63
from PR: https://github.com/stephenh/ts-proto/issues/1110

This change causes nested messages to break from previous behavior.

```
message Scope {
   message Function {
   }
}
```

Used to render as `interface Scope_Function` but now will render as `interface Scope_FunctionMessage` ... which is erroneously appending `Message` to this type name.

jdillon:
Special handling for messages that conflict with special Javascript names (Function, Date) probably should only do so if they are not nested messages?

stephenh:
Hi @jdillon , ah yeah, agreed we don't need the `...Message` suffixing for nested messages.

If you could submit a PR that adds that check (where the `builtIns` array is checked and `Message` is conditionally appended), that would be great! Thanks!

jdillon:
@stephenh I can look, though at first glance I'm not sure how to tell if `message: DescriptorProto` is a nested message or not.  so if you have any hints, then I can put a simple PR together for you.

stephenh:
Np! I'm 80% sure that `protoPrefix` is only non-empty for nested messages:

```
  messages.forEach((message, index) => {
    // I.e. Foo_Bar.Zaz_Inner
    const protoFullName = protoPrefix + message.name;
    // I.e. FooBar_ZazInner
    const tsFullName = tsPrefix + maybeSnakeToCamel(messageName(message), options);
```

That call-site also has a `isRootFile` local variable that is _probably_ true for top-level messages, and false for nested messages.

Those are the two things I'd try, and hopefully one of the two works.

halkeye:
I have a PR up, i didn't see the new ticket created

https://github.com/stephenh/ts-proto/pull/1123

stephenh:
Fixed in #1123 , thanks @halkeye !

jdillon:
thank you!

PR #1123, opened 2024-10-15
fix: Unbreak a use case for #1110 / fix for #1121

halkeye:
"Function" is builtin, but anything prefixed or suffixed for whatever reason shouldn't be handled the same way. So check if the final result is in builtin list, not the original unmodified name

I have concerns about maybeSnakeToCamel being called earlier now, but i can't find any tests for options.snakeToCamel.includes("keys") so i'm not sure if i broke anything

```
  /** Converts `key` to TS/JS camel-case idiom, unless overridden not to. */
💡export function maybeSnakeToCamel(key: string, options: Pick<Options, "snakeToCamel">): string {
    if (options.snakeToCamel.includes("keys") && key.includes("_")) {
      return snakeToCamel(key);
    } else {
      return key;
    }
  }
```

But all tests pass

### DIFF
--- a/src/visit.ts
+++ b/src/visit.ts
@@ -39,7 +39,7 @@ export function visit(
     const protoFullName = protoPrefix + enumDesc.name;
     // I.e. FooBar_ZazInner
     const tsFullName = tsPrefix + maybeSnakeToCamel(enumDesc.name, options);
-    const tsFullNameWithAffixes = `${options.typePrefix}${tsFullName}${options.typeSuffix}`;
+    const tsFullNameWithAffixes = messageName(`${options.typePrefix}${tsFullName}${options.typeSuffix}`);
     const nestedSourceInfo = sourceInfo.open(childEnumType, index);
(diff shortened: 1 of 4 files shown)


src/visit.ts
```diff
@@ -51,8 +51,8 @@ export function visit(
     // I.e. Foo_Bar.Zaz_Inner
     const protoFullName = protoPrefix + message.name;
     // I.e. FooBar_ZazInner
-    const tsFullName = tsPrefix + maybeSnakeToCamel(messageName(message), options);
-    const tsFullNameWithAffixes = `${options.typePrefix}${tsFullName}${options.typeSuffix}`;
+    const tsFullName = tsPrefix + maybeSnakeToCamel(message.name, options);
+    const tsFullNameWithAffixes = messageName(`${options.typePrefix}${tsFullName}${options.typeSuffix}`);
```
stephenh, line 55: Nice! Great fix to just move where we check the `builtInNames`

stephenh (approved):

halkeye:
🙏 can i get another label :P

stephenh:
Of course! Added.

Issue #975, opened 2023-12-08
`outputSchema=true` throwing when compiling `googleapis` protos


rosslavery:
Hi there,

Running with the following options is throwing with the below error.

```
protoc --plugin=./node_modules/.bin/protoc-gen-ts_proto \
  --experimental_allow_proto3_optional \
  --proto_path $path \
  --ts_proto_out=$outdir \
  --ts_proto_opt=outputSchema=true \
  ./google/ads/googleads/$version/**/*.proto
```

```
FAILED!Cannot read properties of undefined (reading '17')TypeError: Cannot read properties of undefined (reading '17')
    at encodedOptionsToOptions (/Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/schema.js:191:51)
    at /Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/schema.js:203:34
    at Array.forEach (<anonymous>)
    at resolveMessageOptions (/Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/schema.js:201:19)
    at /Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/schema.js:86:33
    at Array.forEach (<anonymous>)
    at generateSchema (/Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/schema.js:85:34)
    at generateFile (/Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/main.js:252:52)
    at /Users/someuser/Desktop/Work/somerepo/node_modules/.pnpm/ts-proto@1.164.0/node_modules/ts-proto/build/plugin.js:43:54
    at Array.map (<anonymous>)--ts_proto_out: protoc-gen-ts_proto: Plugin failed with status code 1.
```

Which points to this line (reading `'17'` of `extensionCache[extendee]`):
```ts
const extension = extensionCache[extendee][parseInt(key, 10) >>> 3];
```

Looking to output the schema to use in some downstream tooling, but not having any luck 🙃

tmtd4c93aac:
[...]

PR #1127, opened 2024-10-22
fix: Added null propagation and guard

ChapterSevenSeeds:
This PR addresses the issue presented in https://github.com/stephenh/ts-proto/issues/975.


integration/affixes/google/protobuf/timestamp.ts
```diff
@@ -0,0 +1,226 @@
+// Code generated by protoc-gen-ts_proto. DO NOT EDIT.
+// source: google/protobuf/timestamp.proto
+
+/* eslint-disable */
+import { BinaryReader, BinaryWriter } from "@bufbuild/protobuf/wire";
```
stephenh, line 5: Ah interesting; @ChapterSevenSeeds do you know why these two files were generated now? Was it from your bug fix, or were they just not included on a prior PR? :thinking:

stephenh (approved):
This is great, thank you!

integration/affixes/google/protobuf/timestamp.ts
ChapterSevenSeeds: They appeared when I followed the instructions for running the tests. I imagine your assumption that they weren't included from a prior PR is correct 😛

Issue #570, opened 2022-05-23
support for custom options in service / rpc


NimrodAvni:
I think a great feature to have is for [custom options](https://developers.google.com/protocol-buffers/docs/proto3#options)
 from the proto definition to be present in the generated code
something similar to [protobuf-ts](https://github.com/timostamm/protobuf-ts/blob/master/MANUAL.md#custom-options) is doing.
for example when generating a generic service definition from

```proto
service MyService {
    rpc MyMethod (MyMethodRequest) returns (MyMethodResponse) {
        option (google.api.http) = {
            get: "/v1/my-method"
        };
    };
}
```

the resulting code would be something like

```typescript
export const MyServiceDefinition = {
  name: "MyService",
  fullName: "com.something.v1.MyService",
  methods: {
    myMethod: {
      name: "MyMethod",
      requestType: MyMethodRequest,
      requestStream: false,
      responseType: MyMethodResponse,
      responseStream: false,
      options: {
         "google.api.http": {
            "get":"/v1/my-method""
         }
      },
    }
}
```

i know that options is reserved for more well-defined options like `idempotencyLevel`,
so it might be better to have it in a separate options object like `unknownOptions` or something like that

stephenh:
Hi @NimrodAvni ; @Vilsol just landed a PR that outputs options, if you use the `outputSchema` flag:

https://github.com/stephenh/ts-proto/pull/437

It looks like you want the option output as part of the service definitions, and maybe just that, not the entire `outputSchema` output?

If so, that makes sense; that's probably something that could be built on top of @Vilsol 's work?

Happy to have a PR for it if you want to work on it. Thanks!

NimrodAvni:
ohh i didn't know the outputSchema option does that
thats actually exactly what im looking for
unfortunately when i enabled it the outputed schema does not compile for me when i run tsc

<img width="1151" alt="image" src="https://user-images.githubusercontent.com/35308213/171061572-0e50955f-71d2-4562-a1ce-75a3c8b3b960.png">


this error with the `& Record<unique symbol | unique symbol | "flatMap" | "flat" | "at", never>` comes up on a lot of fields in the `FileDescriptorProto1.fromPartial` method in all generated proto files
i dont know if my tsconfig is configure badly or something because the type signature seems to match and not contain the `$ Record<...>` type, but building succeeds without the  outputschema option

am i missing something? anything needs to be configured differently @stephenh @Vilsol ?

Vilsol:
[...]

PR #1133, opened 2024-11-10
feat(Test): add sample for testing proto3 custom options

zakhenry:
This PR is more of a request for comment - ts-proto doesn't yet support proto3 [custom options](https://protobuf.dev/programming-guides/proto3/#customoptions) (they are defined in proto2 docs, but are actually well supported in proto3). I've added a sample proto of the kinds of fields that can be used, and want to start discussion on how this might be incoporated into the ts-proto api (behind which flags etc)


integration/extensions-proto3/parameters.txt
```diff
@@ -0,0 +1 @@
+outputExtensions=true,unknownFields=true,outputServices=generic-definitions,outputServices=default
```
zakhenry, line 1: I initially thought I could get this working with `outputExtensions=true`, but actually I suspect this needs to be an entirely new flag `outputCustomOptions=true` ?

zakhenry:
wait this might just be https://github.com/stephenh/ts-proto/issues/570. I'll sanity check myself then close if it's covered by that flag

zakhenry:
yea it is, whoops

stephenh:
Np @zakhenry , good find that there is an existing issue for this -- re-reading that issue, sounds like there were some distractions/complications from the `Exact` type, but that otherwise the `outputSchema` afaiu includes options...

Granted, probably as part of its larger "schema for all your messages" output, but in theory it has code that reads the options, that could then be included in the grpc service definition output...

If you'd like to make a PR for #570 that just always outputs options into the grpc service output (if I'm following along), that'd be great! Thanks!

PR #1136, opened 2024-11-13
feat: add support for comments on union fields in generateOneofProperty

sefaphlvn:
- Added functionality to `generateOneofProperty` to include comments for each union field, enhancing documentation support for union fields in TypeScript output.
- Previously, comments for individual fields within a `oneof` type were omitted. This update gathers comments from each field and appends them to the generated type for clarity.

Issue: [#1122](https://github.com/stephenh/ts-proto/issues/1122)

PR #1140, opened 2024-11-25
feat: Avoid adding empty trailing comments to oneof unions

haines:
This PR tweaks the output from #1136 slightly to avoid adding the trailing `//` after the union member separator if there isn't a documentation comment for the following oneof field (or if comments are turned off).


stephenh (approved):
Nice! Thanks!

PR #1146, opened 2024-12-09
fix: google protobuf timestamps don't properly get suffixed when useDate=false and prefix/suffix

halkeye:


halkeye:
Also when I ran the `yarn build:test && yarn test` type stuff i ended up with some uncommitted files, not sure if thats right?
```
        integration/output-decode-include-types/google/
        integration/output-encode-decode-include-types-combined/google/
        integration/output-encode-include-types/google/
```

stephenh:
Looks great, thnaks @halkeye !

Per the uncommitted files, it's likely some of the last few contributors didn't include them.

If you'd like to do a separate PR that adds them, that'd be great, but also np/I can get around to it eventually as well. Thanks!

halkeye:
> If you'd like to do a separate PR that adds them, that'd be great, but also np/I can get around to it eventually as well. Thanks!

https://github.com/stephenh/ts-proto/pull/1147

PR #1148, opened 2024-12-09
feat: Make sure all types support prefix/suffix

halkeye:


halkeye:
I'm a little scared of how big this change is. And I was trying to save some battery power so didn't run complete test suite on my laptop.

halkeye:
seems to be happy.
It could be argued its a breaking change, since the apis could change, but since the google + prefix didn't compile, i don't think anyone can be using it to break for.

src/utils.ts
```diff
@@ -344,3 +344,7 @@ async function readPackageJson(): Promise<any> {
     return await import("../package.json" as string);
   }
 }
+
+export function wrapTypeName(options: Options, name: string) {
```
stephenh, line 348: Great clean up!

src/encode.ts
```diff
@@ -11,7 +11,7 @@ export function generateEncoder(ctx: Context, typeName: string): Code {
   }
 
   if (name == "Timestamp") {
-    const TimestampValue = impProto(ctx.options, "google/protobuf/timestamp", name);
+    const TimestampValue = impProto(ctx.options, "google/protobuf/timestamp", wrapTypeName(ctx.options, name));
```
stephenh, line 14: Ah sure, makes sense we need to wrap all of these imports :+1:

stephenh (approved):
Thanks for flushing out the coverage!

PR #1149, opened 2024-12-09
chore: Catch untracked codegen files

stephenh:


.github/workflows/build.yml
```diff
@@ -28,9 +28,11 @@ jobs:
         run: ./integration/pbjs.sh
       - name: Codegen
         run: yarn proto2ts
-      # This will fail if any git-tracked file has changed
+      # This will fail if any git-tracked file have changed
       - name: Diff Output
-        run: git status && git diff --exit-code
+        run: |
+          git add .
```
stephenh, line 34: @halkeye I tried out this `git add .` and it worked! Thanks for the suggestion!

PR #1153, opened 2024-12-27
fix: update basic grpc implementation

monster-anshu:
In this basic grpc implementation example `serialize` callback is not correct and throws error in testing it. I have update `serialize` with `Buffer.from` which is working fine.


stephenh:
Thanks @monster-anshu !

Issue #1167, opened 2025-03-21
ControllerMethods would check descriptor exists


aceHubert:
``` javascript
  for (const method of grpcMethods) {
      const descriptor: any = Reflect.getOwnPropertyDescriptor(constructor.prototype, method);
      if (descriptor) {
        GrpcMethod("MediaService", method)(constructor.prototype[method], method, descriptor);
      }
    }
```

`Reflect.getOwnPropertyDescriptor` could not get property descriptor from extend class, so it's null will cause error

stephenh:
@aceHubert ah yeah 🤔 that is unfortunate; if you could submit a PR that fixes this, that would be great! Thanks!

PR #1169, opened 2025-03-24
fix: descriptor is null cause error

aceHubert:
* Reflect.getOwnPropertyDescriptor could not get its descriptor from extends class #1167


src/generate-nestjs.ts
```diff
@@ -177,12 +177,12 @@ export function generateNestjsGrpcServiceMethodsDecorator(ctx: Context, serviceD
         const grpcMethods: string[] = [${grpcMethods.join(", ")}];
         for (const method of grpcMethods) {
           const descriptor: any = Reflect.getOwnPropertyDescriptor(constructor.prototype, method);
-          ${GrpcMethod}('${serviceDesc.name}', method)(constructor.prototype[method], method, descriptor);
+          descriptor && ${GrpcMethod}('${serviceDesc.name}', method)(constructor.prototype[method], method, descriptor);
```
stephenh, line 180: Hi @aceHubert !

If the `descriptor` comes back null, is it really fine to just skip calling this? It will probably mean the base class methods won't get hooked up correctly?

I wonder if we need a utility method like:

```ts
function getPropertyDescriptorFromChain(obj, prop) {
  let currentObj = obj;
  
  while (currentObj !== null) {
    const descriptor = Object.getOwnPropertyDescriptor(currentObj, prop);
    if (descriptor) {
      return descriptor;
    }
    currentObj = Object.getPrototypeOf(currentObj);
  }
  
  return undefined;
}
```

aceHubert: good idea, saving write `GrpcMethod()` on each subclass method 
your way is to throw error on runtime, implements ServiceController would check type on development.

stephenh:
@aceHubert can you run `yarn proto2ts` to update the codegen output with your change? Then hopefully the build will be green.

PR #1187, opened 2025-05-24
fix: support null in DeepPartial when useNullAsOptional

erayaydn0:
Added `null` to the `Builtin` union when `useNullAsOptional=true`, so that `DeepPartial<T>` correctly preserves `null` values.


src/main.ts
```diff
@@ -688,7 +688,7 @@ function makeDeepPartial(options: Options, longs: ReturnType<typeof makeLongUtil
     "Builtin",
     code`type Builtin = Date | Function | Uint8Array | string | number | boolean |${
       options.forceLong === LongOption.BIGINT ? " bigint |" : ""
-    } undefined;`,
+    } ${options.useNullAsOptional === true ? "null |" : ""} undefined;`,
```
stephenh, line 691: I didn't think of this approach (adding `null` to `Builtin`), but it's passing all the tests, so sgtm! :-D

We can merge this as-is, but fwiw it might be a good idea to create a `use-null-as-optional.test.ts` file, next to the `use-null-as-optional.ts` file and do something like:

```
const u = User.createPartial({
  // ideally this was a compile error w/o your change, but now works
  profile: null,
});
expect(u.profile).someAssertion();
```

Just to make sure we don't accidentally regress this change in the future.

Happy to merge this as-is, but just will be some caveats that I could see someone thinking "null shouldn't be in Builtin" down the road, and not realize they're breaking your use case.

stephenh (approved):
Lmk if you're fine adding the test; happy to merge this either way. Thanks!

src/main.ts
LuckyEden: I have a question, if the value is null in a generate operation with the `useNullAsOptional` option, shouldn't it be considered null? 

The default is undefined;
`message.profile = (object.profile !== undefined && object.profile !== null)
      ? ProfileInfo.fromPartial(object.profile)
      : undefined;`
      
 Maybe it can be done with other options, I need to review them again. Sorry if it sounds ridiculous. :D

stephenh: Hi @LuckyEden , sorry for the late reply here; yeah, I think you're right...

Historically ts-proto leaned towards my own bias of "always use `undefined`" and so would convert null and undefined to "just undefined" for simplicity.

Primarily b/c on the wire, protobuf doesn't have a way to different "what is null" vs. "what is undefined" so I didn't think it made sense for the programmer to have to write code that was like "if undefined do this, if null do that" when really only one of those `if`s would ever run.

However with `useNullAsOptional`, I'd have to refresh my memory on what that opt is for :sweat_smile: but it seems reasonable for that `: undefined` to be `: null`...

erayaydn0:
Hey, sorry I haven't had a chance to look at this due to being busy.

I merged it with the current main branch and integrated the conflict into the updated code.

I'm following up. 😅

PR #1203, opened 2025-07-22
fix NullValue handling when oneof=unions-value is set

### LINKED ISSUE #1202: `google.protobuf.Value` unwrap method returns enum value instead of null
The generated `unwrap` method for `google.protobuf.Value` has incorrect behavior for null values:

```ts
unwrap(message: Value): string | number | boolean | Object | null | Array<any> | undefined {
  return message.kind?.value;
}
```

**Issue:** When the protobuf value is `null`, the method returns `0` (the enum value) instead of `null` as indicated by the return type signature.

**Impact:** Cannot distinguish between a numeric `0` and `null` value.

**Expected:** Method should return `null` for null protobuf values, not the underlying enum value.

errorhandler:
Fixes #1202.

### DIFF
--- a/integration/oneof-unions-value/google/protobuf/struct.ts
+++ b/integration/oneof-unions-value/google/protobuf/struct.ts
@@ -461,7 +461,7 @@ export const Value: MessageFns<Value> & AnyValueWrapperFns = {
   wrap(value: any): Value {
     const result = createBaseValue();
     if (value === null) {
-      result.kind = { $case: "nullValue", value };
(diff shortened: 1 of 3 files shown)


zsalzbank:
Just took me a whole day to figure out this was my problem.

stephenh:
Hm, well that sounds frustrating @zsalzbank -- thanks for at least tracking it down to this PR and mentioning it; not really sure how this fell through the cracks, other than just I've been busy and behind on triage of issues/PRs.

I'll hit merge now; ty @errorhandler !

Issue #1207, opened 2025-08-08
Repeated fields are always packed


etiennep-chromium:
Repeated fields are always packed in binary format, regardless of the packed attribute.
See https://protobuf.dev/editions/features/#repeated_field_encoding

Generating binding for:
```
syntax = "proto2";

message Foo {
  repeated int32 foo = 1;
}
```

Expected behavior: Each `foo` is encoded a an individual varint
Actual behavior: `foo` is always encoded a length delimited array. The generated code uses write.fork() / join().

I believe this happens here:
https://github.com/stephenh/ts-proto/blob/ef98faab891adc447c57587d2d2f8ce935b589da/src/main.ts#L1741

where packedType() should check field.options.packed

Why this matters: perfetto uses a custom c++ implementation of protobuf, protozero which doesn't support forward compatibility re. packed fields. Thus it expects a non packed repeated field to never be packed.
Now neither perfetto and ts-proto are compliant w.r.t the wire format, but in a incompatible way.

stephenh:
Fixed in #1208 , ty @etiennep-chromium !

etiennep-chromium:
Hey sorry I got it wrong for proto2.
I'm working on the fix here: https://github.com/stephenh/ts-proto/pull/1209

PR #1210, opened 2025-08-08
fix: Honour packed repeated fields encoding for proto2

etiennep-chromium:
Follow up to https://github.com/stephenh/ts-proto/pull/1208/files
I got the proto2 logic wrong.

https://github.com/stephenh/ts-proto/issues/1207

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -165,7 +165,7 @@ export function toReaderCall(field: FieldDescriptorProto): string {
 export function packedField(field: FieldDescriptorProto, isProto3Syntax: boolean): number | undefined {
   if (isProto3Syntax && field.options?.packed === false) {
     return undefined;
-  } else if (!isProto3Syntax && !!field.options?.packed) {
+  } else if (!isProto3Syntax && !(field.options?.packed === true)) {
     return undefined;
   }
   return packedType(field.type);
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -165,7 +165,7 @@ export function toReaderCall(field: FieldDescriptorProto): string {
 export function packedField(field: FieldDescriptorProto, isProto3Syntax: boolean): number | undefined {
   if (isProto3Syntax && field.options?.packed === false) {
     return undefined;
-  } else if (!isProto3Syntax && !!field.options?.packed) {
+  } else if (!isProto3Syntax && !(field.options?.packed === true)) {
```
stephenh, line 168: Hi @etiennep-chromium ! I wonder if this would be easier to follow:

```ts
// proto3 defaults to packed, proto2 does not
const shouldPack = field.options?.packed ?? isProto3Syntax;
return shouldPack ? packedType(field.type) : undefined;
```

Which I think is what we're trying to do? I.e. if `field.options.packed` is defined (true or false), use it; if its undefined, then `iProto3Syntax` defaults to packed, but proto 2 defaults to not packed.

etiennep-chromium: Good idea that reads a lot eaasier.

stephenh (approved):

stephenh:
Thanks for the quick follow-up!

PR #1219, opened 2025-10-16
feat: add option for using Temporal timestamp representation

confusingstraw:
This PR adds support for a new `useDate` option: `temporal`. When used, it will cause the generated code to use the new [Temporal.Instant ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Instant) object.

The advantages of supporting the new API include higher precision timestamps (up to nanosecond precision, which aligns with the Protobuf spec), as well as a variety of Temporal-related utilities, such as [Duration](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration). In a future update, it may be useful to add a similar flag for using the `Temporal.Duration` object for `Duration` representation.

### DIFF
--- a/integration/tsconfig.proto.json
+++ b/integration/tsconfig.proto.json
@@ -4,7 +4,8 @@
     "lib": ["es2018", "DOM"],
     "esModuleInterop": false,
     "noUnusedParameters": false,
-    "noUnusedLocals": false
+    "noUnusedLocals": false,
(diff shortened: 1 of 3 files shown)


stephenh (approved):
Love to see temporal support; thank you @confusingstraw !

Looks like there was a lint error; can you take a look at that, and then I'll hit merge?

Issue #1165, opened 2025-02-25
Support for Protobuf editions


ek-1n574:
I'm trying to migrate Golang application to the new [Protobuf Opaque API](https://go.dev/blog/protobuf-opaque) using this [guide](https://protobuf.dev/reference/go/opaque-migration/). This requires `.proto` files with Protobuf Editions. Now trying to generate TypeScript from these proto files fails with

```
abc.proto: is an editions file, but code generator protoc-gen-ts_proto hasn't been updated to support editions yet.  Please ask the owner of this code generator to add support or switch back to proto2/proto3.
```

As suggested by the error message, `ts-proto` would need editions support for inter compatibility with Go using the Opaque API. Could Protobuf editions be supported?

PR #1234, opened 2025-12-19
feat(editions): support editions in plugin

kellen-miller:
## Overview

Implements https://github.com/stephenh/ts-proto/issues/1165

Enable edition-aware behavior in ts-proto and advertise edition support to protoc. Uses Edition enum from ts-proto-descriptors to derive proto3 syntax when syntax != proto2 and edition is 2023/2024. 

#### Note
The `package.json` temporarily uses `ts-proto-descriptors: "file:protos"` for testing, will set it to the published `2.1.0` after the related protos PR is merged/released. 

All the changes under `protos/google/protobuf` can be ignored, they are a part of the preliminary PR below. I just needed them here for testing.

## Related PRs
- https://github.com/stephenh/ts-proto/pull/1233

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -15,6 +15,7 @@ module.exports = {
   resolver: 'jest-ts-webcompat-resolver',
   moduleFileExtensions: ['js', 'ts'],
   testMatch: ['<rootDir>/integration/**/*-test.ts', '<rootDir>/tests/**/*-test.ts'],
+  prettierPath: require.resolve('prettier-legacy'),
   testPathIgnorePatterns: [
       '<rootDir>/integration/simple-esmodule-interop/*',
       '<rootDir>/integration/batching-with-context-esModuleInterop/*'
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "nice-grpc": "^2.1.4",
     "object-hash": "^3.0.0",
     "prettier": "^3.5.3",
+    "prettier-legacy": "npm:prettier@2.8.8",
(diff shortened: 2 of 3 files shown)


kellen-miller:
sure thing let me sync and rebase

stephenh:
@kellen-miller ty! Looks like our CI check is failing just b/c of all the codegen churn caused by the protobuf bump -- can you run the codegen step locally and just push up all the changes?

kellen-miller:
@stephenh Ran `yarn build` and `yarn build:test` and pushed the changes, looked like that fixed the diff output failures. Also fixed an issue that arose after that with the `yarn tsc:check` with this commit https://github.com/stephenh/ts-proto/pull/1234/commits/17a7553dafbc14f27369ea0cba17d3e1cb6f84e8

kellen-miller:
Seems like Jest’s inline snapshot updater calls Prettier v2 sync APIs (resolveConfig.sync / getFileInfo.sync). Prettier v3 removed those, so jest -u (and inline snapshot updates) break unless you provide a compatible Prettier. I had to use a hack to update those files and get around `jest -u` not updating the snapshot. Snapshots updated in the this commit https://github.com/stephenh/ts-proto/pull/1234/changes/9c93ed76487672a5b89278c36ecf3c33280884d4

I've added a fix to pin the jest snapshotter to use prettier v2 in this commit https://github.com/stephenh/ts-proto/pull/1234/commits/ee5ba50371a8ddab5c5f42b66a8291bfc2c045df. I can revert that change if you don't want to keep that in.

Issue #1239, opened 2025-12-25
Missing globalThis in Object.* calls in generated code


pomo-mondreganto:
I've had a proto message called `Object` for a while, and it worked until addition of `map<string, string>` field. That led to `Object.entries(...)` call to appear in the generated TS code in the context of `Object` class that shadowed the global `Object`. Similar problem was fixed a while ago with the global `Error` type in #988 so it seems like the same fix should be applied here.

Repro:

```proto
message Object {
  string a = 1; // Works
  map<string, string> b = 2; // Leads to breakage
}
```

Generated `fromJSON`: 

```typescript
  fromJSON(object: any): Object {
    return {
      a: isSet(object.a) ? globalThis.String(object.a) : "",
      b: isObject(object.b)
        ? Object.entries(object.b).reduce<Map<string, string>>((acc, [key, value]) => {
          acc.set(key, String(value));
          return acc;
        }, new Map())
        : new Map(),
    };
  }
```

TS error:
```
test.ts(76,18): error TS2339: Property 'entries' does not exist on type 'MessageFns<Object>'.
```

PR #1240, opened 2025-12-25
fix: Use globalThis.Object to avoid name collision with the message name

pomo-mondreganto:
This fixes issue #1239 the same way as PR #988 did.


stephenh (approved):
Looks great @pomo-mondreganto , thank you!

PR #1252, opened 2026-04-30
fix: wrappers are not generated for nested google.protobuf.Timestamp and google.protobuf.Struct

nzrvg:
When both `nestJs = true` and `useDate = true` are set in `buf.gen.yaml`, wrapper types are not emitted in the generated TypeScript output if `google.protobuf.Timestamp` or `google.protobuf.Struct` are only used in nested messages.

For example, given the following proto:

```proto
syntax = "proto3";
package example;

message Parent {
  string id = 1;
  Child child = 2;
}

message Child {
  google.protobuf.Timestamp created_at = 1;
  google.protobuf.Struct metadata = 2;
}
```

The generated TypeScript output does not include all required wrapper types/imports for those protobuf types.

**Expected behavior:** the generated output should include all required wrapper types/imports even when `google.protobuf.Timestamp` or `google.protobuf.Struct` are only referenced from nested messages.

If this PR is merged, I'd appreciate it if the fix could also be backported to v1.181.2, since some projects are still pinned to the 1.x line.


stephenh:
Hi @GrigoryNazarov96 ; thanks for the PR -- fwiw I hadn't really looked at the CI failures, and was just mashing the "Approve the CI" button -- was something particularly annoying happening?

GrigoryNazarov96:
Hi, I reopened this PR [here](https://github.com/stephenh/ts-proto/pull/1254)

PR #1254, opened 2026-04-30
fix: wrappers are not generated for nested google.protobuf.Timestamp and google.protobuf.Struct

GrigoryNazarov96:
When both `nestJs = true` and `useDate = true` are set in `buf.gen.yaml`, wrapper types are not emitted in the generated TypeScript output if `google.protobuf.Timestamp` or `google.protobuf.Struct` are only used in nested messages.

For example, given the following proto:

```proto
syntax = "proto3";
package example;

message Parent {
  string id = 1;
  Child child = 2;
}

message Child {
  google.protobuf.Timestamp created_at = 1;
  google.protobuf.Struct metadata = 2;
}
```

The generated TypeScript output does not include all required wrapper types/imports for those protobuf types.

**Expected behavior:** the generated output should include all required wrapper types/imports even when `google.protobuf.Timestamp` or `google.protobuf.Struct` are only referenced from nested messages.

If this PR is merged, I'd appreciate it if the fix could also be backported to v1.181.2, since some projects are still pinned to the 1.x line.


stephenh:
hi @GrigoryNazarov96 -- I just merged the fix, thank you for the PR!

Per the ask to backport -- honestly I'm fairly hesitant b/c I don't have a lot of time for ts-proto atm, and dealing with keeping the v1 lineage up-to-date/etc is nothing something I really want/can invest time in ... but here's a ~5 minute attempt in this PR:

https://github.com/stephenh/ts-proto/pull/1256

If it "just works" that's great -- if it doesn't, well, I tried :sweat_smile:

GrigoryNazarov96:
hi, @stephenh, I actually tried the fix on v1.181.2 version as well by patching `main.js` in `node_modules/ts-proto` in `postinstall` script. It worked okay:)

PR #1257, opened 2026-05-05
fix: precision and canonical form in useDuration=string Duration JSON helpers

### LINKED ISSUE #1176: fromJSON duration uses Number rather than Long when using useDuration=string option
Hello,

I'm using the new duration which is working very well, but its upsetting typescript as the actual seconds assigned to the Duration is not a long, but a number.

Would it be possible to cast this value to a long?

Thanks

sallustfire:
## Summary

The `useDuration=string` Duration helpers funnel `seconds` and `nanos` through a JS `number`, with three concrete defects:

- **Precision loss** past ~2^22 seconds (≈48 days): float64 ULP exceeds 1 ns, so round-trip silently drifts. Up to ~2 µs at i64 max.
- **Scientific notation emit** for sub-microsecond magnitudes: `Number.toString` produces `"1e-9s"` for `{seconds: 0, nanos: 1}`, which the decoder regex then rejects. Round-trip throws.
- **Spec non-conformance**: encoder doesn't emit canonical 0/3/6/9 fractional digits per [proto3 JSON](https://protobuf.dev/programming-guides/proto3/#json) - `"1.5s"` instead of
  `"1.500s"` - breaking byte-identical interop with peers.

\#1176 falls out as a side effect: with the previous code, `fromJSON` always returned `seconds` as a `number`, regardless of the declared `forceLong` mode.

## Fix

Rewrite both `generateCanonicalToJson` and `generateFromJson` Duration branches in `src/main.ts` so they keep `seconds` in the declared TypeScript domain (`number` / `bigint` / `Long` /
`string`) at all times.

- **Encoder** formats magnitude from the declared seconds type directly; selects 0/3/6/9 fractional digits via integer arithmetic on `absNanos`. For `forceLong=long`, magnitude uses
  `toString().slice(1)` so `Long.MIN_VALUE` survives the asymmetric-i64 overflow that breaks `negate()`.
- **Decoder** trims the `s` suffix, splits on `.`, and lets `BigInt` / `Long.fromString` / `Number` throw on non-numeric input. Empty seconds component is rejected up front.

Wire format is unchanged. Emit is now byte-identical to `protojson.Marshal` for the Duration WKT.

## Tests

Four integration packages, one per `forceLong` mode, totaling **76 tests**:

The existing `use-duration` package previously had a hand-maintained `google/protobuf/duration.ts` that didn't regenerate from its `parameters.txt`; both are aligned so `update-code.sh`
produces the committed fixture.

## Behavior changes worth release-noting

1. **Canonical emit format**: output strings change for ms / µs / trimmed-ns values (`"1.5s"` → `"1.500s"`, etc.). Anyone doing exact-string comparison against output sees a diff.
2. **Stricter parsing**: strings with an empty seconds component, missing `s` suffix, or non-numeric seconds throw instead of silently producing garbage. Previous regex was permissive.
3. **Type fidelity**: `fromJSON` returns the declared type. Callers that observed `seconds: number` at runtime under `forceLong=bigint` may need to update — though in practice the previous
   behavior crashed anything that did `seconds === 0n` or passed `seconds` into a binary `encode`, so the change is strictly an improvement.

Closes #1176.


stephenh (approved):
This is amazing @sallustfire , thank you!

PR #1259, opened 2026-05-14
feat: add useContextDataloaders option to opt out of DataLoaders cont…

### LINKED ISSUE #1217: `context=true` forces `getDataLoaders` contract
We're experiencing some issues with `async_hooks` and I am looking to remove call sites where our usage is unnecessary. One of them is that we're using `async_hooks` to pass request context to propagate metadata to our rpc requests. I spotted there is a `context` option that even points out our use case as a reason for using this option:

> With --ts_proto_opt=context=true, the services will have a Go-style ctx parameter, which is useful for tracing/logging/etc. if you're not using node's `async_hooks` api due to performance reasons.

However, we now need to honour the dataloader contract:

> You also need to enable the useContext=true build-time parameter, which gives all client methods a Go-style ctx parameter, with a getDataLoaders method that lets ts-proto cache/resolve request-scoped [DataLoaders](https://github.com/graphql/dataloader), which provide the fundamental auto-batch detection/flushing behavior.

I don't want to add dataloaders. Can we make dataloaders an explicit opt-in rather than implicit when using `context=true`? 

My recommendation would be to introduce a `useContextDataloaders=?` option or something similar.

m-sanders:
Closes #1217

Adds `useContextDataloaders` option (default `true`) to opt out of the DataLoaders contract when using `context=true`.

Currently `context=true` forces `Context extends DataLoaders` and emits DataLoader-backed wrappers around `Get*` / `Batch*` methods. That's overkill if you only want `ctx` for tracing/metadata.

With `useContextDataloaders=false`:
- `Context` has no constraint
- no `DataLoader` / `object-hash` imports
- no synthesized singular `GetXxx` from `BatchXxx`
- no caching wrapper around `Get*` methods

Default stays `true` so existing fixtures regenerate byte-identical.


stephenh:
@m-sanders this PR looks great -- CI is failing, which is common for PRs that just need to "check in all the re-codegen-d" code, but for yours I didn't expect to need as much of that, since in theory this isn't a breaking change.

Can you dig around into the CI failures? Happy to approve/merge once those are passing.

Thanks!

PR #1261, opened 2026-06-26
fix: prevent code injection via string option/extension values in out…

hibrian827:
getExtensionValue() interpolated string extension values directly into generated TypeScript source via `code`"${value}"``. A string option value containing a quote and arbitrary code (e.g. `"});require('child_process')...({"`) would break out of the string literal and inject attacker-controlled code into the generated module, which executes when the output is imported or built.

Escape the value with JSON.stringify() so it is always emitted as a safe, properly-quoted string literal regardless of its contents.


stephenh (approved):

stephenh:
Looks great! Thanks @hibrian827 !

PR #1263, opened 2026-06-29
fix: Update js map check to include forceLong=bigInt as well

### LINKED ISSUE #940: Add support for bigint in maps
Hi, great project.

The library supports bigints but it does not support maps for those (map<uint64, uint64>)

I think something like [this issue](https://github.com/stephenh/ts-proto/issues/708) is missing. Basically, in this line https://github.com/stephenh/ts-proto/blob/main/src/types.ts#L685

```
    (isLong(mapType.keyField) && ctx.options.forceLong === LongOption.LONG)

```
should include the option bigint

Thanks

chiragpat:
### Why
Previous fix (#905) for this only fixed forceLong=long option. This PR adds support for forceLong=bigInt as well which fixes #940 

### How
- Update `shouldGenerateJSMapType` check to include bigint option as well
- Add new map field to existing bigint test


stephenh:
This is great @chiragpat , thank you!

PR #1265, opened 2026-07-02
docs: Update protobuf.js maintenance status

dcodeIO:
Maintainer of protobuf.js here. It's been actively maintained for a while again, so I figured it might be good to keep the ts-proto README current on its status as well.


stephenh (approved):
@dcodeIO Of course! Glad to have you back. :-)

PR #1266, opened 2026-07-02
feat: coerces bigint inputs inside fromPartial to bigint

### LINKED ISSUE #1264: fromPartial for bigint values doesn't support number and string types
## Description

I'm migrating ts-proto generated code from `Long` to bigint. `Long.fromValue` accepts `string`, `number`, `bigint` and `Long` which sometimes is very handy. But the same doesn't work with `forceLong=bigint`. It forces clients to always pass `bigint`. 

Is there any reason why it doesn't accept `string`, `number` and `bigint` and automatically convert value to bigint by using `BigInt()` constructor?

If no, then I could create a PR which handles this.

stalniy:
Closes #1264

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -321,7 +321,7 @@ Generated code will be placed in the Gradle build directory.
 
   With `--ts_proto_opt=forceLong=string`, all 64-bit numbers will be output as strings.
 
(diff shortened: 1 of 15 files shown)


stephenh:
Looks great @stalniy , thank you!

PR #1267, opened 2026-08-18
fix: avoid prototype pollution when decoding JSON maps

hibrian827:
The generated `fromJSON` reducer for a `map<string, message>` field assigned each entry with `acc[key] = ...` into a plain object accumulator. A JSON body whose map contains a `__proto__` key parses that key as an own enumerable property, so `Object.entries(...)` yields it and the bracket assignment invokes the `__proto__` setter — replacing the decoded map instance's prototype with attacker-controlled data.

Emit the assignment via `Object.defineProperty(acc, key, { value, ... })` so a `__proto__` (or any other) key is written as an own data property and never triggers the prototype setter.


stephenh:
hey @hibrian827 , thanks for the PR! Can you rerun the codegen and include that as part of the PR? We purposefully include the codegen output in the PRs as basically a diff/review of the output itself. Thanks!

hibrian827:
Sure! just pushed those that needed fix from the codegen you requested. Thanks for the reminder

stephenh:
Looks great!

PR #1268, opened 2026-08-18
fix: avoid prototype pollution in generated toJSON for __proto__ json…

hibrian827:
…_name

When a message field's `json_name` is `__proto__`, `getPropertyAccessor` emits dot notation and `toJSON` generated `obj.__proto__ = <object>` into a plain `{}` accumulator. For an object-valued field (message, map, repeated, oneof) that assignment invokes the `__proto__` setter and replaces the returned object's prototype with attacker-influenced descriptor data.

Route the object-valued assignments through a `setJsonProperty` helper that emits `Object.defineProperty(obj, "__proto__", { value, enumerable: true })` when the json_name is exactly `__proto__`, writing an own property instead of triggering the setter. Scalar assignments are unaffected (assigning a primitive to `__proto__` is a no-op in JS).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -2683,6 +2683,10 @@ function generateToJson(
     const jsonName = getFieldJsonName(field, options);
     const jsonProperty = getPropertyAccessor("obj", jsonName);
     const messageProperty = getPropertyAccessor("message", fieldName);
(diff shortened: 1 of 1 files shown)


stephenh:
These were all great PRs @hibrian827 , thank you!

PR #1274, opened 2026-09-15
fix: use jsonName for struct wrapper field names when useJsonName is set

### LINKED ISSUE #1125: using `--ts_proto_opt=useJsonName=true` breaks generated code for `google/protobuf/struct.proto`
We rely on `--ts_proto_opt=useJsonName=true` option so that the generated typing for our types has the correct casing. However, using this option causes the generated code for `google/protobuf/struct.proto` to be incorrect [like so](https://github.com/srliao/srsim/blob/769d819abbb33d6f5d98088a23ebe2b7317a70f8/js/packages/ts-types/src/generated/google/protobuf/struct.ts#L428-L438`):

```js
  wrap(value: any): Value {
    const result = createBaseValue();
    if (value === null) {
      result.null_value = NullValue.NULL_VALUE;
    } else if (typeof value === "boolean") {
      result.bool_value = value;
    } else if (typeof value === "number") {
      result.number_value = value;
    } else if (typeof value === "string") {
      result.string_value = value;
    } else if (globalThis.Array.isArray(value)) {
      result.list_value = value;
    } else if (typeof value === "object") {
      result.struct_value = value;
    } else if (typeof value !== "undefined") {
      throw new globalThis.Error("Unsupported any value type: " + typeof value);
    }
    return result;
  },
  ```
But the actual fields are named `nullValue`, `boolValue` etc... as can be seen as follows:
```js
function createBaseValue(): Value {
  return {
    nullValue: undefined,
    numberValue: undefined,
    stringValue: undefined,
    boolValue: undefined,
    structValue: undefined,
    listValue: undefined,
  };
}
  ```

Disabling `--ts_proto_opt=useJsonName=true` will result in all the field being named snake case for the generated `google/protobuf/struct.proto` code and the issues goes away but we are left with incorrect casing for our own definitions.

srliao:
Fixes #1125 

Disclaimer: This was authored with AI assistance as I didn't have the time previously to study how this project and the tests works to fix it when I initially filed the bug report.

I tested the CI on my own fork and everything passes.


srliao:
~~Sorry opened this PR on wrong repo. Meant to test on my own first. Will re-open once it's good to go~~

All good now

stephenh (approved):
Looks great @srliao , thanks!

